PR #5, opened 2017-02-18
Wire up JSON textarea to state so it displays the JSON

bennpearson:
## What is the context of this PR?

Wired up JSON textarea to state so it displays the JSON 
Related issue #4 

## How to review

Check json window is wired up correctly

### DIFF
--- a/src/components/JsonWindow.js
+++ b/src/components/JsonWindow.js
@@ -3,8 +3,18 @@ import './JsonWindow.css'
 
 export default class JsonWindow extends Component {
   render() {
+    const { description, id, number, title, type, guidanceTitle, guidanceText, answers } = this.props.question
     return (
-      <div className="json-window">{this.props.json}</div>
+      <div className="json-window">
+        <div className="json--description">'description': {description},</div>
+        <div className="json--id">'id': {id},</div>
+        <div className="json--number">'number': {number},</div>
+        <div className="json--title">'title': {title},</div>
+        <div className="json--type">'type': {type},</div>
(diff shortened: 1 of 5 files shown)


src/components/JsonWindow.js
```diff
@@ -3,8 +3,18 @@ import './JsonWindow.css'
 
 export default class JsonWindow extends Component {
   render() {
+    const { description, id, number, title, type, guidanceTitle, guidanceText, answers } = this.props.question
     return (
-      <div className="json-window">{this.props.json}</div>
+      <div className="json-window">
```
hamishtaplin: This should output the JSON as one big string instead of mapping it onto `divs`

src/containers/JsonWindow.js
```diff
@@ -1,6 +1,20 @@
 import { connect } from 'react-redux'
 import JsonWindow from '../components/JsonWindow'
+import { change } from '../actions'
 
-const JsonWindowContainer = connect()(JsonWindow)
+const mapStateToProps = (state, ownProps) => {
+  return { question: state.question }
+}
+
+const mapDispatchToProps = (dispatch, ownProps) => {
```
hamishtaplin: No need for this as the JsonWindow doesn't dispatch any actions.

samiwel:
I noticed while testing on this branch that when the JSON is rendered that there are no quotes displayed around string values (e.g. the Question Title). I'm guessing that the values are being treated as strings but for some reason the rendered JSON does not display string quotes.

PR #8, opened 2017-02-22
Wire up answers to update the JSON window.

### LINKED ISSUE #7: Render Response JSON for answers
The response fields should populate the relevant sections of the JSON.

samiwel:
Fixes #7.

This PR adds rendering of JSON for the response fields.

### DIFF
--- a/src/components/ResponseFields.js
+++ b/src/components/ResponseFields.js
@@ -25,6 +25,12 @@ export default class ResponseFields extends Component {
     }
   }
 
+  handleRemoveOption = e => {
+    const optionIndex = e.target.name
+    this.props.onRemoveOption(this.props.responseIndex, optionIndex)
+    e.stopPropagation()
+  }
+
   render () {
     const {id, mandatory, guidance, options, description, ...otherProps} = this.props.response
     const validationMessage = otherProps.validation.messages.MANDATORY
@@ -46,8 +52,10 @@ export default class ResponseFields extends Component {
           </Field>
           <Field>
               <Label htmlFor="options">Response Options</Label>
-              {/* <TextArea value={JSON.stringify(options, null, 2)} name="options" rows="10"/> */}
-              <OptionsInputList options={options} />
+              <OptionsInputList options={options}
+                onAddOption={this.handleAddOption}
+                onRemoveOption={this.handleRemoveOption}
+                emptyText="There are currently no options, add one below!" />
           </Field>
           <Field>
               <Label htmlFor="description">Response Description</Label>
--- a/src/components/ResponseFields.js
+++ b/src/components/ResponseFields.js
@@ -1,0 +1,62 @@
+import React, { Component } from 'react'
+
+import Field from './Field.js'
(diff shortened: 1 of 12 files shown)


src/reducers/question.js
```diff
@@ -11,14 +10,50 @@ const initialState = {
...
 }
 
 const question = (state = initialState, action) => {
   switch (action.type) {
-    case CHANGE:
+    case CHANGE_QUESTION:
       return {...merge(state, action.value)}
+    case CHANGE_RESPONSE:
+      const index = action.index
+      return Object.assign({}, state, {
```
hamishtaplin: Would be better to use destructuring over `Object.assign` for the sake of consistency

src/components/ResponseFields.js
```diff
@@ -0,0 +1,86 @@
...
+              <OptionsInputList options={options}
+                onAddOption={this.handleAddOption}
+                onRemoveOption={this.handleRemoveOption}
+                emptyText="There are currently no options, add one below!" />
+          </Field>
+          <Field>
+              <Label htmlFor="description">Response Description</Label>
+              <Input value={description} name="description"/>
+          </Field>
+          <Field>
```
hamishtaplin: I'm wondering if we even need this as it's not in Alexs prototype

src/components/ResponseFields.js
```diff
@@ -0,0 +1,86 @@
...
+
+  handleChange = e => {
+    const { name, value } = e.target
+    this.props.onChangeResponse(this.props.responseIndex, {[name]: value})
+    if (e.stopPropagation) {
+      e.stopPropagation()
+    }
+  }
+
+  handleChecked = e => {
```
samiwel: I think with checkbox that makes sense, since it's really just another CHANGE I guess. I'll take a look and see if I can refactor that. But what do you think about the add/remove options handlers? I think since the logic there is a little different they are better having their own handlers.

src/components/ResponseList.js
```diff
@@ -0,0 +1,22 @@
+import React, { Component } from 'react'
+
+import ResponseFields from './ResponseFields.js'
+
+export default class ResponseList extends Component {
+  render () {
+    return (
+      <div className="response-list">
+        { this.props.responses.map(response => {
```
hamishtaplin: This can be refactored to be more concise:
```javascript
{ this.props.responses.map(response => (
   <ResponseFields etc 
))}
```

src/components/ResponseFields.js
hamishtaplin: Yeah that add/remove ones are fine as they are a different action

src/components/ResponseFields.js
samiwel: @hamishtaplin I'm leaving this as it is for the time being as long as @alexmorris doesn't mind the deviation from the Codepen design? Just think the sooner this PR is merged it will be a better platform for Benn to work on his task.

All other changes have been made.

PR #11, opened 2017-02-27
WIP Answer options (don't merge)

hamishtaplin:
Did this as an exercise to see how far I could take the mapping of `name` to bits of state thing.

Answer: not very far.


hamishtaplin:
@samiwel good spot

You can't remove options either, I don't think this would work at all tbh and requires a more Redux-centric approach to be able to get the state to work correctly

PR #16, opened 2017-02-28
Store data between sessions

### LINKED ISSUE #15: Store data between sessions
Need the ability to persist the state between browser sessions.
eg. Wire up initial state to browsers local storage so it saves state between sessions

## Expected behaviour:

Navigate to page, make change to the question/answer, close the page, navigate back to the page, previous change is persisted.

## Current behaviour:

Changes are lost upon navigating back to the page.

samiwel:
## Description
Fixes #15 by using local storage to store the Redux store. This allows for changes in Question and Answer state to be persisted between sessions.

### DIFF
--- a/src/components/Checkbox.js
+++ b/src/components/Checkbox.js
@@ -2,11 +2,10 @@ import React, {Component} from 'react'
 
 export default class Checkbox extends Component {
   render() {
-    const { defaultChecked, checked, onChange, ...otherProps } = this.props
+    const { checked, onChange, ...otherProps } = this.props
 
     return (
-      <input type="checkbox" defaultChecked={defaultChecked}
-        checked={checked} onChange={onChange} {...otherProps} />
+      <input type="checkbox" checked={checked} onChange={onChange} {...otherProps} />
     )
   }
 }
--- a/src/components/Checkbox.js
+++ b/src/components/Checkbox.js
@@ -2,10 +2,10 @@ import React, {Component} from 'react'
 
 export default class Checkbox extends Component {
   render() {
-    const { checked, onChange, ...otherProps } = this.props
+    const { checked, ...otherProps } = this.props
 
     return (
-      <input type="checkbox" checked={checked} onChange={onChange} {...otherProps} />
+      <input type="checkbox" checked={checked} {...otherProps} />
     )
   }
 }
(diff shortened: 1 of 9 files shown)


samiwel:
This seems to work fine with all except for some reason the mandatory checkbox field and the options. I am looking into why these are not stored correctly...

samiwel:
With options, it seems that the `mapStateToProps` is called before the `load` from local storage is completed. I'm observing that the expected values are being stored in local storage when new options are added/removed, but at the point where the render is performed, the initialState seems to be used. Still looking into it, but it seems that the a load from local storage could be performed when setting the initial state.

samiwel:
After more investigation it seems that options are being read from local storage. My current thinking is that something weird happens when merging the Array values for `options` from local storage with the `initialState` property in the reducer.

If new options are added, then a subsequent re-visit tot he page will display the updated values.

`redux-storage` seems to have a few different mergers to choose from. The default being a [https://github.com/react-stack/redux-storage-merger-simple/blob/master/src/index.js](simple) merger. I have tried the `immutablejs` merger, but it didn't seem to fix the problem.

As there is another task to change the options component to break on newline, my recommendation is to accept this pull request as-is for now (with a slight weird behaviour loading options), as the rest of the components seem to function correctly.

hamishtaplin:
@samiwel that initialState in the reducer should be replaced by this, was just a placeholder really

src/components/Checkbox.js
```diff
@@ -2,8 +2,10 @@ import React, {Component} from 'react'
 
 export default class Checkbox extends Component {
   render() {
+    const { checked, onChange, ...otherProps } = this.props
+
     return (
-      <input type="checkbox" value={this.props.checked} onChange={this.props.onChange} {...this.props} />
+      <input type="checkbox" checked={checked} onChange={onChange} {...otherProps} />
```
hamishtaplin: Does `onChange` need to be here? I thought it would be caught by the handler in `QuestionForm`?

samiwel: Still needs to be handled by the ResponseFields handler, as it needs to know which answer (index) to update in the store. Not a big deal for a single answer, but if you added a second answer to the initial state then it would need some way of distinguishing between the two Checkbox fields.

hamishtaplin: @samiwel It doesn't seem to do anything at the moment? The change event bubbles up to the handler on `ResponseFields`. If you remove the prop entirely from `Checkbox` it still functions as expected. It just seems to be bypassing the event bubbling by passing/calling that handler in.

PR #17, opened 2017-03-01
Generate author components from schema-v1.json

samiwel:
This is just a Spike branch that I created to see how easy/difficult it is to generate fields dynamically from the EQ Survey Runner JSON schema.

So far I have only generated the Required fields at the top level of the survey.

But this branch shows that it can be done, with a bit of thought.
Being able to generate the fields from the schema will help when trying to generate the QBS schema via the authoring tool.


hamishtaplin:
This looks good, definitely has legs. I think that some higher level components composing the field, label and inputs together as a question type (eg. PositiveInteger) would be a good idea. We could probably do with refactoring how those components work anyway as there is a lot of repetition around name attributes, etc.

samiwel:
Yeah, it feels like we almost need some kind of mapping from the schema-v1 to what will be displayed in the authoring tool (Label, Component, DefaultValue etc..) probably worth us all having a session with Alex and going through the code and working out what needs to change at a high level, from that we can maybe break out into some more tasks.

samiwel:
@hamishtaplin Added some logic to recursively generate the components all the way down. There's still a few I haven't covered (I've added warnings for these in the Console so I don't forget.) Might be useful when we re-work this.

src/components/QuestionForm.js
```diff
@@ -14,10 +15,80 @@ export default class QuestionForm extends Component {
     })
   }
 
+  capitalize = str => {
```
hamishtaplin: These 3 utility functions might be convered by lodash, is worth checking through the docs

src/components/QuestionForm.js
```diff
@@ -14,10 +15,80 @@ export default class QuestionForm extends Component {
...
+  }
+
+  labelize = str => {
+    return this.capitalize(this.deunderscore(str))
+  }
+
+  renderField = (field, fieldName) => {
+    const fieldType = field.type
+
+    switch(fieldType) {
```
hamishtaplin: What do you think about using components for each field type? this would help with composing question/answer types later on

samiwel:
I'm going to close this PR for now, but I'll leave the branch on the server in case we want to refer back to it.

PR #21, opened 2017-03-03
Moved components into folders

hamishtaplin:
A quick restructure to move components into their own folders, grouping form elements together.

Convention is as follows:
```
/ComponentName
- index.js
- style.css
```


src/components/OptionsInputList/index.js
```diff
@@ -1,9 +1,9 @@
 import React, {Component} from 'react'
-import './OptionsInputList.css'
-import Label from './Label.js'
-import Field from './Field.js'
-import Input from './Input.js'
-import Link from './Link.js'
+import Label from 'components/forms/Label'
+import Field from 'components/forms/Field'
+import Input from 'components/forms/Input'
+import Link from 'components/Link'
```
samiwel: Any reason why Link is in a different directory to the other components? There might be a good reason for this? e.g. is the idea that we're splitting them out into privative components (standard HTML element wrappers) and composite components (combination of one-or-many primitives) or custom components (variations of primitive components such as Link) in a separate folders?

samiwel: It's only a small observation, but I've found that getting the taxonomy right can be quite difficult in the past. Just wondering if something like this would be better going forward:
```
components
├── composite
├── custom
└── primative
```

samiwel:
When I try to run on this branch I get the following error:
I did an `npm install` but didn't seem to pull anything down, did you check in changes to package.json or did you forget to `--save`?

image

hamishtaplin:
@samiwel We'll probably add react-router at some point, so I'd rather follow the "[fractal](https://github.com/davezuko/react-redux-starter-kit#application-structure)" convention of organising components. Was using this in the main author app and it works well.

The issue was the .env file I added that enables absolute imports was gitignored

PR #26, opened 2017-03-03
Fixes validation message fields don't update JSON window correctly

### LINKED ISSUE #24: Entering text into second validation component doesn't update JSON
## Steps to reproduce
1. Open up the authoring prototype.
2. By default there will be 2 responses
3. Enter text into the validation message on the second response.

## Expected behaviour
Text appears in the validation message field and JSON is updated.

## Actual behaviour
Text doesn't appear in the validation message field and JSON isn't updated.

## Screenshot
validation-message

samiwel:
## Description
Fixes #24.
Validation message component on subsequent answers doesn't update the JSON schema.

## How to test
Modify the default JSON so that it includes multiple answers.
Enter text in both validation message fields and observe that the JSON updates appropriately.

### DIFF
--- a/src/components/AnswerFields/index.js
+++ b/src/components/AnswerFields/index.js
@@ -48,6 +48,9 @@ export default class AnswerFields extends Component {
     const {id, mandatory, guidance, options, description, ...otherProps} = this.props.answer
     const validationMessage = otherProps.validation.messages.MANDATORY
 
+    const validationName = 'answers.{index}.validation.messages.MANDATORY'
+      .replace(/{index}/, this.props.responseIndex)
+
     return (
       <div onChange={this.handleChange}>
           <h3>Answer {this.props.answerIndex + 1}</h3>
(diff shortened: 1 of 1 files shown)


src/components/ResponseFields/index.js
```diff
@@ -48,6 +48,9 @@ export default class ResponseFields extends Component {
     const {id, mandatory, guidance, options, description, ...otherProps} = this.props.response
     const validationMessage = otherProps.validation.messages.MANDATORY
 
+    const validationName = 'answers.{index}.validation.messages.MANDATORY'
```
hamishtaplin: Use template strings here instead:

```javascript
const validationName = `answers.${this.props.responseIndex}.validation.messages.MANDATORY`
```

samiwel: 👍 I was looking for a way to do just that!

PR #31, opened 2017-03-06
Refactor several components to stateless functions.

### LINKED ISSUE #30: Refactor components to use stateless functions where possible
https://medium.com/@joshblack/stateless-components-in-react-0-14-f9798f8b992d#.l58kej5r8

samiwel:
## Description
Fixes #30. This PR refactors several of the existing components that do not have any internal state to be stateless functions instead.
In theory, any component that would render the same given the same props could be converted into stateless functions to improve performance. In practice, there are a few components that I have left as they were because they have other logic associated with them (event handler logic etc.).

## How to test
Ensure that the author application still behaves correctly.
Review the code changes and advise on any components that I may have missed.

### DIFF
--- a/src/components/ResponseList/index.js
+++ b/src/components/ResponseList/index.js
@@ -1,21 +1,15 @@
-import React, { Component } from 'react'
+import React from 'react'
 
 import ResponseFields from 'components/ResponseFields'
 
-export default class ResponseList extends Component {
-  render () {
-    return (
-      <div className="response-list">
-        { this.props.responses.map((response, index) => (
-          <ResponseFields key={`response-${index}`}
-              response={response}
-              responseIndex={this.props.responses.indexOf(response)}
-              onChangeResponse={this.props.onChangeResponse}
-              onChangeQuestion={this.props.onChangeQuestion}
-              onRemoveOption={this.props.onRemoveOption}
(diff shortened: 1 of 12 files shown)


src/components/ResponseList/index.js
```diff
@@ -1,21 +1,15 @@
...
-              onChangeResponse={this.props.onChangeResponse}
-              onChangeQuestion={this.props.onChangeQuestion}
-              onRemoveOption={this.props.onRemoveOption}
-              onAddOption={this.props.onAddOption} />
-        )) }
-      </div>
-    )
-  }
-}
+const ResponseList ({responses, ...other}) => (
```
hamishtaplin: `...other` is `...otherProps` elsewhere in the codebase, should be consistent.

src/components/forms/Checkbox/index.js
```diff
@@ -1,11 +1,7 @@
-import React, {Component} from 'react'
+import React from 'react'
 
-export default class Checkbox extends Component {
-  render() {
-    const { checked, ...otherProps } = this.props
+const Checkbox = (props) => (
+  <input type="checkbox" checked={props.checked} {...props} />
```
hamishtaplin: Destructure the props/otherProps.

PR #36, opened 2017-03-07
Refactor answers

hamishtaplin:
### Changes

- Flattened the state so `question` and `answers` are discrete chunks.
- Moved answer components out of the question component
- Tidied up some naming


src/containers/Question.js
```diff
@@ -0,0 +1,19 @@
...
+import { bindActionCreators } from 'redux'
+import Question from 'components/Question'
+import { changeQuestion } from 'actions/question'
+
+const mapStateToProps = (state) => ({
+  question: state.question
+})
+
+const mapDispatchToProps = (dispatch) => (
+  { changeQuestion: bindActionCreators(changeQuestion, dispatch) }
```
samiwel: What is the benefit of this approach over `dispatch(changeQuestion)`?

samiwel:
Also branch needs conflicts resolved.

src/containers/Question.js
hamishtaplin: It's just cleaner as you don't have to name the arguments

Issue #48, opened 2017-03-14
Create/load a survey


hamishtaplin:
- Create an survey empty survey
- Load from disk

samiwel:
We already have a mechanism for creating the empty survey, we need the mechanism to load in a schema and have it's contents populate all of the React component state tree.

samiwel:
Closing since PR already merged.

PR #73, opened 2017-04-06
Enables user to load schema file.

samiwel:
## Description
This PR fixes issue #48 by allowing them to select a JSON schema to load into the React component tree state.

### DIFF
--- a/src/components/FileUpload/index.js
+++ b/src/components/FileUpload/index.js
@@ -1,0 +1,44 @@
+import React from 'react'
+
+const FileUpload = (props) => {
+  let fileUpload = null
+
+  const onRef = (input) => {
+    fileUpload = input
+  }
+
+  const onClick = (e) => {
+    fileUpload.click()
+  }
+
+  const FileInput = (props) => (
+    <input type="file" ref={props.onRef}
+      hidden={props.hidden}
+      accept={props.accept}
+      onChange={props.onChange} />
+  )
+
+  const FileUploadWrapper = (props) => (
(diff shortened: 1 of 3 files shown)


src/components/FileUpload/index.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import styled from 'styled-components'
+
+const FileInput = (props) => (
```
hamishtaplin: Not sure this needs to be a separate component? It's not something we would use on it's own, it could just encapsulated in the FileUploadButton.

src/components/FileUploadButton/index.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react'
+import FileUpload from '../FileUpload'
+
+const FileUploadButton = (props) => {
```
hamishtaplin: As this isn't a Button but a wrapper for a Button I'd just called it FileUpload

src/components/FileUploadButton/story.js
```diff
@@ -0,0 +1,54 @@
...
+      <button>Open a file</button>
+    </FileUploadButton>
+  ))
+  .add('Restrict file type', () => (
+    <FileUploadButton accept=".json">
+      <button>Open a file</button>
+    </FileUploadButton>
+  ))
+  .add('Custom Button Style', () => {
+    const CustomButton = styled.button`
```
hamishtaplin: Maybe just use the existing Button component?

PR #76, opened 2017-04-11
Fixes broken GitHub Pages continuous delivery.

samiwel:
This PR aims to fix the broken continuous delivery to GitHub pages.

Since the introduction of react router into the project GitHub pages has been displaying a blank page.
After some research it seems there are a few factors at play.

1.  The `homepage` in the `project.json`.
2. The type of history used. Consensus being that [using hashHistory is preferable](http://stackoverflow.com/questions/40687493/seeing-empty-page-after-deployed-to-s3-with-react-route) for use on GitHub pages.
3. The `basename` used when constructing the router history. This does not reflect the root basename when published to GH pages.


samiwel:
Be good to get this reviewed and merged to try and fix our CD pipeline 👍

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "eq-author-prototypes",
   "version": "0.0.1",
   "private": true,
-  "homepage": "https://ONSdigital.github.io/eq-author-prototypes",
+  "homepage": "./",
```
samiwel: For reference, this allows you to serve the site locally from the build folder, as well as from GH pages.

PR #79, opened 2017-04-11
Refactor create survey page to utilise Grid component.

samiwel:
Bit of a tidy up to use the Grid component instead of rolling Yet Another React Component (YARC) for columns.


src/pages/survey_create.js
```diff
@@ -51,16 +44,16 @@ const SurveyCreatePage = () => {
               <Label>Description</Label>
               <Input name="number" />
             </Field>
-            <TwoCols>
-              <Field50Percent id="number">
+            <Grid>
+              <Column id="number">
```
hamishtaplin: Are these id's used for anything? They should be unique.

PR #92, opened 2017-05-09
Fix routing so that author deploys correctly to GitHub pages.

samiwel:
## Description
This PR fixes the broken routing on GitHub pages.

Although the authoring prototype was being deployed correctly, routing was broken and lead to 404 error being displayed when the page loads.
This fix sets the `basename` for the router history so that the react router can successfully resolve the correct components to load based on the browser history path.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -16,7 +16,14 @@ import rootReducer from 'reducers'; // Or wherever you keep your reducers
 import App from 'containers/App';
 
 // Create a history of your choosing (we're using a browser history in this case)
-const history = createHistory();
+let basename = ''
+if (process.env.NODE_ENV === 'production') {
+  basename = '/eq-author-prototypes'
+}
+
+const history = createHistory({
+  basename: basename
+})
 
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -16,7 +16,9 @@ import rootReducer from 'reducers'; // Or wherever you keep your reducers
 import App from 'containers/App';
 
 // Create a history of your choosing (we're using a browser history in this case)
-const history = createHistory();
+const history = createHistory({
```
hamishtaplin: Is it possible to only do this on a production build?

samiwel: Probably, I'd need to look into it. Does yarn have any concept of production/development environment?

samiwel: Maybe something like this http://stackoverflow.com/questions/35469836/detecting-production-vs-development-react-at-runtime

hamishtaplin: @samiwel Yeah that looks like it would work.

PR #102, opened 2017-05-16
1/load schema

WickyNilliams:
This feature was mostly complete, so this PR is adding (and establishing best practices) for tests. 

### How to review 
Run test suite (`yarn test`), ensure tests pass


WickyNilliams:
[Load Existing Schema (5)](https://trello.com/c/JJ3gUf94/1-load-existing-schema-5)

src/actions/survey.js
```diff
@@ -23,6 +29,27 @@ export function loadSurvey(surveyData) {
 
 export function clearSurvey() {
   return {
-    type: CLEAR_SURVEY
+    type: SURVEY_CLEAR
+  };
+}
+
+export function loadSurveyFailure(error) {
```
hamishtaplin: should probably rename action creators with the same  structure as the action names (noun first)

src/actions/survey.test.js
```diff
@@ -0,0 +1,84 @@
+import { CALL_HISTORY_METHOD } from "react-router-redux";
+import {
+  SURVEY_LOAD_FAILURE,
+  SURVEY_LOAD_SUCCESS,
+  loadSurvey,
+  loadSurveyFailure,
+  loadSurveySuccess
+} from "actions/survey";
+
+function actionMatching(type, payload = expect.anything()) {
```
hamishtaplin: Be good to move these into test utils file 🙃

src/actions/survey.test.js
```diff
@@ -0,0 +1,84 @@
...
+
+function createTextFile(contents = "", name = "foo.json") {
+  return new File(contents.split(""), name);
+}
+
+function createJsonFile(obj = {}, name) {
+  return createTextFile(JSON.stringify(obj), name);
+}
+
+describe("actions", () => {
```
hamishtaplin: maybe "survey actions"?

src/actions/survey.test.js
WickyNilliams: how about `actions/survey`?

src/actions/survey.test.js
WickyNilliams: How about `actions/survey`? So it mirrors the path

src/actions/survey.js
WickyNilliams: I followed the advice here in this article: https://medium.com/@kylpo/redux-best-practices-eef55a20cc72

They recommend `NOUN_VERB` for action names, since it gives them namespacing. But then they recommend `verbNoun()` for the action creator, since it's the more natural function name - if you wrote the function with no concern for redux, `loadSurvey` would be how you'd name it.

Seemed to make sense to me. What do you think?

src/actions/survey.test.js
WickyNilliams: Did you create a test utils folder somewhere in the end? If so, what was the path?

src/actions/survey.js
hamishtaplin: Cool, so long as there's a convention.

src/actions/survey.test.js
hamishtaplin: `src/tests/utils`

I put it into `src` because I couldn't get Jest to look outside that directory without using a relative path.

src/actions/survey.test.js
hamishtaplin: 👍

src/actions/survey.test.js
WickyNilliams: Sweet, i'll add them there

WickyNilliams:
@hamishtaplin all points addressed

WickyNilliams:
@samiwel i've rebased now, so is good to merge. Up to you if you squash. I've already squashed the noisy commits locally, so the commits should all be atomic

PR #110, opened 2017-05-18
Split surveys reducer into `meta` and `items`

hamishtaplin:

Quick bit of refactoring to split the `surveys` reducer into `meta` and `items`. Also did the same with the relevant actions.

Added some additional tests to increase coverage but not on the `items` bit as this is due a refactor anyway.

### How to review 

- Ensure all tests pass
- Ensure linting passes
- Ensure app works as expected


src/actions/survey.js
```diff
@@ -1,4 +1,11 @@
 import readFileAsJSON from "utils/readFileAsJson";
+
+export const LOAD_SURVEY = "LOAD_SURVEY";
```
samiwel: Do we need actions for both `LOAD_SURVEY` and `SURVEY_LOAD`? Might be confusing to have both actions there.
From what I've read, I think the convention should be `SURVEY_LOAD` for action names and the action creator function should be `loadSurvey()` because it is more natural.
Same for the SAVE, CLEAR and META_UPDATE below.

hamishtaplin: No we don't, that's just some bad rebasing on my part—will remove.

src/reducers/survey/index.test.js
```diff
@@ -0,0 +1,14 @@
+import { createStore } from "redux";
+
+import survey from "reducers/survey/";
+import meta from "reducers/survey/meta";
+import items from "reducers/survey/items";
+
+let store = createStore(survey);
```
WickyNilliams: Why do you need a store in some reducer tests, but not others?

src/actions/survey/index.js
```diff
@@ -10,21 +11,23 @@ export const SURVEY_LOAD_FAILURE = "SURVEY_LOAD_FAILURE";
...
     type: SURVEY_LOAD_SUCCESS,
     payload: {
-      ...entities.survey[result],
-      groups,
-      blocks,
-      sections,
-      questions,
-      answers
+      meta: {
+        ...entities.survey[result]
```
WickyNilliams: No need to create a new obj and spread it? This should work:

`meta : entities.surey[result]`

WickyNilliams:
[Refactor Reducers](https://trello.com/c/fFyhDiyi/34-refactor-reducers)

src/actions/survey/index.js
hamishtaplin: Yeah no idea why I did that!

src/reducers/survey/index.test.js
hamishtaplin: Good spot. Just two different ways of testing the initialState is working (which itself isn't a particularly valuable test!)

I guess the question is which one is preferable?

hamishtaplin: I don't know of any other way to get to the state of the (meta, items) sub-reducers via the main `survey` reducer.

PR #112, opened 2017-05-18
Refactor basename check into environment variable.

samiwel:
This PR aims to leverage the Webpack environment configuration to determine if `basename` should be set between development and production.
During the build script, the `BASE_NAME` environment variable is set for 'production' builds.
The Webpage config defaults to an empty base name whilst in development mode.
The code which uses the basename has been refactored so as to not require the environment check.

### How to review 
Ensure that base name is configured appropriately for development and production builds.
Advise on any improvements to how I'm handling config in this PR.


hamishtaplin:
@samiwel the basename seems to be causing issues on ghp at the moment, eg. clicking the "create" button takes you `https://onsdigital.github.io/eq-author/#/eq-author/create`. Does this PR fix this too?

WickyNilliams:
That, to me, suggests we don't actually need a base path since we switched to hash routing?

samiwel:
I think browserHistory routing was preferred but I read somewhere that it didn't play too nicely with GitHub pages.

WickyNilliams:
Browser history is a little pointless without server-side rendering, or a server that knows to catch all requests and return the same result for any path

WickyNilliams:
[Environment-specific vars (1)](https://trello.com/c/kqaSU38a/25-environment-specific-vars-1)

PR #118, opened 2017-05-22
29/split up entry point

WickyNilliams:
This change simplifies the entry point (`src/index.js`), by moving code around to increase cohesion (related code lives together). It also removes some indirection around configuring routing. All routes are now clearly visible in `App` component

### How to review 
Check changes are sensible. Ensure app works as before


WickyNilliams:
[Split up entry point JS module (2)](https://trello.com/c/hlLeAXkT/29-split-up-entry-point-js-module-2)

WickyNilliams:
Weirdly, codecov was failing on first push. So i rebased (just in case), pushed again, and now there's no codecov at all??!

src/index.js
```diff
@@ -1,76 +1,32 @@
...
-
-import thunk from "redux-thunk";
-
-import rootReducer from "reducers"; // Or wherever you keep your reducers
-
+import { configureStore, configurePersistedStore } from "store/configureStore";
 import App from "containers/App";
 
-let useStorage = false;
+let useStorage = true;
```
hamishtaplin: A good time to improve this perhaps?

src/index.js
```diff
@@ -1,76 +1,32 @@
...
-};
-
-useStorage
-  ? load(store).then(newState => {
-    render(App);
-  })
-  : render(App);
-
 if (module.hot) {
+  module.hot.accept("reducers", () => {
```
hamishtaplin: out of interest, was there a reason to swap these around?

src/index.js
WickyNilliams: Crap, should have left that as false. What kind of improvements are you thinking?

src/index.js
WickyNilliams: Accident! Will revert

src/index.js
hamishtaplin: Dunno, only think I can think of is a query param?

src/index.js
hamishtaplin: Doesn't matter, just wondering if there was a reason

PR #119, opened 2017-05-22
Backfill tests for Title component.

samiwel:
Add test which asserts missing prop types errors for Title component.
This is the first attempt at writing a snapshot test which will fail if a component changes. Title seemed like the easiest component to begin with.
This is designed to prevent inadvertent changes to components as discussed [here](https://facebook.github.io/jest/blog/2016/07/27/jest-14.html).
I will raise separate PR's for the other components.

### How to review 
Tests should pass and team should agree with approach of using snapshot tests to supplement component Unit tests.


src/components/Title/title.test.js
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import Title from "components/Title";
+import renderer from "react-test-renderer";
```
WickyNilliams: Why not use enzyme, which we already use elsewhere?

src/components/Title/title.test.js
```diff
@@ -0,0 +1,20 @@
...
+import { stubConsoleError } from "tests/utils/stubConsoleError";
```
WickyNilliams: I was going to look at configuring jest to always fail whenever an error is logged to console, to avoid having to remember to test for it explicitly like this

src/components/Title/title.test.js
```diff
@@ -0,0 +1,20 @@
...
+import { assert } from "sinon";
```
WickyNilliams: jest offers `expect` in global scope, no need for sinon

src/components/Title/title.test.js
```diff
@@ -0,0 +1,20 @@
...
+
+describe("Prop types work as expected", () => {
```
WickyNilliams: title of `describe` block should be the thing you're testing, in this case "prop types". I tend to nest them such that:

```js
describe("module name", () => {
  describe("module feature", () => {
    it("should meet some requirement", () => {

    });
  });
});
```

src/tests/utils/stubConsoleError.js
```diff
@@ -0,0 +1,11 @@
+import sinon from "sinon";
+
+export function stubConsoleError() {
+  beforeEach(() => {
+    sinon.stub(console, "error");
```
WickyNilliams: You shouldn't need to use sinon. Jest also has it's own mocking functionality, see https://facebook.github.io/jest/docs/en/mock-functions.html#content

src/components/Title/title.test.js
samiwel: No reason other than it was what was used in [the example I followed](https://facebook.github.io/jest/blog/2016/07/27/jest-14.html). I'll look at changing this to use enzyme.

samiwel: Had to add another dependency `enzyme-to-json` for the snapshot testing to work with enzyme. But got rid of sinon and react-test-renderer in the process.

samiwel:
[Storybook (5)](https://trello.com/c/u70TnvYV/11-storybook-5)

PR #120, opened 2017-05-22
Improve jest setup

WickyNilliams:
We want tests to fail if a component under test receives incorrect/missing prop types. This PR causes `console.error` (that React uses to log about prop types) to throw an error, which gets treated as a failure by Jest.

### How to review 
Make a test fail by changing a prop type to something incorrect. e.g. change the type of a prop on `FileUpload` and run tests with `yarn test`. This should cause a test failure


WickyNilliams:
[Improve jest test setup](https://trello.com/c/H8fHznnl/35-improve-jest-test-setup)

package.json
```diff
@@ -96,7 +96,8 @@
       "src/**/*.{js,jsx}"
     ],
     "setupFiles": [
-      "<rootDir>/config/polyfills.js"
+      "<rootDir>/config/polyfills.js",
```
hamishtaplin: Can this stuff be moved out pf package.json altogether? Is there a dotfile or something?

WickyNilliams: There is such an option, but it failed horribly when I attempted it. Seems it was having a problem resolving module paths once the config had been extracted.

PR #121, opened 2017-05-23
Backfill tests rich text area

samiwel:
Backfill tests for the rich text area component.
Add story demonstrating different features of the component.

### How to review 
Normal rich text editor functionality is observed.

### DIFF
--- a/src/components/RichTextArea/richtextarea.test.js
+++ b/src/components/RichTextArea/richtextarea.test.js
@@ -1,0 +1,46 @@
+import React from "react";
+import RichTextArea from "components/RichTextArea";
+import { mount } from "enzyme";
+
+describe("rich text area", () => {
+  describe("internal state the component", () => {
+    it("should default its state to the value passed in via props", () => {
+      const initialValue = "Should be in state";
+      const wrapper = mount(<RichTextArea value={initialValue} />);
+      expect(wrapper.state().value._cache.html).toEqual(initialValue);
+    });
+
+    it("should update its state if the value prop changes", () => {
+      const initialValue = "Initial value";
+      let wrapper = mount(<RichTextArea value={initialValue} />);
+
+      const newValue = "New value";
+      wrapper = mount(<RichTextArea value={newValue} />);
+
+      expect(wrapper.state().value._cache.html).toEqual(newValue);
+    });
+  });
+
+  describe("toolbar configuration", () => {
+    it("should have Bold button", () => {
+      const wrapper = mount(<RichTextArea />);
+      expect(wrapper.find("[title='Bold']").length).toEqual(1);
+    });
+
+    it("should have Italic button", () => {
+      const wrapper = mount(<RichTextArea />);
+      expect(wrapper.find("[title='Italic']").length).toEqual(1);
+    });
+
+    it("should have UL button", () => {
+      const wrapper = mount(<RichTextArea />);
+      expect(wrapper.find("[title='UL']").length).toEqual(1);
+    });
+
+    it("should have OL button", () => {
+      const wrapper = mount(<RichTextArea />);
+      expect(wrapper.find("[title='OL']").length).toEqual(1);
+    });
+  });
+
+});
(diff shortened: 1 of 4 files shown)


src/components/RichTextArea/richtextarea.test.js
```diff
@@ -0,0 +1,54 @@
...
+      let wrapper = mount(<RichTextArea value={initialValue} />);
+
+      const newValue = "New value";
+      wrapper = mount(<RichTextArea value={newValue} />);
+
+      expect(wrapper.state().value._cache.html).toEqual(newValue);
+    });
+  });
+
+  describe("toolbar configuration", () => {
```
WickyNilliams: Are these needed? Seems to (at least partially) be covered by the snapshot. You can see this in the `toolbarConfig` of the snapshot

src/components/RichTextArea/richtextarea.test.js
```diff
@@ -0,0 +1,54 @@
...
+describe("rich text area", () => {
+  describe("proptypes", () => {
+    it("should require onChange prop", () => {
+      const rta = shallow(<RichTextArea />);
+      expect(toJson(rta)).toMatchSnapshot();
+    });
+  });
+
+  describe("change state", () => {
+    it("internal state value should default to prop ", () => {
```
WickyNilliams: Can you rename these tests so they read better? "it internal state..." doesn't work very well :) "it should default internal state to prop" is better

src/components/RichTextArea/richtextarea.test.js
samiwel: Yeah you're probably right. I could remove these, I guess. But on the other hand, this particular component is proving a bit tricky to Snapshot test. There seems to be some auto-generated values in the Tree which changes all the time. I'm not sure if this is due to the internal state of the component or the Wrapper component which surrounds the rich text area. So rather than remove these tests, I wonder if it would be better to ditch the Snapshot test for this particular component?

src/components/RichTextArea/richtextarea.test.js
```diff
@@ -0,0 +1,54 @@
+import React from "react";
+import RichTextArea from "components/RichTextArea";
+import { shallow, mount } from "enzyme";
+import toJson from "enzyme-to-json";
+
+describe("rich text area", () => {
+  describe("proptypes", () => {
+    it("should require onChange prop", () => {
```
hamishtaplin: Not sure how this offers anything more than using proptypes?

src/components/RichTextArea/richtextarea.test.js
```diff
@@ -0,0 +1,54 @@
...
+    });
+
+    it("changing value prop should update the state", () => {
+      const initialValue = "Initial value";
+      let wrapper = mount(<RichTextArea value={initialValue} />);
+
+      const newValue = "New value";
+      wrapper = mount(<RichTextArea value={newValue} />);
+
+      expect(wrapper.state().value._cache.html).toEqual(newValue);
```
hamishtaplin: `_cache` seems like a private value, is this something we should be accessing?

src/components/RichTextArea/richtextarea.test.js
WickyNilliams: Yeah, with my changes in #120, this kind of thing will cause test failures so no need to explicitly test

src/components/RichTextArea/richtextarea.test.js
WickyNilliams: Possibly ditch the snapshot test for now, yeah?

samiwel: Ditched

src/components/RichTextArea/richtextarea.test.js
samiwel: Discussed with Nick, agreed to leave tests as is for now. They require fixing but the approach to how we're doing forms in general needs to be changed, which might help when re-working this rich text editor component.

WickyNilliams:
Noticed there's no codedcov on this PR...

samiwel:
[Storybook (5)](https://trello.com/c/u70TnvYV/11-storybook-5)

PR #131, opened 2017-06-06
36 New header

hamishtaplin:

[Trello Card](https://trello.com/c/YoljB4Oe/36-new-header-5)

- changed header component to latest design
- renamed from `TopBar`
- introduced `IconButton` component to render an icon as a button with (browser default) tooltip 
- integrated Breadcrumbs with Header and routing
- added some tests

### How to review 

- ensure header matches design


src/components/IconButton/index.js
```diff
@@ -0,0 +1,34 @@
...
+    opacity: 1;
+  }
+  &[disabled] {
+    opacity: 0.5;
+  }
+`;
+
+const IconButton = ({ icon, title, disabled, handleClick }) =>
+  <Button title={title} onClick={handleClick} disabled={disabled}>
+    <img src={require(`./icon-${icon}.svg`)} alt={title} />
```
WickyNilliams: This (dynamic imports) seems like it shouldn't work.

tests/spec/smoketest-spec.js
```diff
@@ -22,9 +22,9 @@ describe("eQ Author", () => {
 
     it("presents user with buttons to create or load survey", () => {
       const buttons = findPageElements("button");
-      expect(buttons.length).toEqual(2);
-      expect(getElementText(buttons[0])).toEqual("Create survey");
-      expect(getElementText(buttons[1])).toEqual("Load survey");
+      expect(buttons.length).toEqual(4);
```
WickyNilliams: Perhaps this test should be updated to be more specific, rather than updated to include the extra buttons?

src/components/IconButton/index.js
hamishtaplin: There's nothing wrong with dynamic requires, afaik. What problems do you foresee? 

The proptypes should catch any attempts to use a file that's not there.

Can you see a better of doing this? I think importing the icon should be a concern of the component so I wouldn't want to pass it in as a child.

WickyNilliams: Hmm... seems that [webpack 1 supported it](http://webpack.github.io/docs/context.html#dynamic-requires), but it looks like it might be getting [deprecated in webpack 2](https://webpack.js.org/guides/migrating/#full-dynamic-requires-now-fail-by-default).

Dynamic imports are a code smell IMO. You could instead statically import the two icon SVGs at the top of the file?

hamishtaplin: Not sure I agree they are a code smell. It's a feature of webpack 2, https://webpack.js.org/guides/migrating/#dynamic-expressions

WickyNilliams: That's for code splitting

hamishtaplin: You're right, doesn't work in this case.

src/components/IconButton/index.js
```diff
@@ -21,11 +28,11 @@ export const Button = styled.button`
 
 const IconButton = ({ icon, title, disabled, handleClick }) =>
   <Button title={title} onClick={handleClick} disabled={disabled}>
-    <img src={require(`./icon-${icon}.svg`)} alt={title} />
+    <img src={icon} alt={title} />
   </Button>;
 
 IconButton.propTypes = {
-  icon: PropTypes.oneOf(["export", "preview"]),
+  icon: PropTypes.oneOf(values(icons)),
```
WickyNilliams: Is values correct? I'd have thought you want `keys()`. Doesn't seem right to pass in the whole `<svg>...</svg>` as a prop, unless you're allowing arbitrary SVG to be passed

hamishtaplin:
[New Header (5)](https://trello.com/c/YoljB4Oe/36-new-header-5)

config/webpack.config.dev.js
```diff
@@ -33,6 +33,7 @@ module.exports = {
   // This means they will be the "root" imports that are included in JS bundle.
   // The first two entry points enable "hot" CSS and auto-refreshes for JS.
   entry: [
+    require.resolve("react-hot-loader/patch"),
```
samiwel: What does this do?

src/components/Breadcrumb/breadcrumb.test.js
```diff
@@ -18,35 +16,14 @@ describe("Breadcrumb", () => {
...
-        {threeLinks}
-      </Breadcrumb>
-    );
+      </Breadcrumb>;
 
     it("should display links", () => {
       const wrapper = mount(<BreadcrumbWithTwoLinks />);
       expect(wrapper.find("a")).toHaveLength(2);
     });
-
```
samiwel: Rather than remove these altogether, should they be re-worked to look for the chevron <svg> tags instead?

src/components/Breadcrumb/index.js
```diff
@@ -2,47 +2,66 @@ import React from "react";
...
 
-const BreadcrumbLinkPropType = PropTypes.oneOfType([
+const BreadcrumbList = styled.ul`
+  list-style: none;
+  display: flex;
+  margin: 0;
+  padding: 0;
+`;
+
+const BreadcrumbItem = styled.li`
```
samiwel: Nice tidy up but will not work if Breadcrumb eventually goes > 2 levels deep.

src/components/Header/index.js
```diff
@@ -0,0 +1,79 @@
...
+  text-decoration: none;
+`;
+
+export const UtilityArea = styled(ButtonGroup)`
+  justify-content: flex-end;
+`;
+
+export const breadcrumbLinks = (path, routes) => {
+  const links = [<Link to="/" key="home">Home</Link>];
+  path !== "/" &&
```
WickyNilliams: if statement?

src/components/Breadcrumb/index.js
```diff
@@ -2,47 +2,66 @@ import React from "react";
 import PropTypes from "prop-types";
 import styled from "styled-components";
 import { colors } from "constants/theme";
-import Chevron from "./chevron";
+import iconChevron from "./chevron.svg";
 
-const BreadcrumbLinkPropType = PropTypes.oneOfType([
+const BreadcrumbList = styled.ul`
```
WickyNilliams: Technically breadcrumbs are an ordered list

src/components/Breadcrumb/index.js
```diff
@@ -2,47 +2,66 @@ import React from "react";
...
+  margin: 0;
+  padding: 0;
+`;
+
+const BreadcrumbItem = styled.li`
+  margin: 0;
+  &:last-child:not(:only-child) {
+    &:before {
+      margin: 0 0.5em;
+      content: url('${iconChevron}');
```
WickyNilliams: Whoa, didn't realise you could set an image as content! I'd always set content to `""` then used background-image.

src/components/Header/index.js
```diff
@@ -0,0 +1,79 @@
...
+  color: white;
+  position: relative;
+  text-decoration: none;
+`;
+
+export const UtilityArea = styled(ButtonGroup)`
+  justify-content: flex-end;
+`;
+
+export const breadcrumbLinks = (path, routes) => {
```
WickyNilliams: Wasn't the requirement of the breadcrumb to show the title of the survey, not the title of the page?

samiwel:
Should the breadcrumb link be changing here? I think it should just display the Survey title or none? Or is this out of scope for this PR?

breadcrumb-change

config/webpack.config.dev.js
hamishtaplin: makes hot reloading work again

src/components/Breadcrumb/breadcrumb.test.js
hamishtaplin: I've moved the chevron into CSS to simplify things a bit—we can't really test that though (not sure it's even necessary now)

src/components/Breadcrumb/index.js
hamishtaplin: That's the intention, we're only going one level deep for now

src/components/Header/index.js
hamishtaplin: You're right, I didn't look at the AC of the breadcrumbs story—will fix

src/components/Header/index.js
hamishtaplin: Got into the habit of using `&&`—why an `if`?

src/components/Breadcrumb/index.js
hamishtaplin: Good spot 👍

src/components/Header/index.js
WickyNilliams: because `&&` is a workaround for inline conditional components, you're pushing to an array here, so makes sense to use the right tool for the job?

hamishtaplin: I was assumed it was a shorthand like `const = mightBeUndefined || 0`

src/components/Breadcrumb/breadcrumb.test.js
samiwel: Okay, fair enough

src/components/Header/index.js
WickyNilliams: yeah, it can work like that, but should always use an `if` if possible. Clearer and easier to read/scan the code

src/components/Breadcrumb/index.js
```diff
@@ -1,48 +1,52 @@
...
+        )}
+    </BreadcrumbList>
   </StyledBreadcrumb>;
 
 Breadcrumb.propTypes = {
-  children: PropTypes.arrayOf(BreadcrumbLinkPropType).isRequired
+  breadcrumbs: PropTypes.arrayOf(PropTypes.object).isRequired
 };
 
 export default Breadcrumb;
```
WickyNilliams: Generally, for Breadcrumb, what about ARIA? 

See: http://w3c.github.io/aria-practices/examples/breadcrumb/

src/components/Header/Header.test.js
```diff
@@ -0,0 +1,40 @@
+import React from "react";
+import toJson from "enzyme-to-json";
+import { shallow } from "enzyme";
+
+import { Header, UtilityBtns, Logo } from "components/Header";
+import Breadcrumb from "containers/Breadcrumb";
+
+describe("components/Header", function() {
+  it("renders correctly ", function() {
+    const wrapper = shallow(<Header />);
```
WickyNilliams: Move this repeated code to `beforeEach` perhaps?

src/components/Header/logo.svg
```diff
@@ -0,0 +1,9 @@
...
+            <path d="M8.07616546,4.08203997 L18.6913768,4.08203997 L18.6913768,6.99313385 L8.07616546,6.99313385 L8.07616546,4.08203997 Z M2.17645481,4.08203997 L5.53183771,4.08203997 L5.53183771,6.99313385 L2.17645481,6.99313385 L2.17645481,4.08203997 Z M8.07616546,13.5992319 L18.6913768,13.5992319 L18.6913768,16.5103258 L8.07616546,16.5103258 L8.07616546,13.5992319 Z M2.17645481,13.5992319 L5.53183771,13.5992319 L5.53183771,16.5103258 L2.17645481,16.5103258 L2.17645481,13.5992319 Z M0.0734467921,8.84063593 L10.6886581,8.84063593 L10.6886581,11.7517298 L0.0734467921,11.7517298 L0.0734467921,8.84063593 Z M13.2969362,8.84063593 L16.6523191,8.84063593 L16.6523191,11.7517298 L13.2969362,11.7517298 L13.2969362,8.84063593 Z" id="Combined-Shape" fill="#9fa6ab"></path>
+        </g>
+    </g>
+</svg>
```
WickyNilliams: Might be worth running this through svgo (webapp: https://jakearchibald.github.io/svgomg/) to see if/how much it can be optimised

src/components/IconButton/IconButton.test.js
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+import { shallow } from "enzyme";
+
+import IconButton, { icons } from "components/IconButton";
+
+describe("components/IconButton", function() {
+  const mockFn = jest.fn();
+  const iconButton = shallow(
```
WickyNilliams: this should be in `beforeEach`

src/components/IconButton/icon-export.svg
```diff
@@ -0,0 +1,13 @@
...
+    <title>icon-export</title>
+    <desc>Created with Sketch.</desc>
+    <defs></defs>
+    <g id="Page-1" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g id="icon-export" transform="translate(-5.000000, -4.000000)">
+            <path d="M19,10 L15,10 L15,4 L9,4 L9,10 L5,10 L12,17 L19,10 Z M5,19 L5,21 L19,21 L19,19 L5,19 Z" id="Shape" fill="#FFFFFF" fill-rule="nonzero"></path>
+            <polygon id="Shape" points="0 0 24 0 24 24 0 24"></polygon>
+        </g>
+    </g>
+</svg>
```
WickyNilliams: svgomg again

src/routes.js
```diff
@@ -1,15 +1,18 @@
...
+
 import SurveyPage from "containers/Survey";
 import CreateSurvey from "containers/CreateSurvey";
 import DesignSurvey from "containers/DesignSurvey";
 import NotFound from "pages/NotFound";
 
 import { SidebarPageLayout, TabbedPageLayout, FullPageLayout } from "layouts";
 
-const routes = [
+export const routes = [
```
WickyNilliams: why export as named _and_ default?

src/components/IconButton/index.js
```diff
@@ -0,0 +1,41 @@
...
+  }
+`;
+
+const IconButton = ({ icon, title, disabled, handleClick }) =>
+  <Button title={title} onClick={handleClick} disabled={disabled}>
+    <img src={icon} alt={title} />
+  </Button>;
+
+IconButton.propTypes = {
+  icon: PropTypes.oneOf(values(icons)),
```
WickyNilliams: why use values instead of keys? you're essentially passing the entire SVG markup as a prop, which is fine if you allow arbitrary SVGs to be passed. But we don't, so may as well use a key?

src/containers/App.js
```diff
@@ -7,30 +7,28 @@ import { Switch } from "react-router-dom";
 import { Route } from "react-router";
 import routes from "routes";
 
-const renderRoute = route => props => (
-  <route.layout title={route.title}>
+const renderRoute = route => props =>
+  <route.layout title={route.title} nav={route.nav}>
```
WickyNilliams: `route.nav` never seems to be set anywhere?

src/containers/App.js
```diff
@@ -7,30 +7,28 @@ import { Switch } from "react-router-dom";
...
-const App = ({ store, history }) => (
+const App = ({ store, history }) =>
   <AppContainer>
     <Provider store={store}>
       <ConnectedRouter history={history}>
         <Switch>
-          {routes.map((route, i) => (
+          {routes.map((route, i) =>
             <Route
               exact
```
WickyNilliams: `routes.js` sets `exact` field, but it's never used. Should the field be removed from `routes.js`, or to to the `Route` component?

src/components/Header/logo.svg
hamishtaplin: Agreed but I think this should be done as part of the webpack task rather than a manual process

src/routes.js
hamishtaplin: I find it just cuts down on the number of mistakes made around importing modules. This is done in many places in the app—if we're going to enforce a style around exports we'll need to lint for it.

src/containers/App.js
hamishtaplin: it used to be passed to the route component, I'll just wire it back up

src/containers/App.js
hamishtaplin: dead code, will remove

src/routes.js
hamishtaplin: Created a task to review this https://trello.com/c/W3P7ZT8k/77-linting-rules-around-import-export

src/components/Header/logo.svg
hamishtaplin: Task for this https://trello.com/c/ZgxANoSH/76-svgo

PR #132, opened 2017-06-06
Make theme colours more consistent

hamishtaplin:

- Made how colours work more consistent so all the variable names describe the colour, not their implementation.
- Moved the colours into their own option
- Prefer import of constants over the theming mechanics of styled-components

#### Bonus content

- Renamed some files

### How to review 

- Ensure all colours still being used


src/components/AddButton/story.js
```diff
@@ -2,10 +2,10 @@ import React from "react";
 import { storiesOf } from "@kadira/storybook";
 import AddButton from "components/AddButton";
 import styled from "styled-components";
-import { colorDarkBlue } from "constants/theme";
+import { darkBlue } from "constants/theme/colours";
```
hamishtaplin: The thing is that when you name them according to specific use-cases such as `sidebarBackgroundColor` it becomes weird when you use those colours elsewhere. For example, the orange used for the beta tag might be used in a warning component. If they're only being used for one thing then there's no point in using variables in the first place.

Agreed that it's difficult to scale the light/dark convention but, in my experience, you'll only have a few shades of each colour and any more becomes a design smell (should be using a limited palette). 

One alternative is to use a more abstract naming convention such as "slate", "pavement", "cloud" for shades of grey, eg. but it's hard to do that for shades of blue.

Happy to take suggestions though 👍

samiwel: Yeah I see your point there would be some potential duplication of colour values. But that could be fixed with some variables in the theme file e.g. 
```
const colors =  {
  orange: "...",
  blue: "..."
}

const theme = {
  header: {
    background: colors.blue
  },
  title: {
     betaText: colors.orange
  }
}
```

hamishtaplin: Not sure what the value of declaring the colour values of components in the theme is? There's no requirement for theming so far—it's just a way of defining a colour palette and making the values available in components.

samiwel: The benefit being that you decouple the colours from where they are used. i.e. you can just use `theme.header.background` in the styled component. The added benefit being that you can replace `theme` for another theme down the line. But maybe it's overkill for now, like you say.

hamishtaplin: Happy to cross that bridge if we come to it—I don't forsee us having a huge amount of colours in here anyway so it'll be easy enough refactor if it becomes a problem.

WickyNilliams: How many colours are there (ignoring text, white, black etc)? Can you organise them into e.g. primary, secondary, tertiary? Colour palettes are rarely more extensive than that. Additional colours are usually variants like a darkened primary

hamishtaplin: There's only a few colours. In my experience, unless the brand is defined up front in terms of primary, secondary, etc colours then you can't assume they will be used as such. Case in point: I tried to do that early in the survey runner and subsequent designs didn't reflect that system and so it fell apart.

WickyNilliams: Perhaps it's something Dan might be able to help define? In my experience colour names quickly get confusing. But I'm not particularly precious about it. Just a suggestion :)

hamishtaplin: We're risking overthinking what are, essentially, some variable names :)

Happy to take suggestions of specific improvements though. Otherwise, we should probably just go with what is there and adapt as we progress. Half these colours have gone already from the designs.

PR #133, opened 2017-06-06
Re-work of Breadcrumb component to match the new design.

samiwel:
Refactor the breadcrumb component so that it more closely matches the new design.
Mostly stylistic updates, functionally should behave the same as before when an array of links is passed to it.
The breadcrumb defaults to the Dark theme (as per the design). But you can switch it (see the story for an example of a Light theme).
I added some additional Jest tests to cover some of the breadcrumb functionality.
Since the breadcrumb sits within the Header, some additional tweaking may be required to get it looking as expected.

### How to review 
All tests should pass.
Existing behaviour should continue to work.
Style of breadcrumb should match the designs.

### DIFF
--- a/src/components/Breadcrumb/chevron.js
+++ b/src/components/Breadcrumb/chevron.js
@@ -1,37 +1,13 @@
 import React from "react";
-import PropTypes from "prop-types";
 import styled from "styled-components";
+import ChevronSvg from "./chevron.svg";
 
-const ChevronSvg = props => (
-  <svg width="10px" height="12px" viewBox="47 10 5 7" version="1.1">
-    <polyline
-      id="Combined-Shape"
-      stroke={props.chevronColor}
-      strokeWidth="1"
-      fill="none"
-      opacity="0.521338619"
-      transform="translate(49.153996, 13.750000) rotate(-270.000000) translate(-49.153996, -13.750000) "
-      points="51.8986355 15 49.1539961 12.5 46.4093567 15"
-    />
-  </svg>
-);
-
-ChevronSvg.propTypes = {
-  chevronColor: PropTypes.string
-};
-
-ChevronSvg.defaultProps = {
-  chevronColor: "#000"
-};
-
-const StyledSpan = styled.span`
-  margin: 16px;
+const ChevronIcon = styled.span`
+  display: inline-block;
+  width: 1em;
+  height: 1em;
+  padding: 0 1em;
+  background: url('${ChevronSvg}') center center no-repeat;
 `;
 
-export default props => {
-  return (
-    <StyledSpan>
-      <ChevronSvg {...props} />
-    </StyledSpan>
-  );
-};
+export default props => <ChevronIcon {...props} />;
--- a/src/components/Breadcrumb/chevron.js
+++ b/src/components/Breadcrumb/chevron.js
@@ -1,16 +1,37 @@
 import React from "react";
+import PropTypes from "prop-types";
+import styled from "styled-components";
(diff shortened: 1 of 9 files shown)


src/components/Breadcrumb/index.js
```diff
@@ -1,39 +1,34 @@
...
     {props.link}
-    {!props.isLast && <ChevronIcon />}
+    {!props.isLast && <Chevron chevronColor={props.theme.textColor} />}
   </span>
 );
 
 BreadcrumbLink.defaultProps = {
-  isLast: false
+  isLast: false,
+  theme: dark
```
hamishtaplin: Do we need this? I'm not sure it will be used.

src/components/Breadcrumb/index.js
```diff
@@ -1,39 +1,34 @@
...
-  margin: 0 0.5em 0 0.5em;
-`;
-
 const StyledBreadcrumb = styled.div`
-  background-color: ${colorGrey};
-  font-size: 0.7em;
-  display: flex;
-  align-items: center;
-  padding: 1em 1em;
+  font-size: 16px;
```
samiwel: @hamishtaplin The design says 16px, can we set this up somehow so that 1em = 16px?

src/components/Breadcrumb/index.js
samiwel: Not sure, on the design it's white text on dark background. In storybook we need someway of seeing it against a light background if we want to use it in it's raw form.

src/components/Breadcrumb/chevron.js
```diff
@@ -1,16 +1,37 @@
...
-        stroke="#979797"
-        strokeWidth="1"
-        fill="none"
-        opacity="0.521338619"
-        transform="translate(49.153996, 13.750000) rotate(-270.000000) translate(-49.153996, -13.750000) "
-        points="51.8986355 15 49.1539961 12.5 46.4093567 15"
-      />
-    </svg>
-  </span>;
+const ChevronSvg = props => (
```
samiwel: Will take a look at that post, thanks. Do you mean save the SVG's in separate .svg files and read them in?

hamishtaplin: yes 👍

src/components/Breadcrumb/index.js
hamishtaplin: This should already be the case as the base font-size is `16px` unless a parent element has redeclared the font-size (it cascades)

src/components/Breadcrumb/index.js
hamishtaplin: I think that problem should be solved in the story. You can just wrap the component in an element with a dark background.

src/components/Breadcrumb/index.js
samiwel: Ah, okay I see, I'll look into that.

src/components/Breadcrumb/index.js
samiwel: Fair enough, I'll remove the theme stuff for now. Less is better 👍

WickyNilliams:
I think we should change the breadcrumb component, so that instead of taking a `links` prop, it should receive children:

```js
<Breadcrumb>
    <Link href="#">Text 1</Link>
    <Link href="#">Text 2</Link>
</Breadcrumb>
```

samiwel:
[Breadcrumbs (2)](https://trello.com/c/PllHhYtG/37-breadcrumbs-2)

PR #134, opened 2017-06-07
Tab styles

bennpearson:
Created new Nav component and added styling to Tabs component

### How to review 
Check Nav component 
Check Nav tests
Check Tab styling
Check storybook for both components

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,5 +17,3 @@ yarn-error.log*
 
 /storybook-static
 .vscode/
-
-/**/*/__snapshots__
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ yarn-error.log*
 
 /storybook-static
 .vscode/
+
+/**/*/__snapshots__
--- a/src/components/Nav/index.js
+++ b/src/components/Nav/index.js
@@ -1,0 +1,59 @@
+import React from "react";
+import styled from "styled-components";
+import { Column } from "components/Grid";
+import { NavLink } from "react-router-dom";
+import { colorBlue, colorText } from "constants/theme";
+
+const StyledGrid = styled.div`
+  display: flex;
+  width: 100%;
+  flex: 0 1 auto;
+  border-bottom: 1px solid ${props => props.theme.colorBorders};
+`;
+
+const StyledNav = styled.nav`
+  margin: 0;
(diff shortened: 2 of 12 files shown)


samiwel:
@bennpearson looks like some of your snapshots are out of date. Run `yarn test` locally then select `u` to update snapshots. Commit and push 👍

.gitignore
```diff
@@ -16,4 +16,6 @@ yarn-debug.log*
 yarn-error.log*
 
 /storybook-static
-.vscode/
\ No newline at end of file
+.vscode/
+
+/**/*/__snapshots__
```
samiwel: Not sure we want to ignore snapshots. It would defeat the point of snapshot testing. I'd suggest either updating the snapshot as described in my comment above, or removing snapshot test for the Nav component if you feel it isn't add any value.

src/components/Nav/index.js
```diff
@@ -0,0 +1,60 @@
+import React from "react";
+import styled from "styled-components";
+import { Column } from "components/Grid";
+import { NavLink } from "react-router-dom";
+import { colorBlue, colorText } from "constants/theme";
+
+const StyledGrid = styled.div`
```
hamishtaplin: We don't need to prefix everything with `Styled`—just the cases where it distinguishes a styled component from the main component (as you have with the `Nav`/`StyledNav`.

hamishtaplin:
@bennpearson why have you removed the snapshots?

bennpearson:
@hamishtaplin added the snapshot test back and renamed StyledGrid to just Grid 👍

src/components/Nav/nav.test.js
```diff
@@ -0,0 +1,15 @@
...
+import { mount, shallow } from "enzyme";
+import Nav, { StyledNav } from "components/Nav";
+
+describe("components/Nav", function() {
+  it("should render Nav", function() {
+    const wrapper = shallow(<Nav />);
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it("should contain links", function() {
```
hamishtaplin: This isn't really testing anything that the snapshot won't cover (ie. it renders some links). For presentational components with no logic, it's probably not worth writing anything other than snapshot tests tbh.

PR #135, opened 2017-06-08
Rename survey to questionnaire across the app.

samiwel:
This PR refactors the Author application to keep the nomenclature consistent.
References to Survey have now been replaced by references to Questionnaire.

### How to review 
All tests should continue to pass.
Application should run as normal and functionality should remain the same.
All references to survey in the Author application should have been replaced (excluding any data that was in the JSON schema file itself).


samiwel:
[Rename 'Survey' to 'Questionnaire'](https://trello.com/c/OlS7vUdm/54-rename-survey-to-questionnaire)

samiwel:
@WickyNilliams Thanks for the tip. I was kind of hoping that Webstorm would have done a `git mv` when using it's built in refactoring tools, guess not!

hamishtaplin:
@samiwel @WickyNilliams is it worth doing that way?

WickyNilliams:
@samiwel git can detect the move if the file itself doesn't change much. But when a move is combined with a large change, git can't infer it's the same file.

@hamishtaplin maybe... but could be a lot of work to redo the PR?

PR #140, opened 2017-06-14
Create new Questionnaire (integration)

hamishtaplin:
## Changes

Added clientside logic for connecting to Author API:

- Integrated Apollo Client
- Removed questionnaire `meta` from Redux
- Moved update logic to form input/textarea/select components
- Connected frontend to Apollo backend
- Updated breadcrumbs to work with backend
- Added tests

## Testing

- Check process of adding a Questionnaire works and is stored in backend


src/pages/CreateQuestionnaire.js
```diff
@@ -21,76 +23,105 @@ const ActionButtonGroup = styled(ButtonGroup)`
...
-    mime_type: PropTypes.string,
-    questionnaire_id: PropTypes.string,
-    schema_version: PropTypes.string,
-    theme: PropTypes.string,
-    title: PropTypes.string
-  }),
-  onChange: PropTypes.func
+  history: PropTypes.object.isRequired, // eslint-disable-line
+  loading: PropTypes.bool.isRequired,
+  onUpdate: PropTypes.func.isRequired,
```
bennpearson: do we need `error` propType here?

hamishtaplin: @bennpearson currently not dealing with errors, probably need to at some point though!

src/components/Forms/Input/index.js
```diff
@@ -26,28 +26,44 @@ const StyledInput = styled.input`
...
+  type: "text",
+  handleChange: noop
 };
 
 Input.propTypes = {
+  handleChange: PropTypes.func,
+  id: PropTypes.string,
   type: PropTypes.oneOf(["text", "checkbox", "radio"]).isRequired,
-  value: PropTypes.oneOfType([
+  defaultValue: PropTypes.oneOfType([
```
WickyNilliams: why `defaultValue` is you're now using `onChange` handlers?

src/containers/Breadcrumb.js
```diff
@@ -1,24 +1,48 @@
...
 import Breadcrumb from "components/Breadcrumb";
 import { getRouteByPath } from "routes";
 
-export const mapStateToProps = (state, _) => {
-  const { title } = state.questionnaire.meta;
-  const { pathname } = state.router.location;
-  const route = getRouteByPath(pathname);
+const getQuestionnaire = gql`
+  query GetQuestionnaire {
+    questionnaire(id: 1) {
```
WickyNilliams: this should be paramterised

WickyNilliams: Why is the breadcrumb requesting this data? Seems like an odd place for the request to be made. Especially when CreateQuestionnaire seems to do the same. Seems like the type of thing that should flow down through props

src/containers/CreateQuestionnaire.js
```diff
@@ -1,26 +1,68 @@
...
+export const getQuestionnaire = gql`
+  query GetQuestionnaire($id: ID!) {
+    questionnaire(id: $id) {
+      id,
+      title,
+      description,
+      navigation,
+      legalBasis,
+      theme,
+      __typename
```
WickyNilliams: I don't think you need this in your query. Is there a reason for it?

src/containers/CreateQuestionnaire.js
```diff
@@ -1,26 +1,68 @@
...
+      description,
+      theme,
+      legalBasis,
+      navigation
     }
+  }
+`;
+
+const mapResultsToProps = results => {
+  const { data: { loading, questionnaire } } = results;
```
WickyNilliams: if you're going to do deep destructuring, what do you think about formatting it so it mirrors the structure of the object? e.g.

```js
const {
    data : {
       loading,
       questionnaire
  }
} = results;
```

alternatively:

```js
const { loading, questionnaire } = results.data;
```

src/containers/CreateQuestionnaire.js
```diff
@@ -1,26 +1,68 @@
...
+  // console.log(results.data.error);
```
WickyNilliams: remove pls :)

src/containers/CreateQuestionnaire.js
```diff
@@ -1,26 +1,68 @@
...
+    loading
   };
 };
 
-export default connect(mapStateToProps, mapDispatchToProps)(
-  CreateQuestionnairePage
-);
+export const withData = graphql(getQuestionnaire, {
+  props: mapResultsToProps,
+  options: { variables: { id: 1 } }
```
WickyNilliams: shouldn't have the `id` hard-coded

src/pages/CreateQuestionnaire.js
```diff
@@ -21,76 +23,105 @@ const ActionButtonGroup = styled(ButtonGroup)`
...
 
-const CreateQuestionnairePage = ({ meta, onChange }) => {
-  const { title, description, theme, legal_basis } = meta;
-  return (
-    <Center>
-      <form onChange={onChange}>
+export class CreateQuestionnairePage extends Component {
+  constructor(props) {
+    super(props);
+    this.debouncedChangeHandler = debounce(this.onChange.bind(this), 300, {
```
WickyNilliams: surprised this needs binding if you're using a property initializer below, since that should already be bound to `this`

src/pages/CreateQuestionnaire.js
```diff
@@ -21,76 +23,105 @@ const ActionButtonGroup = styled(ButtonGroup)`
...
-            <Input type="checkbox" />
-            <Label inline>Navigation</Label>
-          </Field>
-          <ActionButtonGroup horizontal>
-            <LinkButton to="/design" primary>Next</LinkButton>
-            <LinkButton to="/" secondary>Cancel</LinkButton>
-          </ActionButtonGroup>
+          {loading
+            ? <div>Loading...</div>
+            : <Form handleSubmit={this.onSubmit}>
```
WickyNilliams: Can you extract the two branches into methods? Much clearer to read `loading ? this.renderLoadingIndictator() : this.renderForm()` than a huge block of JSX

src/reducers/index.js
```diff
@@ -1,11 +1,12 @@
...
 import questionnaire from "reducers/questionnaire/";
 
-const rootReducer = combineReducers({
-  questionnaire,
-  router: routerReducer
-});
+const createReducer = reducers =>
+  combineReducers({
+    questionnaire,
+    router: routerReducer,
```
WickyNilliams: could always put `{ routerReducer as router }` in your import statement if you want to consistently use shorthand properties

src/apollo/client.js
```diff
@@ -0,0 +1,18 @@
...
+  dataIdFromObject: result => {
+    if (result.id && result.__typename) {
+      // eslint-disable-line no-underscore-dangle
+      return result.__typename + result.id;
+      // eslint-disable-line no-underscore-dangle
+    }
+    return null;
+  },
+  networkInterface: createNetworkInterface({
+    uri: "http://localhost:4000/graphiql"
```
samiwel: This should probably be an environment variable.

src/components/Forms/Input/index.js
hamishtaplin: If we use `value` then you're depending on a debounced roundtrip to the api to update the `value`—makes for a rather laggy typing experience.

Using `defaultValue` sets the value on initial render but then the state of the value is handled internally, with updates being periodically sent to the API.

src/containers/Breadcrumb.js
hamishtaplin: I wasn't entirely happy with this but the alternative is a significant refactor. Structurally, `CreateQuestionnaire` and the `Breadcrumbs` are in different timezones, the data fetching would have to be done at layout level and then the props passed down through several layers of components, which comes with its own problems.

src/containers/CreateQuestionnaire.js
hamishtaplin: This was added whilst attempting to get `MockedProvider` to work in the tests. See: https://github.com/apollographql/react-apollo/issues/674#issuecomment-298496730

src/containers/CreateQuestionnaire.js
hamishtaplin: Prefer the latter 👍

src/containers/CreateQuestionnaire.js
hamishtaplin: Oops

src/containers/CreateQuestionnaire.js
hamishtaplin: Where should this live?

src/pages/CreateQuestionnaire.js
hamishtaplin: I was also surprised but it doesn't work otherwise, I think related to [this answer on SO](https://stackoverflow.com/questions/23123138/perform-debounce-in-react-js#answer-28046731).

src/pages/CreateQuestionnaire.js
hamishtaplin: This just moves the huge chunk of JSX problem elsewhere, I've done it anyway though :)

src/reducers/index.js
hamishtaplin: 👍

src/apollo/client.js
hamishtaplin: It bloody should be 🎉

src/containers/CreateQuestionnaire.js
samiwel: Can you create a function called withData which accepts a parameter (the Id) and returns the graphql? e.g. something like:

```
const withData = questionnaireId => {
  return graphql(getQuestionnaire, {
    props: mapResultsToProps,
    options: { variables: { id: questionnaireId } }
  });
}
```

Then call with `withData(1)`

src/containers/Breadcrumb.js
samiwel: Might be worth revisiting the architecture at some point. Particularly at which "level" we use container components. The Layout/Page hierarchy was something I threw together to get it working initially. If it is causing problems then we can revisit it. But it's probably outside the scope of this PR.

samiwel: I paramterised the gql on Publisher, see [here](https://github.com/ONSdigital/eq-publisher/blob/master/src/api/GraphQLApi.js#L20) for an example.

src/components/Forms/Input/index.js
WickyNilliams: We shouldn't be saving as a user types. The value of the form shouldn't be tied to a request to the server. It should go into the store, then pulled out when necessary. 
In any case, saving as a user types opens up corner cases like rollback happening as user types - would actually cause them to lose work!

I personally think we should do auto-saving as a separate story rather than ad-hoc, as that tends to result in a better solution (e.g. perhaps it can be done cleanly with middleware?)

src/containers/Breadcrumb.js
WickyNilliams: The value goes into the store, presumably? So couldn't Breadcrumb be connected to store and pull the value out of there?

hamishtaplin: I think we should refactor _if_ it becomes an issue. At the moment it's not causing any problems.

src/containers/CreateQuestionnaire.js
WickyNilliams: Did it make it work? If not, can it be removed?

WickyNilliams: Ignore, I see you removed it

src/containers/CreateQuestionnaire.js
WickyNilliams: The id should come from the url i guess?

hamishtaplin: This just moves the hardcoded value elsewhere?

src/pages/CreateQuestionnaire.js
WickyNilliams: Hmm, that answer suggested this should work:

```js
   onChange = debounce(value => {
      this.setState(value, () => {
        this.props.onUpdate(this.state);
      });
    });
```

src/containers/Breadcrumb.js
hamishtaplin: @samiwel I think we should refactor if it becomes an issue. At the moment it's not causing any problems.

src/pages/CreateQuestionnaire.js
WickyNilliams: Haha fair point :) But a huge chunk of JSX isn't a problem in itself, it's just not great hanging off a ternary operator.

src/containers/Breadcrumb.js
hamishtaplin: @WickyNilliams We don't want to be pulling graphQL data (the title) from the store? This was discussed previously.

src/containers/CreateQuestionnaire.js
hamishtaplin: @WickyNilliams That would work but in the interests of moving PRs on a bit quicker, I think it's better that we solve problems (such as being able to dynamically select a questionnaire) when we need that feature.

src/components/Forms/Input/index.js
hamishtaplin: Can we address this as a separate "saving/auto-saving" story when we have some more UI to work with?

src/pages/CreateQuestionnaire.js
hamishtaplin: @WickyNilliams that was the first thing I did, it doesn't work. :(

src/containers/Breadcrumb.js
WickyNilliams: Hmm yeah. But that was reading directly from the store and being dependent on the keying strategy used. Though there are methods for pulling stuff from the store without hitting the server: http://dev.apollodata.com/core/read-and-write.html. `readQuery` or `readFragment` are what we want, I think

src/pages/CreateQuestionnaire.js
WickyNilliams: Hmm, screw it then. Leave as-is for now!

src/components/Forms/Input/index.js
WickyNilliams: I think that's wise. How about changing to autosave on blur for now? Conceptually simpler and less corner cases

src/containers/CreateQuestionnaire.js
WickyNilliams: The URL stuff we could do later as that's a bigger task. But for now, you could pass the id back from the component?

src/components/Forms/Input/index.js
hamishtaplin: 👍

src/containers/Breadcrumb.js
hamishtaplin: Wasn't aware of that, I'll look into it. However, does't that rely on the request having been made elsewhere? Eg. if you navigate straight to the `/design` page there's no request currently so the breadcrumbs won't have clue what's going on.

src/containers/CreateQuestionnaire.js
hamishtaplin: That's still just hardcoding the value elsewhere, solving a problem we don't have yet. We likely won't be passing that id in from the component in future so it will require another refactor further down the line.

src/containers/Breadcrumb.js
WickyNilliams: Yes it would rely on that. Haha, guess this goes back to the `id` in the url issue. OK let's leave for now. Do you want to add some tasks to trello to cover all the stuff we're going to defer for now?

hamishtaplin: Looking ahead—the Header/Nav will probably need to be more connected. I think that will present opportunities to connect higher up and pass things like title down to Breadcrumbs.

src/containers/CreateQuestionnaire.js
WickyNilliams: I mean it should receive as prop and pass back? Or the callback could grab the value from somewhere before calling mutate. Either way, I absolutely think we be shouldn't be hardcoding values now

src/containers/Breadcrumb.js
WickyNilliams: That sounds sensible. I wouldn't be surprised if it eventually ends up connected at page level, with everything flowing down from there

src/pages/CreateQuestionnaire.js
hamishtaplin: Moving to update on `blur` just made this irrelevant anyway.

WickyNilliams: All part of my master plan

dr-evils-laughing

samiwel: "La-zer"

.env.production
```diff
@@ -1 +1,2 @@
 REACT_APP_BASE_NAME="/eq-author"
+REACT_APP_API_URL=""
```
samiwel: Are we able to override the `REACT_APP_API_URL` value with an environment variable. Or do we need to default it to `process.env.REACT_APP_API_URL` here?

We need to be able to customise this for successful deployment to CloudFoundry.

bennpearson:
Too long, can't review this. Might be good to split this up to front frontend and back frontend!

package.json
```diff
@@ -87,11 +88,13 @@
   "dependencies": {
     "connected-react-router": "^4.2.1",
     "enzyme": "^2.8.2",
+    "global": "^4.3.2",
```
WickyNilliams: What's this?

src/apollo/client.js
```diff
@@ -0,0 +1,18 @@
+import { ApolloClient, createNetworkInterface } from "react-apollo";
+
+const client = new ApolloClient({
+  ddTypename: true,
```
WickyNilliams: s/ddTypename/addTypename

src/components/BaseLayout/index.js
```diff
@@ -40,22 +48,25 @@ const BaseLayout = ({ children, hasNav, hasBreadcrumbs, hasUtilityBtns }) =>
...
 };
 
 BaseLayout.defaultProps = {
-  hasNav: false,
-  hasBreadcrumbs: false,
+  hasNav: true,
   hasUtilityBtns: false
 };
 
+BaseLayout.displayName = "BaseLayout";
```
WickyNilliams: What's the need for this?

src/components/Breadcrumb/index.js
```diff
@@ -35,28 +36,25 @@ export const BreadcrumbLink = styled(Link)`
...
-  <BreadcrumbNav aria-label="breadcrumb">
-    <BreadcrumbList>
-      {breadcrumbs.map(({ title, pathname }, index) =>
-        <BreadcrumbItem key={pathname}>
-          <BreadcrumbLink to={pathname}>{title}</BreadcrumbLink>
+const Breadcrumb = ({ breadcrumb }) => {
+  return (
+    <BreadcrumbNav aria-label="breadcrumb">
+      <BreadcrumbList>
+        <BreadcrumbItem key="/">
```
WickyNilliams: Shouldn't need `key` prop, unless elements are in an array.

src/components/Breadcrumb/index.js
```diff
@@ -35,28 +36,25 @@ export const BreadcrumbLink = styled(Link)`
...
-
-Breadcrumb.propTypes = {
-  breadcrumbs: PropTypes.arrayOf(
-    PropTypes.shape({
-      title: PropTypes.string.isRequired,
-      pathname: PropTypes.string.isRequired
-    })
-  ).isRequired
+        <BreadcrumbItem key={breadcrumb.path}>
+          <BreadcrumbLink to={breadcrumb.path}>
```
WickyNilliams: Should have the `aria-current` attribute on the current breadcrumb

src/components/Button/index.js
```diff
@@ -63,21 +63,25 @@ const StyledButton = styled.button`
...
   ${props => props.tertiary && TertiaryButton}
   ${props => props.clear && ClearButton}
 `;
 
-const Button = props =>
-  <StyledButton {...props} type="button">
-    {props.children}
-  </StyledButton>;
+const Button = ({ children, type, ...otherProps }) =>
+  <StyledButton {...otherProps} type={type}>{children}</StyledButton>;
```
WickyNilliams: We should move to styled-components v2 at some point. Has mechanisms for avoiding wrappers like this e.g. https://www.styled-components.com/docs/basics#attaching-additional-props

src/components/Forms/Input/Input.test.js
```diff
@@ -1,20 +1,79 @@
...
   });
 
-  it("should render the correct type", function() {
-    const wrapper = shallow(<Input type="checkbox" />);
-    expect(wrapper.props()).toMatchObject({ type: "checkbox" });
+  describe("Checkbox", () => {
+    let wrapper;
+
+    beforeEach(() => {
+      wrapper = mount(
```
WickyNilliams: why not a checkbox here?

src/components/Forms/Input/index.js
```diff
@@ -26,28 +26,44 @@ const StyledInput = styled.input`
...
     id={id}
     name={id}
-    onChange={noop}
+    onChange={function(e) {
+      const { target } = e;
+      handleChange({
+        [id]: {
+          text: target.value,
+          checkbox: target.checked
+        }[type]
```
WickyNilliams: These lines are hard to read. Could you clean up? Perhaps extract a function to handle this stuff

src/components/Forms/TextArea/index.js
```diff
@@ -10,22 +10,36 @@ const StyledTextArea = styled.textarea`
...
+      handleChange({ [id]: e.target.value });
+    }}
     {...otherProps}
   />;
 
 TextArea.propTypes = {
-  value: PropTypes.string.isRequired,
+  defaultValue: PropTypes.string,
   id: PropTypes.string,
+  handleChange: PropTypes.func,
```
WickyNilliams: onChange should be the name of this kind of prop (here and anywhere else), to align with regular DOM components. handleChange should only be used internally e.g.

```js
class ExampleComponents extends Component {

  handleChange = (e) {
    const value = e.target.value;
    this.props.onChange(value);
  };

  render() {
    return (
      <input
        type="text"
        value={this.props.value}
        onChange={this.handleChange}
      /> 
    );
  }

}
```

src/components/QuestionnaireMeta/index.js
```diff
@@ -0,0 +1,98 @@
...
+  questionnaire,
+  handleSubmit,
+  handleBlur,
+  handleChange,
+  children
+}) => {
+  return (
+    <Center>
+      <Panel>
+        <Form handleSubmit={handleSubmit}>
```
WickyNilliams: as before, this should be onSubmit. onX for external props, handleX for internal method

src/containers/QuestionnaireCreatePage/QuestionnaireCreate.js
```diff
@@ -0,0 +1,69 @@
...
+      legalBasis: "StatisticsOfTradeAct",
+      navigation: false
+    };
+  }
+
+  componentWillReceiveProps({ questionnaire }) {
+    this.setState(questionnaire);
+  }
+
+  onChange = value => this.setState(value);
```
WickyNilliams: as discussed, these should be called handleChange, handleSubmit etc

src/containers/QuestionnaireMetaPage/QuestionnaireMeta.js
```diff
@@ -0,0 +1,73 @@
...
+
+  onChange = value => {
+    this.setState(value);
+  };
+
+  onBlur = e => {
+    this.props.onUpdate(this.state);
+  };
+
+  onSubmit = e => e.preventDefault();
```
WickyNilliams: `handleX`

src/containers/QuestionnaireMetaPage/index.js
```diff
@@ -0,0 +1,74 @@
...
+  props: ({ ownProps, mutate }) => ({
+    onUpdate(variables) {
+      return mutate({ variables });
+    }
+  })
+});
+
+export default compose(connect(mapStateToProps), withData, withMutation)(
+  QuestionnaireMeta
+);
```
WickyNilliams: There is a lot of duplicated code between this and `src/containers/QuestionnaireDesignPage/index.js`. Should extract common queries etc

package.json
hamishtaplin: An accident!

src/components/BaseLayout/index.js
hamishtaplin: A workaround for [this bug](https://github.com/facebook/jest/issues/1824)

src/components/Button/index.js
hamishtaplin: Yeah I saw that 👍

WickyNilliams:
[Create New Q (8)](https://trello.com/c/dcs26jy9/2-create-new-q-8)

src/components/Breadcrumb/index.js
hamishtaplin: Linter complains about using `aria-current`, apparently it's not in the spec.

https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/aria-props.md

src/components/Breadcrumb/index.js
hamishtaplin: True

src/components/Forms/Input/Input.test.js
hamishtaplin: Yeah none of that makes sense actually!

src/components/Forms/Input/index.js
hamishtaplin: Done

src/containers/QuestionnaireMetaPage/index.js
hamishtaplin: Think the best way to do so is to implement the [webpack loader plugin](http://dev.apollodata.com/react/webpack.html), can this be done as a separate task?

WickyNilliams: Added task to trello: https://trello.com/c/pBSrctE2/99-investigate-use-of-webpack-loader-for-graphql-files

src/components/Breadcrumb/index.js
WickyNilliams: Hmm that's odd. Appears to be part of the standard here https://www.w3.org/TR/wai-aria-1.1/#aria-current. Perhaps we should raise a bug against the eslint plugin?

src/components/Forms/Form/Form.test.js
```diff
@@ -3,11 +3,11 @@ import { shallow } from "enzyme";
 import { Form } from "components/Forms/Form";
 
 let wrapper;
-const handleSubmit = jest.fn();
+const onSubmit = jest.fn();
```
WickyNilliams: These functions should actually be called handleSubmit haha. But don't worry if they're just tests!

src/components/RichTextArea/index.js
```diff
@@ -52,7 +52,7 @@ export default class RichTextArea extends Component {
     };
   }
 
-  handleChange = value => {
+  onChange = value => {
```
WickyNilliams: this should be handleChange

PR #150, opened 2017-07-06
40/add page

WickyNilliams:
Adds components that constitute the `Canvas` for designing Pages. This adds only the presentational components, and fakes the rest to allow for testing.

### How to review 
Create a new questionnaire, when redirected to the design page, edit the text on the Canvas

### NOTE

Originally the components were going to be called `PageCanvas`, `PageSection` etc but we decided the Page and Section were overloaded terms in our vocabulary. We have a concept of pages and sections, yet a PageSection related to neither. Therefore PageCanvas was named Canvas, and PageSection CanvasSection

### DIFF
--- a/src/components/QuestionnaireDesign/Canvas.js
+++ b/src/components/QuestionnaireDesign/Canvas.js
@@ -1,6 +1,6 @@
 import styled from "styled-components";
 
-const PageCanvas = styled.div`
+const Canvas = styled.div`
   width: calc(50% - 1em);
   margin: 2em auto;
   padding: 0;
@@ -10,4 +10,4 @@ const PageCanvas = styled.div`
     0 1px 5px 0 rgba(0, 0, 0, .12);
 `;
 
-export default PageCanvas;
+export default Canvas;
--- a/src/components/QuestionnaireDesign/Canvas.js
+++ b/src/components/QuestionnaireDesign/Canvas.js
@@ -1,8 +1,8 @@
 import styled from "styled-components";
 
 const Canvas = styled.div`
-  width: calc(50% - 1em);
-  margin: 2em auto;
+  width: 75%;
+  margin: 2em 0 0 1em;
   padding: 0;
   border-radius: 2px;
   background-color: #fff;
--- a/src/components/QuestionnaireDesign/Canvas.js
+++ b/src/components/QuestionnaireDesign/Canvas.js
@@ -1,13 +1,9 @@
 import styled from "styled-components";
+import Panel from "components/Panel";
 
-const Canvas = styled.div`
+const Canvas = styled(Panel)`
   width: 75%;
   margin: 2em 0 0 1em;
-  padding: 0;
-  border-radius: 2px;
-  background-color: #fff;
-  box-shadow: 0 2px 2px 0 rgba(0, 0, 0, .14), 0 3px 1px -2px rgba(0, 0, 0, .2),
-    0 1px 5px 0 rgba(0, 0, 0, .12);
 `;
 
 export default Canvas;
--- a/src/components/QuestionnaireDesign/CanvasSection.js
+++ b/src/components/QuestionnaireDesign/CanvasSection.js
@@ -3,20 +3,21 @@ import PropTypes from "prop-types";
 import { noop } from "lodash";
 import styled from "styled-components";
 import { colors } from "constants/theme";
+import { transparentize } from "polished";
 
 const StyledCanvasSection = styled.div`
   padding: 2em 2.625em;
   margin-bottom: -2px;
   border-bottom: 2px dashed #c6c6c6;
-  outline-offset: -3px;
+  outline-offset: -2px;
   outline: ${props =>
-    props.focused ? `3px solid ${colors.lightBlue}` : "none"};
+    props.focused ? `2px solid ${colors.lightBlue} !important` : "none"};
 
   &:last-child {
     border: none;
   }
   &:hover {
-    outline: 3px solid ${colors.lightBlue};
+    outline: 2px solid ${transparentize(0.6, colors.lightBlue)};
   }
 `;
 
--- a/src/components/QuestionnaireDesign/CanvasSection.js
+++ b/src/components/QuestionnaireDesign/CanvasSection.js
@@ -10,7 +10,7 @@ const StyledCanvasSection = styled.div`
   border-bottom: 2px dashed #c6c6c6;
   outline-offset: -3px;
   outline: ${props =>
-    props.focussed ? `3px solid ${colors.lightBlue}` : "none"};
+    props.focused ? `3px solid ${colors.lightBlue}` : "none"};
 
   &:last-child {
     border: none;
@@ -24,14 +24,14 @@ export default class CanvasSection extends Component {
   static propTypes = {
     children: PropTypes.node,
     onFocus: PropTypes.func,
-    focussed: PropTypes.bool,
+    focused: PropTypes.bool,
     id: PropTypes.string
   };
 
   static defaultProps = {
     onFocus: noop,
     onBlur: noop,
-    focussed: false
+    focused: false
   };
 
   handleFocus = e => {
@@ -39,11 +39,11 @@ export default class CanvasSection extends Component {
   };
 
   render() {
-    const { children, focussed } = this.props;
+    const { children, focused } = this.props;
 
     return (
-      <StyledCanvasSection onFocus={this.handleFocus} focussed={focussed}>
-        {Children.map(children, child => cloneElement(child, { focussed }))}
+      <StyledCanvasSection onFocus={this.handleFocus} focused={focused}>
+        {Children.map(children, child => cloneElement(child, { focused }))}
       </StyledCanvasSection>
     );
   }
--- a/src/components/QuestionnaireDesign/CanvasSection.js
+++ b/src/components/QuestionnaireDesign/CanvasSection.js
@@ -9,15 +9,20 @@ const StyledCanvasSection = styled.div`
   padding: 2em 2.625em;
   margin-bottom: -2px;
   border-bottom: 2px dashed #c6c6c6;
+
+  transition: outline-color 0.1s ease-in;
   outline-offset: -2px;
-  outline: ${props =>
-    props.focused ? `2px solid ${colors.lightBlue} !important` : "none"};
+  outline-width: 2px;
+  outline-style: solid;
+  outline-color: ${props =>
+    props.focused ? `${colors.lightBlue} !important` : "transparent"};
+
+  &:hover {
+    outline-color: ${transparentize(0.6, colors.lightBlue)};
+  }
 
   &:last-child {
     border: none;
-  }
-  &:hover {
-    outline: 2px solid ${transparentize(0.6, colors.lightBlue)};
   }
 `;
 
--- a/src/components/QuestionnaireDesign/CanvasSection.js
+++ b/src/components/QuestionnaireDesign/CanvasSection.js
@@ -4,7 +4,7 @@ import { noop } from "lodash";
 import styled from "styled-components";
 import { colors } from "constants/theme";
 
-const StyledPageSection = styled.div`
+const StyledCanvasSection = styled.div`
   padding: 2em 2.625em;
   margin-bottom: -2px;
   border-bottom: 2px dashed #c6c6c6;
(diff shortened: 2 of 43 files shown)


WickyNilliams:
[Create PageSection component (3)](https://trello.com/c/IZga7MOs/101-create-pagesection-component-3)

WickyNilliams:
[Create PageField component (2)](https://trello.com/c/aozkpQ4G/102-create-pagefield-component-2)

WickyNilliams:
[Create PageCanvas component (2)](https://trello.com/c/cLkexVI2/103-create-pagecanvas-component-2)

src/components/QuestionnaireDesign/Canvas.js
```diff
@@ -0,0 +1,13 @@
+import styled from "styled-components";
+
+const Canvas = styled.div`
```
samiwel: There's no story for Canvas, do we need one, or is it so simple that it wouldn't offer any benefit?

WickyNilliams: Explained the reasoning for this in the **NOTE** section above

src/components/QuestionnaireDesign/CanvasSection.js
```diff
@@ -0,0 +1,50 @@
...
+import PropTypes from "prop-types";
+import { noop } from "lodash";
+import styled from "styled-components";
+import { colors } from "constants/theme";
+
+const StyledCanvasSection = styled.div`
+  padding: 2em 2.625em;
+  margin-bottom: -2px;
+  border-bottom: 2px dashed #c6c6c6;
+  outline-offset: -3px;
```
hamishtaplin: Going on the Sketch files, the focus outline is 2px.

src/components/QuestionnaireDesign/Canvas.js
```diff
@@ -0,0 +1,13 @@
...
+  width: 75%;
+  margin: 2em 0 0 1em;
+  padding: 0;
+  border-radius: 2px;
+  background-color: #fff;
+  box-shadow: 0 2px 2px 0 rgba(0, 0, 0, .14), 0 3px 1px -2px rgba(0, 0, 0, .2),
```
hamishtaplin: Why the multiple shadows? This is much softer in the Sketch file: `box-shadow: 0 2px 4px 0 rgba(0, 0, 0, 0.1);` which is the same as the shadow in the `Panel` component. We can export this from theme consts or use some HoC to share these values (white bg, radius, shadow)

src/components/QuestionnaireDesign/Canvas.js
```diff
@@ -0,0 +1,13 @@
+import styled from "styled-components";
+
+const Canvas = styled.div`
+  width: 75%;
+  margin: 2em 0 0 1em;
+  padding: 0;
+  border-radius: 2px;
```
hamishtaplin: There's a theme global for this. Maybe there's a better way to share these styles?

src/components/QuestionnaireDesign/Canvas.js
```diff
@@ -0,0 +1,13 @@
...
+  background-color: #fff;
```
hamishtaplin: Could use the theme global.

src/components/QuestionnaireDesign/withSeamlessness.js
```diff
@@ -0,0 +1,106 @@
...
+import { colors } from "constants/theme";
+
+const sizes = {
+  large: css`
+    font-size: 1.75em;
+    font-weight: bold;
+  `,
+
+  medium: css`
+    font-size: 1.125em;
```
hamishtaplin: Can we round these up? `1.1em` etc. Not sure getting precise px values here is important.

src/components/QuestionnaireDesign/withSeamlessness.js
```diff
@@ -0,0 +1,106 @@
...
+    font-weight: bold;
+  `,
+
+  small: css`
+    font-size: 0.875em;
```
hamishtaplin: `0.9em`?

src/components/QuestionnaireDesign/CanvasSection.js
```diff
@@ -0,0 +1,50 @@
...
+  outline: ${props =>
+    props.focussed ? `3px solid ${colors.lightBlue}` : "none"};
+
+  &:last-child {
+    border: none;
+  }
+  &:hover {
+    outline: 3px solid ${colors.lightBlue};
```
hamishtaplin: It's a bit confusing having the `:hover` style the same as focus as it's really easy to have both states active simultaneously on different sections. Similarly, them being the same creates the affordance that clicking on a (hovered) `CanvasSection` is analogous to focusing one.

This might actually be the case, as it seems reasonable that a user would click a `CanvasSection` to explore the (section, question, answer) properties on the right I think?

Regardless, I'd remove the ambiguity by knocking the `:hover` state back to a much lighter blue (just using opacity is fine) and add a transition.

src/components/QuestionnaireDesign/CanvasSection.js
```diff
@@ -0,0 +1,50 @@
...
+import styled from "styled-components";
+import { colors } from "constants/theme";
+
+const StyledCanvasSection = styled.div`
+  padding: 2em 2.625em;
+  margin-bottom: -2px;
+  border-bottom: 2px dashed #c6c6c6;
+  outline-offset: -3px;
+  outline: ${props =>
+    props.focussed ? `3px solid ${colors.lightBlue}` : "none"};
```
hamishtaplin: A transition would be good here. Also, going on the Sketch files, I think the outline should be 2px.

src/components/QuestionnaireDesign/CanvasSection.js
```diff
@@ -0,0 +1,50 @@
...
+  &:hover {
+    outline: 3px solid ${colors.lightBlue};
+  }
+`;
+
+export default class CanvasSection extends Component {
+  static propTypes = {
+    children: PropTypes.node,
+    onFocus: PropTypes.func,
+    focussed: PropTypes.bool,
```
hamishtaplin: This spelling bothers me! A quick google and most articles seem to use "focused" 

eg: https://developer.mozilla.org/en-US/docs/Web/API/Document/activeElement

src/components/QuestionnaireDesign/withSeamlessness.js
```diff
@@ -0,0 +1,106 @@
...
+  &:focus {
+    outline: none;
+  }
+
+  &::placeholder {
+    color: #a3a3a3;
+  }
+
+  &[aria-hidden="true"] {
+    position: absolute;
```
hamishtaplin: This effect can be jarring, can we look at fading the field in/out before applying/removing this?

src/components/QuestionnaireDesign/CanvasSection.js
samiwel: That's a good point, when I was reviewing this PR earlier I got really confused why the section title had blue border permanently when I tabbed focus to the question section. Turned out my mouse cursor was hovering over the section, but I didn't realise it was there because when I type in Chrome, the mouse cursor vanishes!

src/components/QuestionnaireDesign/Canvas.js
samiwel: Sorry, what I meant was that you're missing a `canvas/story.js` file. Maybe it doesn't matter, but the canvas has some associated styles, so perhaps there should be some way to view it in Storybook. But it's not a big deal to be honest.

src/components/QuestionnaireDesign/withSeamlessness.js
WickyNilliams: I just wanted to do the minimum amount possible. Fading in/out means having to setup transitionend events etc, which isn't a small task. UI polish can come later, surely?

src/components/QuestionnaireDesign/CanvasSection.js
WickyNilliams: Good idea. I was just following Dan's HTML prototype, which had exactly this

src/components/QuestionnaireDesign/Canvas.js
WickyNilliams: Ah, I see. I did consider that, but it's only a white box with an shadow. It has no props (beyond children) or behaviour. So seems like it wouldn't benefit much from storybook? Will add if consensus says yes. @hamishtaplin, what do you think?

src/components/QuestionnaireDesign/withSeamlessness.js
WickyNilliams: I would have thought that precise values for fonts (proper scaling) is important? For stuff like layout, it's not such an issue to be pixel perfect, but font scale is usually designed very specifically?

src/components/QuestionnaireDesign/withSeamlessness.js
```diff
@@ -0,0 +1,106 @@
...
+
+      return (
+        <StyledSeamless
+          placeholder={placeholder}
+          autoFocus={autoFocus}
+          value={value}
+          onChange={onChange}
+          name={id}
+          size={size}
+          aria-hidden={shouldBeHidden}
```
WickyNilliams: The reason most HTML attributes are camelCase is because they directly correspond to JS properties on the DOM objects. There are no such properties for aria (or `data-*`), they only exist as attributes, so they're written like that

src/components/QuestionnaireDesign/CanvasSection.js
WickyNilliams: Even the HTML spec uses the single 's' version. I'll change

src/components/QuestionnaireDesign/Canvas.js
WickyNilliams: Copied from Dan's prototype. We should just use the theme functionality of SC, no need for HoC!

src/components/QuestionnaireDesign/withSeamlessness.js
hamishtaplin: It's harder to keep that scale consistent when the values are so obscure though, unless we define a proper type scale somewhere? I think the type being whole pixels is just a side-effect of the design tools doing so.

src/components/QuestionnaireDesign/withSeamlessness.js
hamishtaplin: Fair enough, I thought I'd camel-cased an `aria` attribute elsewhere but it seems not.

src/components/QuestionnaireDesign/Canvas.js
hamishtaplin: Yeah I was using the theme props for a bit but decided that just importing values was cleaner since we'll never have another "theme".

src/components/QuestionnaireDesign/withSeamlessness.js
hamishtaplin: Not a big job but happy to address separately, would be good to run this by Dan I think.

src/components/QuestionnaireDesign/Canvas.js
hamishtaplin: Worth considering writing all the stories at `Canvas` level as the `CanvasSection` components are kind of tied to their container, visually? Think it would make more sense from a users perspective, kind of like the way the various `Field/Input/Select` components work.

WickyNilliams: How do you mean? For `CanvasSection` I have already included `SeamlessInput`s to show how focus works. Do you mean above and beyond that?

src/components/QuestionnaireDesign/withSeamlessness.js
WickyNilliams: Agreed, it's not very clear where those values come as it stands. I'm sure I saw Dan had an actual typographic scale written up somewhere. I'll try and dig it out and add the values to the theme, then use those values?

src/components/QuestionnaireDesign/withSeamlessness.js
WickyNilliams: Yeah, Dan's input on that would be valuable. Let's leave for now

WickyNilliams:
@hamishtaplin i'm able to tab into the properties pane, but I notice the focus (styling only) stays on the final `CanvasSection` when you tab to the sidebar. 

tab-focus

hamishtaplin:
@WickyNilliams LGTM apart from the focus thing I mentioned?

PR #154, opened 2017-07-07
Add linting rules for imports/exports

samiwel:
This PR adds sensible linting rules for imports and exports to Author.

### How to review 
Linting should pass without warnings/errors.
Build should pass.
Breaking the import/export linting rules should result in an appropriate error and break the build.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,7 +3,8 @@
   "plugins": [
     "react",
     "babel",
-    "jasmine"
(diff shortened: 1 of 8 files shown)


WickyNilliams:
[Linting rules around import/export (0.5)](https://trello.com/c/W3P7ZT8k/77-linting-rules-around-import-export-05)

PR #156, opened 2017-07-10
allow focus on CanvasSection with a click

WickyNilliams:
Clicking on a CanvasSection should be the same as tabbing into a field - optional && empty inputs become visible. Currently this is not the case. This PR fixes that.

### How to review 
Run tests. Inspect code. Test in browser:

1. go to QuestionnairDesign page
2. delete text from *section description field*
3. tab to another section
4. observe *section description field* is not visible
5. click anywhere in Section section (uh, naming)
6. observe that *section description field* is now visible, even if no field has focus

### DIFF
--- a/src/components/QuestionnaireDesign/CanvasSection.js
+++ b/src/components/QuestionnaireDesign/CanvasSection.js
@@ -54,6 +54,7 @@ export default class CanvasSection extends Component {
 
     return (
       <StyledCanvasSection
+        onClick={this.handleFocus}
         onFocus={this.handleFocus}
         onBlur={this.handleBlur}
         focused={focused}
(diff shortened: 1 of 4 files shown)


WickyNilliams:
[Clicking CanvasSection should make optional && empty fields visible](https://trello.com/c/xTNDPk3U/122-clicking-canvassection-should-make-optional-empty-fields-visible)

src/components/QuestionnaireDesign/CanvasSection.js
```diff
@@ -54,6 +54,7 @@ export default class CanvasSection extends Component {
 
     return (
       <StyledCanvasSection
+        onClick={this.handleFocus}
```
hamishtaplin: Is "focus" now a bit of a misnomer? The `CanvasSection` is active/selected rather than "focused"? It's a bit misleading, I think.

WickyNilliams: It wouldn't behave much differently if I were to make the div itself focusable (via `tabindex`). Why do you think it's not correct?

hamishtaplin: Just because it's not in focus (the children are) so would be good to disambiguate. No biggie though can always change it another time if it causes any confusion.

PR #159, opened 2017-07-14
WIP 40 add page

hamishtaplin:

[Trello card](https://trello.com/c/HHkiluUf/40-add-new-question-page-to-section-5)

- Added a basic sidebar with the section/page structure and an "add page" button
- Clicking "add page" button makes roundtrip to the API, optimistically updating the UI with the new page
- Section and Page fields are editable and save on blur

### How to review 

1. Navigate to an existing Q
2. Click "add page"
3. Confirm new page is added
4. Confirm focus is on the Page Title (unless the section title is empty)
5. Ensure tests pass, etc.


src/components/Nav/index.js
```diff
@@ -48,15 +53,22 @@ const Nav = ({ questionnaire }) => {
...
+  match: PropTypes.shape({
+    isExact: PropTypes.bool,
+    params: PropTypes.object.isRequired,
+    path: PropTypes.string.isRequired,
+    url: PropTypes.string.isRequired
+  })
 };
 
-export default Nav;
+export default withRouter(NavWithoutRouter);
```
WickyNilliams: Seems a bit much to require the router for a nav? Why can't some other component pass it what it needs? Would simplify tests also

src/components/QuestionnaireDesign/QuestionnaireDesign.test.js
```diff
@@ -19,6 +19,34 @@ describe("QuestionnaireDesign", () => {
...
+  });
+
+  it("should focus on empty section title upon mount", () => {
+    expect(document.activeElement.name).toEqual("section.title");
+  });
+
+  it("should focus on empty page title when section title is not empty", () => {
+    wrapper.setProps({ section: { title: "Section 1" } });
+    wrapper.unmount();
+    wrapper.mount();
```
WickyNilliams: What's this about? To render the component anew? If so, it distracts from the purpose of the test IMO. Since you're changing the props for every test, why not create a function to wrap up that logic?

src/components/QuestionnaireDesign/QuestionnaireDesign.test.js
```diff
@@ -19,6 +19,34 @@ describe("QuestionnaireDesign", () => {
...
+
+    expect(document.activeElement.name).toEqual("page.title");
+  });
+
+  it("should move focus to empty section title upon navigation to new page", () => {
+    wrapper.setProps({
+      section: { id: "2", title: "" },
+      page: { id: "3", title: "" }
+    });
```
WickyNilliams: Isn't this essentially what your initial props are (empty titles)? so you wouldn't need to set any new props here

src/containers/QuestionnaireDesignPage/QuestionnaireDesignPage.js
```diff
@@ -60,19 +63,72 @@ export class QuestionnaireDesignPage extends Component {
...
+      return;
+    }
+
+    if (focused !== this.state.focused) {
+      this.setState({ focused });
     }
   };
 
+  renderSidebar(questionnaire) {
+    return (
```
WickyNilliams: Why not extract this into a component? That way it's ready for further dev

src/containers/QuestionnaireDesignPage/index.js
```diff
@@ -37,21 +38,79 @@ export const withQuestionnaire = graphql(getQuestionnaireQuery, {
...
+
+export const redirectToDesigner = ownProps => ({ data }) => {
+  const { history, sectionId, questionnaireId } = ownProps;
+  const { id } = data.createQuestionPage;
+  history.push(`/questionnaire/${questionnaireId}/design/${sectionId}/${id}`);
+};
+
+export const mapMutateToProps = ({ ownProps, mutate }) => {
+  return {
+    onAddPage(sectionId) {
```
WickyNilliams: Could you use `options` here (docs don't make it clear if that's a query only thing, or mutations too)? Would save having to pass the sectionId. http://dev.apollodata.com/react/queries.html#options-from-props

src/containers/QuestionnaireDesignPage/index.test.js
```diff
@@ -0,0 +1,155 @@
...
+      });
+      expect(props.loading).toBe(false);
+      expect(props.questionnaire).toEqual(questionnaire);
+      expect(props.section).toEqual(section);
+      expect(props.page).toEqual(page);
+    });
+  });
+
+  describe("createUpdater", () => {
+    it("should have the correct values passed to readQuery/writeQuery and the result to be the correct page", () => {
```
WickyNilliams: "should update the cache"?

src/containers/QuestionnaireDesignPage/index.test.js
```diff
@@ -0,0 +1,155 @@
...
+      const readQuery = jest.fn().mockImplementation(({ query, variables }) => {
+        expect(variables.id).toEqual(questionnaire.id);
```
WickyNilliams: Hmm... not sure i like the expect statements being peppered throughout the test. I'd expect (punintentional) assertions to be at bottom of test.

Why not use the more standard 

```js
expect(readQuery).toHaveBeenCalledWith(...)
```

or

```js
expect(readQuery.mock.calls[0].variables.id).toBe(questionnaire.id)
```

or some other approach (e.g. `toMatchObject`)?

src/containers/QuestionnaireDesignPage/index.test.js
```diff
@@ -0,0 +1,155 @@
...
+      ownProps = { history };
+      props = mapMutateToProps({ ownProps, mutate });
+    });
+
+    it("should have a onAddPage prop", () => {
+      expect(props.onAddPage).toBeInstanceOf(Function);
+    });
+
+    it("should respond with an optimisticResponse", () => {
+      props.onAddPage(section.id).then(() => {
```
WickyNilliams: you should return the promise for async tests. e.g. `return props.onAddPage(...)`, otherwise it can crash the test runner

src/containers/QuestionnaireDesignPage/index.test.js
```diff
@@ -0,0 +1,155 @@
...
+            createQuestionPage: {
+              sectionId: section.id
+            }
+          }
+        });
+      });
+    });
+
+    it("should redirect", () => {
+      props.onAddPage(section.id).then(() => {
```
WickyNilliams: same here, should return the promise

src/components/Nav/index.js
```diff
@@ -56,7 +62,13 @@ const Nav = ({ questionnaire }) => {
...
 
 Nav.propTypes = {
-  questionnaire: CustomPropsTypes.questionnaire
+  questionnaire: CustomPropsTypes.questionnaire,
+  match: PropTypes.shape({
+    isExact: PropTypes.bool,
+    params: PropTypes.object.isRequired,
+    path: PropTypes.string.isRequired,
+    url: PropTypes.string.isRequired
+  })
```
hamishtaplin: I don't actually know. I was trying to find out if react-router exported any PropTypes and I think I just came across this.

src/components/Nav/index.js
hamishtaplin: The problem is that RR is supposed to pass `match.params` into that `isActive` function but it doesn't appear to work correctly. The alternative is to pass the params down from the Page/s  to the BaseLayout and then into the Nav.

I have no problems with the Nav accessing the state of the router—it's the one component (navigation) that seems appropriate to have that information?

src/tests/utils/shallowWithRouter.js
```diff
@@ -0,0 +1,11 @@
...
+import { shallow } from "enzyme";
+import createRouterContext from "react-router-test-context";
+
+const shallowWithRouter = child =>
+  shallow(child, {
+    context: createRouterContext(),
+    childContextTypes: { router: PropTypes.object }
+  });
+
+export default shallowWithRouter;
```
hamishtaplin: I keep having to recreate it to check if it works for a test or not, can't we just keep it around?

src/components/QuestionnaireDesign/QuestionnaireDesign.test.js
hamishtaplin: It works and is obvious what's going on, I'd have to refactor all the tests here to support one thing—seems a bit excessive?

src/containers/QuestionnaireDesignPage/QuestionnaireDesignPage.js
hamishtaplin: It's out of scope for this story and would also mean writing tests for that component, probably better to just remove it and keep the button?

WickyNilliams:
[Add New Question Page to Section (5)](https://trello.com/c/HHkiluUf/40-add-new-question-page-to-section-5)

PR #162, opened 2017-07-18
upgrade react-apollo

### LINKED ISSUE #160: Questionnaire Meta page crashes on editing a field
### Expected behaviour

A user should be able to edit fields.

### Actual behaviour

The app crashes when a field is edited.

1. Navigate to an existing questionnaire (or create a new one)
2. Visit the "Questionnaire meta" tab
3. Edit any field (eg 'Title')
4. Press tab to 'blur'.
5. Error `TypeError: this.props.onUpdate is not a function` will be displayed.

#### Browser

Chrome

### Screenshot

screenshot 2017-07-14 08 57 57

WickyNilliams:
Fixes issue with react-apollo v1.4.3 which prevented you from testing modules that internally used react-apollo 

### How to review 
Run tests, ensure pass. There are already some tests in the suite that would fail under react-apollo 1.4.3, so all tests passing serves as proof of fix.

**EDIT:** this does not fix issue 160, not sure why i thought that

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,11 +15,13 @@
     "storybook-build": "export NODE_PATH=src/ && build-storybook -s public -o storybook-static",
     "storybook-deploy": "yarn storybook-build && gh-pages -d .out",
     "precommit": "lint-staged",
-    "coverage": "yarn test -- --coverage -u",
-    "postinstall": "rm -rf node_modules/react-apollo/node_modules/react-dom && echo 'ran temporary fix for react-apollo'"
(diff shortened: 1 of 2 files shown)


hamishtaplin:
@WickyNilliams LGTM

It doesn't fix #160 but it does partially unblock our ability to write the tests to ensure #160 doesn't happen again, I think?

WickyNilliams:
@hamishtaplin yeah, not sure why i thought that was the case. But yeah it unblocks writing tests to corner off #160

PR #165, opened 2017-07-21
80/popout component

WickyNilliams:
Adds basic Popout component. This component can accept any other component as content, and any other component trigger for toggling visiblity of that content. It also supports animating content in and out of the DOM

### How to review 
Inspect code, run tests, play with storybook


WickyNilliams:
[Pop-Out Component (5)](https://trello.com/c/OoZRNUtk/80-pop-out-component-5)

PR #167, opened 2017-07-24
webpack loader and linter for graphql files

bennpearson:
Added graphql loader, jest-transform-graphql and eslint gql plugin

### How to review 
Check webpack is using graphql loader and that `yarn lint` doesn't produce any errors.


bennpearson:
[use webpack loader and linter for graphql files (2)](https://trello.com/c/pBSrctE2/99-use-webpack-loader-and-linter-for-graphql-files-2)

.eslintrc.js
```diff
@@ -0,0 +1,22 @@
+const schema = require("eq-author-graphql-schema/schema");
```
hamishtaplin: We already have an eslint config in `.eslintrc`—this should go in there.

WickyNilliams: `.eslintrc` was converted to a JS module as the `eslint-plugin-graphql` doesn't seem to support what we need in JSON format (i've [asked](https://github.com/apollographql/eslint-plugin-graphql/issues/58#issuecomment-317442089) if this can be done or if it can be added)

WickyNilliams:
@bennpearson you may need to add `.eslintrc.js` to the `.codecov.yml` ignore list to stop it being included in coverage (and then hopefully have all checks pass)

PR #168, opened 2017-07-25
Integrate the Mock API with Author

samiwel:
This is WIP at the moment. I am in the process of integrating the Mock API implementation into the Author application.

### How to review 
TBD


samiwel:
[Mock API (8)](https://trello.com/c/y092sqhA/100-mock-api-8)

src/tests/utils/MockResolvers.js
```diff
@@ -0,0 +1,115 @@
...
+  if (typeof Storage !== "undefined") {
+    const previousData = localStorage.getItem("mockDataStore");
+    if (previousData) {
+      dataStore = merge(new MockDataStore(), JSON.parse(previousData));
+    }
+  }
+
+  if (dataStore === null) {
+    dataStore = new MockDataStore();
+    updateLocalStorage(dataStore);
```
WickyNilliams: Why updateLocalStorage here? If you've just loaded from localStorage, there no reason to instantly update is there?

src/tests/utils/MockResolvers.js
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  return dataStore;
+};
```
WickyNilliams: Given everything I've said, this function could possibly be simplified to this:

```js
const createMockDataStore = () => {
  const seed = JSON.parse(localStorage.getItem("mockDataStore")) || {};
  return new MockDataStore(seed);
};
```

?

src/tests/utils/MockDataStore.js
```diff
@@ -0,0 +1,175 @@
...
+    const sectionToDelete = this.sections[section.id];
+
+    forEach(sectionToDelete.pages, p => {
+      this.deletePage(p);
+    });
+
+    if (sectionToDelete.questionnaireId) {
+      remove(
+        this.questionnaires[sectionToDelete.questionnaireId].sections,
+        s => s.id === sectionToDelete.id
```
WickyNilliams: Lodash has a handy shorthand for remove where you pass an object. So this (and other uses) could then changed to:

```js
remove(
  this.questionnaires[sectionToDelete.questionnaireId].sections,
  { id : sectionToDelete.id }
)
```

src/tests/utils/MockDataStore.js
```diff
@@ -0,0 +1,175 @@
+import { values, merge, forEach, remove } from "lodash";
+
+class MockDataStore {
+  constructor(seedData = {}) {
+    this.counter = {
+      questionnaire: 0,
+      section: 0,
+      page: 0,
+      answer: 0
```
WickyNilliams: How about some helper function for all these counters, to save having to manually increment?

e.g. 

```js
const createCounter = (start = 0, step = 1) => () => start = start + step;

// usage
this.counter = {
  questionnaire : createCounter()
};

const id = this.counter.questionnaire();
this.questionnaires[id] = merge(questionnaire, { id }, { sections: [] });
```

tests/spec/smoketest-author-spec.js
```diff
@@ -0,0 +1,82 @@
...
+    it("should update navigation title when section title changed", () => {
+      expect(navigationHasDefaultSectionTitle()).toEqual(true);
+
+      setSectionTitle("My Section Title");
+      browser.keys("Tab");
+
+      expect(getFirstSectionTitle()).toEqual("My Section Title");
+    });
+  });
+});
```
WickyNilliams: 👍 for having these tests. They will prove infinitely useful

config/webpack.config.prod.js
```diff
@@ -158,7 +158,11 @@ module.exports = {
       // Process JS with Babel.
       {
         test: /\.(js|jsx)$/,
-        include: paths.appSrc,
+        include: [
+          paths.appSrc,
+          path.resolve(process.cwd(), "node_modules/eq-author-graphql-schema"),
+          path.resolve(process.cwd(), "node_modules/eq-author-mock-api")
```
WickyNilliams: We should add these to `paths` module or something, so that everything like this is concentrated in one module

tests/spec/smoketest-author-spec.js
```diff
@@ -0,0 +1,82 @@
...
+
+  it("should take me to Questionnaire meta screen when createQuestionnire is clicked", () => {
+    clickCreateQuestionnaire();
+
+    expect(onCreateQuestionnairePage()).toEqual(true);
+  });
+
+  describe("creating a new questionnaire", () => {
+    beforeAll(() => {
+      goToUrl("http://localhost:3000/");
```
samiwel: Good to know, will remove 👍

samiwel: Didn't seem to work when I removed, so added back in, will take a look and see if there's some way I can consolidate the `describe` closures.

tests/spec/smoketest-author-spec.js
```diff
@@ -0,0 +1,82 @@
...
+        true
+      );
+      clickCreateButton();
+      expect(onDesignQuestionnairePage()).toEqual(true);
+    });
+  });
+
+  describe("designing the questionnaire", () => {
+    beforeAll(() => {
+      goToUrl("http://localhost:3000/");
```
samiwel: See above.

src/tests/utils/MockResolvers.js
```diff
@@ -0,0 +1,115 @@
...
+const updateLocalStorage = dataStore => {
+  if (typeof Storage !== "undefined") {
+    localStorage.setItem("mockDataStore", JSON.stringify(dataStore));
+  }
+};
+
+const createMockDataStore = () => {
+  let dataStore = null;
+
+  if (typeof Storage !== "undefined") {
```
samiwel: To check that the environment supports localStorage. See https://www.w3schools.com/html/html5_webstorage.asp. Not sure if this really applies to us since we're assuming modern version of Chrome etc. Can possibly remove the check.

src/tests/utils/MockResolvers.js
samiwel: Guess it's not strictly needed here. It will update on each mutation anyway. This was just a belts and braces approach, I think.

src/tests/utils/MockDataStore.js
samiwel: Nice tip, thanks.

samiwel: As discussed, leaving as is since it would cause issues when persisting/retrieving next counter Id values. Since this isn't a fully-fledged data store it probably okay to leave as is.

src/tests/utils/MockResolvers.js
samiwel: This didn't work, since the `parse` function throws a SyntaxError if nothing is retrieved from `localStorage`.

Changed to this instead:
```
const createMockDataStore = () => {
  const seed = localStorage.getItem("mockDataStore");
  return seed ? new MockDataStore(JSON.parse(seed)) : new MockDataStore();
};
```

src/tests/utils/MockDataStore.js
samiwel: Nice tip. Although my IDE complains that it's not a function! :-)

config/webpack.config.prod.js
samiwel: Good idea, but I think introducing a new module should be a separate tech debt task.

tests/spec/smoketest-author-spec.js
samiwel: Agreed, but we need to be disciplined as a team to add additional tests as new features are developed. Perhaps we should consider making it part of the formal acceptance criteria, as is the case on Runner.

WickyNilliams: That's a great idea

src/tests/utils/MockResolvers.js
WickyNilliams: Was it blowing up in tests specifically? That might be a quirk of your fake localStorage implementation. The following doesn't blow up (try in browser console):

```js
JSON.parse(localStorage.getItem("a-key-that-definitely-doesnt-exist")) === null
```

PR #169, opened 2017-07-26
Font weights

bennpearson:
Made font weight's consistent 

### How to review 
Check that `400, 500, 600, 700 and 900` font weights have been replaced with just `400`, `700` and `900`

### DIFF
--- a/src/components/Button/index.js
+++ b/src/components/Button/index.js
@@ -48,7 +48,7 @@ const StyledButton = styled.button`
   padding: 0.8em 3em;
   border-radius: ${radius};
   font-size: 0.8em;
-  font-weight: 500;
+  font-weight: 400;
   cursor: pointer;
   line-height: 1;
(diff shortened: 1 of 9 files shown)


bennpearson:
[Font weight constants (1)](https://trello.com/c/ZzvRAABX/82-font-weight-constants-1)

src/components/PropertyPane/PropertyPaneTitle.js
```diff
@@ -4,7 +4,7 @@ import styled from "styled-components";
 const PropertyPaneTitle = styled.h3`
   font-size: 0.6em;
   text-transform: uppercase;
-  font-weight: 900;
+  font-weight: bold;
```
hamishtaplin: I don't think we should be changing this (from black) to bold? There are other (not yet developed) parts of the app that also use black so we can't just remove it.

PR #174, opened 2017-08-01
Add Checkbox Answer component

samiwel:

__N.B. This PR is still WIP__

This PR adds the Checkbox Answer component and micro interactions.
It also introduces new stories for the Checkbox Answer component.
See [trello card](https://trello.com/c/9Q7MGXS6/159-add-edit-delete-checkbox-answer-component-3) for more details.

### How to review 
All tests should pass.
Component should behave as specified in the design.


samiwel:
[Add / Edit / Delete checkbox answer component (3)](https://trello.com/c/9Q7MGXS6/159-add-edit-delete-checkbox-answer-component-3)

src/components/Answers/CheckboxAnswer/story.js
```diff
@@ -0,0 +1,134 @@
...
+    filter(
+      newState.options,
+      option => option.id === parseInt(optionToUpdate, 10)
+    ).forEach(match => merge(match, { [fieldToUpdate]: value }));
+
+    this.setState(newState);
+  };
+
+  handleAddOption = () => {
+    const nextId = this.state.counter + 1;
```
WickyNilliams: Here you could use the `createCounter` helper i was talking about the other day. 

Also, no need to put things into `state` unless they directly affect rendering (which counter does not). You can store the counter as an instance property e.g. `this.counter++` instead

src/components/Answers/CheckboxAnswer/story.js
```diff
@@ -0,0 +1,134 @@
...
+        options={this.state.options}
+        onChangeLabel={this.handleChange}
+        onChangeDescription={this.handleChange}
+        onAddOption={this.handleAddOption}
+        onDeleteOption={this.handleDeleteOption}
+        onAddOther={this.handleAddOther}
+      />
+    );
+  }
+}
```
WickyNilliams: You might find that the logic for all this is simplified if you use an object keyed by ID to store the options, rather than an array. This is why redux generally uses objects and keys everything by id.

But since this is just a story, it's not a big deal. Something to keep in mind for future :)

src/components/Answers/CheckboxAnswer/index.js
```diff
@@ -0,0 +1,178 @@
...
+
+  &:hover {
+    color: #333;
+    transition: color .2s ease-in-out;
+  }
+`;
+
+const OtherOption = styled.small`margin: 0 .5em;`;
+
+const CheckboxAnswer = ({
```
WickyNilliams: This should probably be renamed to be consistent with the Text answer type. Benn named it `AnswerText`, so that all answers would be grouped when sorted alphabetically

src/components/Answers/CheckboxAnswer/checkboxanswer.test.js
```diff
@@ -0,0 +1,60 @@
...
+  });
+
+  it("should match snapshot", () => {
+    const wrapper = shallow(<CheckboxAnswer {...mockHandlers} />);
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it("should have one option by default", () => {
+    const wrapper = mount(<CheckboxAnswer {...mockHandlers} />);
+    expect(wrapper.find(CheckboxOption)).toHaveLength(1);
```
samiwel: The snapshot will test this but it's a belts and braces approach incase the snapshot is ever updated, I don't think there's any harm in having it there since having a default option is part of the acceptance criteria.

src/components/Answers/CheckboxAnswer/index.js
samiwel: I've moved the AnswerText component into the same folder structure and renamed it to TextAnswer for consistency.

src/components/Answers/CheckboxAnswer/index.js
```diff
@@ -0,0 +1,178 @@
...
+            </CSSTransition>
+          )}
+      </TransitionGroup>
+      <div>
+        <Button secondary onClick={onAddOption}>
+          Add another option
+        </Button>
+        <OtherOption>
+          or{" "}
+          <AddOtherLink href="#" onClick={onAddOther}>
```
samiwel: I've made this a button. 👍

src/components/Answers/CheckboxAnswer/index.js
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  &.option-entered {
+    height: auto;
+  }
+`;
+
+const SeamlessLabel = styled(SeamlessInput)`
+  display: inline-block !important;
+  width: auto;
```
samiwel: This has been done, Github not showing the line change.

src/components/Answers/CheckboxAnswer/checkboxanswer.test.js
WickyNilliams: But they'll never differ? There will never be a case where this test passes and the snapshot doesn't have 1 CheckboxOption. Snapshot updates should be reviewed as part of PRs

PR #176, opened 2017-08-09
156/delete page

WickyNilliams:
Allows user to delete pages

### How to review 
Review code, play with storybook, play with app, run tests

**NOTE:** requires https://github.com/ONSdigital/eq-author-api/pull/24

### DIFF
--- a/src/components/QuestionnaireNav/PageNav.js
+++ b/src/components/QuestionnaireNav/PageNav.js
@@ -1,4 +1,5 @@
 import React from "react";
+import PropTypes from "prop-types";
 import styled from "styled-components";
 
 import CustomPropTypes from "custom-prop-types";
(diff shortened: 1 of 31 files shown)


WickyNilliams:
[Delete Page (5)](https://trello.com/c/h4lkTX9Y/156-delete-page-5)

src/components/QuestionnaireNav/PageNav.js
```diff
@@ -103,6 +142,12 @@ const PageNav = ({ section, questionnaire }) =>
                 {pageNumber} {page.title || "Page Title"}
               </LinkText>
             </Link>
+            <DeleteButton
+              type="button"
+              onClick={partial(onDelete, section.id, page.id)}
+            >
+              ×
```
hamishtaplin: I think it's worth adding something here for screenreaders, even if it's just `aria-label="Close"`

samiwel:
Not sure if this is something that I've introduced with my changes to the API or not... But I just checked out the latest from this branch and it wasn't deleting pages for me. Clicked the X button, but the page doesn't get deleted. I get the following stack trace in the API. 
Stack trace when trying to delete a page.

To me, this suggests the `pageType` is not being passed with the query. Might wan't to double check this.

PR #178, opened 2017-08-14
Added Dockerfile and updated build

jonnyshaw89:
I have added a Dockerfile and updated the Travis build to push the images to DockerHub

### How to review 
Build the Docker image `docker build -t onsdigital/eq-author .`
Then run it `docker run -it -p3000:3000 -e REACT_APP_API_URL=http://localhost:4000/graphql onsdigital/eq-author` to see if it works as expected

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,11 @@ script:
   - yarn lint -- --max-warnings=0
   - yarn coverage
   - yarn build
+  - docker login -u $DOCKER_USERNAME -p $DOCKER_PASSWORD
+  - export TAG=`if [ "$TRAVIS_PULL_REQUEST_BRANCH" == "" ]; then echo "latest"; else echo $TRAVIS_PULL_REQUEST_BRANCH; fi`
+  - docker build -t onsdigital/eq-author:$TAG -f Dockerfile .
+  - echo "Pushing with tag [$TAG]"
+  - docker push onsdigital/eq-author:$TAG
 
 after_success:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -63,4 +68,8 @@ deploy:
   on:
     branch:
       - master
-      - 89-harden-storybook
\ No newline at end of file
+      - 89-harden-storybook
```
collisdigital: do we need this branch anymore?

jonnyshaw89: Given it doesn't exist any more i would imagine not.

docker-entrypoint.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+yarn build
```
samiwel: This line would need the environment variables `REACT_APP_USE_MOCK_API=false` and `REACT_APP_API_URL` to be defined.

jonnyshaw89: `REACT_APP_API_URL` is supplied when running the container. with regards to `REACT_APP_USE_MOCK_API` in Runner we default to Production values and override for Dev values. Does author have a consistent approach?

samiwel: Yes, I think so. By default `REACT_APP_USE_MOCK_API` will default to undefined/false. We explicitly have to tell Author if we want it to run against a Mock API implementation. Otherwise it will use whatever is supplied in `REACT_APP_API_URL`.

jonnyshaw89: I am passing this value in when it is run. See here https://github.com/ONSdigital/eq-compose/pull/1

PR #180, opened 2017-08-16
Answer text changes saved graphql

bennpearson:
Answer text input working with graphql

### How to review 
Check answer text component is working with graphql and that all tests pass


src/containers/QuestionnaireDesignPage/QuestionnaireDesignPage.js
```diff
@@ -16,49 +16,60 @@ export class QuestionnaireDesignPage extends Component {
...
+  handleAddAnswer = type => {
+    this.props.onAddAnswer(type);
   };
 
   handleBlur = focused => {
     this.setFocused(focused);
 
-    switch (this.state.focused) {
-      case "section":
+    switch (true) {
```
WickyNilliams: I think this switch statement would be clearer as a series of if/else if/else.

src/components/Answers/TextAnswer.js
```diff
@@ -2,20 +2,21 @@ import React from "react";
 import PropTypes from "prop-types";
 import CustomPropTypes from "custom-prop-types";
 import { Field, Input } from "components/Forms";
-import SeamlessInput from "./SeamlessInput";
-import SeamlessTextArea from "./SeamlessTextArea";
+import SeamlessInput from "components/QuestionnaireDesign/SeamlessInput";
```
samiwel: Any reason why these have been moved back? Discussed with @hamishtaplin previously that SeamlessInput should really be it's own component, not nested down inside QuestionnaireDesign.

src/components/Answers/TextAnswer.js
```diff
@@ -2,20 +2,21 @@ import React from "react";
...
+import SeamlessTextArea from "components/QuestionnaireDesign/SeamlessTextArea";
```
samiwel: Same as above.

bennpearson:
[Add/edit text field answer (5)](https://trello.com/c/mQZxAPTs/71-add-edit-text-field-answer-5)

PR #185, opened 2017-08-21
Delete answers

samiwel:
This PR adds the ability to delete answers from the canvas.

### How to review 
All tests should pass.
Should be possible to remove answers by clicking on the X button next to the answer.

### DIFF
--- a/src/components/DeleteButton/deletebutton.test.js
+++ b/src/components/DeleteButton/deletebutton.test.js
@@ -7,7 +7,7 @@ describe("DeleteButton", () => {
   it("should render", () => {
     expect(mount(<DeleteButton size="small" />)).toMatchSnapshot("small");
     expect(mount(<DeleteButton size="medium" />)).toMatchSnapshot("medium");
-    expect(mount(<DeleteButton size="large" />)).toMatchSnapshot("largeu");
+    expect(mount(<DeleteButton size="large" />)).toMatchSnapshot("large");
   });
 
   it("should invoke callback when clicked", () => {
@@ -17,11 +17,5 @@ describe("DeleteButton", () => {
     wrapper.simulate("click");
 
     expect(onClickCallback).toHaveBeenCalled();
-  });
-
-  it("should display an &times;", () => {
-    const wrapper = mount(<DeleteButton />);
-
-    expect(wrapper.text()).toEqual("×");
   });
 });
--- a/src/components/DeleteButton/deletebutton.test.js
+++ b/src/components/DeleteButton/deletebutton.test.js
@@ -1,0 +1,27 @@
+import React from "react";
+
+import { mount } from "enzyme";
+import DeleteButton from "components/DeleteButton";
+
+describe("DeleteButton", () => {
+  it("should render", () => {
+    expect(mount(<DeleteButton size="small" />)).toMatchSnapshot("small");
+    expect(mount(<DeleteButton size="medium" />)).toMatchSnapshot("medium");
+    expect(mount(<DeleteButton size="large" />)).toMatchSnapshot("largeu");
+  });
+
+  it("should invoke callback when clicked", () => {
+    const onClickCallback = jest.fn();
+    const wrapper = mount(<DeleteButton onClick={onClickCallback} />);
+
+    wrapper.simulate("click");
+
+    expect(onClickCallback).toHaveBeenCalled();
+  });
+
+  it("should display an &times;", () => {
+    const wrapper = mount(<DeleteButton />);
+
+    expect(wrapper.text()).toEqual("×");
+  });
+});
--- a/src/components/DeleteButton/index.js
+++ b/src/components/DeleteButton/index.js
@@ -1,9 +1,24 @@
 import React from "react";
-import styled from "styled-components";
+import styled, { css } from "styled-components";
 import PropTypes from "prop-types";
 import { colors } from "constants/theme";
 
-const defaultSize = "1.25em";
+const sizes = {
+  small: css`
+    font-size: 1.25em;
+    padding: 0 auto;
+  `,
+
+  medium: css`
+    font-size: 2em;
+    padding: .3em;
+  `,
+
+  large: css`
+    font-size: 3em;
+    padding: .3em;
+  `
+};
 
 const StyledDeleteButton = styled.button`
   color: ${colors.lightGrey};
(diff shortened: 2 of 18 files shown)


src/components/DeleteButton/deletebutton.test.js
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+
+import { mount } from "enzyme";
+import DeleteButton from "components/DeleteButton";
+
+describe("DeleteButton", () => {
+  it("should render", () => {
+    expect(mount(<DeleteButton size="small" />)).toMatchSnapshot("small");
+    expect(mount(<DeleteButton size="medium" />)).toMatchSnapshot("medium");
+    expect(mount(<DeleteButton size="large" />)).toMatchSnapshot("largeu");
```
WickyNilliams: `largeu` -> `large`

src/components/DeleteButton/deletebutton.test.js
```diff
@@ -0,0 +1,27 @@
...
+
+    wrapper.simulate("click");
+
+    expect(onClickCallback).toHaveBeenCalled();
+  });
+
+  it("should display an &times;", () => {
+    const wrapper = mount(<DeleteButton />);
+
+    expect(wrapper.text()).toEqual("×");
```
WickyNilliams: This test is redundant, the snapshots already prove this, as seen here https://github.com/ONSdigital/eq-author/pull/185/files#diff-ff739f289d840d0c2989dba14963189fR99

src/components/DeleteButton/index.js
```diff
@@ -0,0 +1,52 @@
...
+    transition: color .2s ease-out;
+  }
+
+  font-size: ${props => {
+    if (props.size === "medium") {
+      return "2em";
+    } else if (props.size === "large") {
+      return "3em";
+    } else {
+      return defaultSize;
```
WickyNilliams: It's odd that this is called defaultSize, but the defaultProp is `"medium"`

src/components/DeleteButton/index.js
```diff
@@ -0,0 +1,52 @@
...
+    }
```
WickyNilliams: Instead of if/else, how about extracting an object where the keys are sizes and values css? I did this for `withSeamlessness` if you want an example

src/components/QuestionnaireDesign/QuestionnaireDesign.story.js
```diff
@@ -0,0 +1,69 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import { withKnobs } from "@storybook/addon-knobs";
```
WickyNilliams: No knobs used, so this can be removed

src/components/QuestionnaireDesign/QuestionnaireDesign.test.js
```diff
@@ -72,4 +74,70 @@ describe("QuestionnaireDesign", () => {
...
+      section: { title: "" },
+      page: { id: "1", title: "" },
+      answers: [
+        {
+          id: "1"
+        }
+      ]
+    });
+
+    expect(wrapper.find(DeleteButton).nodes).toHaveLength(1);
```
WickyNilliams: This is also made redundant by the snapshot. See: https://github.com/ONSdigital/eq-author/pull/185/files#diff-a1d8eac033afab4fcc0399f20a6c7932R109

Any tests that assert on output can be better captured with snapshots.

src/components/QuestionnaireDesign/index.js
```diff
@@ -25,10 +26,15 @@ const PageTransition = props =>
     classNames="fade"
   />;
 
+const AnswerTransition = props =>
+  <CSSTransition {...props} timeout={duration} enter exit classNames="fade" />;
```
WickyNilliams: Instead of creating a new component, you could have PageTransition accept exit/enter as a prop:

```js
const PageTransition = props =>
  <CSSTransition
    timeout={duration}
    enter
    exit={false}
    classNames="fade"
    {...props}
  />;
```

Then use later like: 

```js
<PageTransition exit>
  {/* whatever */}
</PageTransition>
```

Also, I wonder if PageTransition should then be renamed FadeTransition, since it's more descriptive of what it is/does

src/components/QuestionnaireDesign/index.js
```diff
@@ -25,10 +26,15 @@ const PageTransition = props =>
...
+
 const AnimatedCanvasSection = styled(CanvasSection)`
   position: relative;
-  &.fade-enter {
+  &.fade-enter,
+  &.fade-exit {
     opacity: 0.25;
+    transform-origin: 50% 0%;
```
WickyNilliams: Don't need this if we're not using scale anymore

src/components/QuestionnaireDesign/index.js
```diff
@@ -182,8 +201,16 @@ class QuestionnaireDesign extends React.Component {
                     focused={focused === `answer-${answer.id}`}
                   >
                     <TextAnswer answer={answer} onChange={onChange} index={i} />
+                    <AnswerDeleteButton
+                      size="medium"
```
WickyNilliams: `"medium"` is default, so this can be removed

src/containers/QuestionnaireDesignPage/withDeleteAnswer.js
```diff
@@ -0,0 +1,45 @@
...
+const findById = (collection, id) => find(collection, { id: parseInt(id, 10) });
+
+export const deleteUpdater = (questionnaireId, sectionId, pageId, answerId) => (
+  proxy,
+  result
+) => {
+  const data = proxy.readQuery({
+    query: getQuestionnaireQuery,
+    variables: { id: questionnaireId }
+  });
```
WickyNilliams: Consider using `readFragment`/`writeFragment` for this. It makes things simpler as you don't have to "dive" down from questionnnaire to section to page to answer etc etc. It also makes tests simpler, as you don't need as much setup data.

See `withCreateAnswer`, if you want to see how it's used.

src/containers/QuestionnaireDesignPage/withDeleteAnswer.js
```diff
@@ -0,0 +1,45 @@
...
+  onDeleteAnswer(answerId) {
+    const variables = { id: answerId };
+    const update = deleteUpdater(
+      ownProps.questionnaireId,
+      ownProps.sectionId,
+      ownProps.pageId,
+      answerId
+    );
+
+    return Promise.resolve(mutate({ variables, update }));
```
WickyNilliams: `Promise.resolve` is redundant, `mutate` returns a promise anyway. So you can just `return mutate(...)`

src/containers/QuestionnaireDesignPage/withDeleteAnswer.test.js
```diff
@@ -0,0 +1,117 @@
...
+    };
+
+    onDeleteAnswer = jest.fn(() => Promise.resolve());
+
+    ownProps = {
+      questionnaire,
+      questionnaireId: questionnaire.id,
+      sectionId: currentSection.id,
+      pageId: currentPage.id,
+      history,
```
WickyNilliams: history isn't actually used anywhere, can be removed

samiwel:
[Delete Answer (3)](https://trello.com/c/T4479PtK/7-delete-answer-3)

PR #189, opened 2017-09-01
Number input component

samiwel:
image
This PR introduces a number spinner component.

### How to review 
All tests should pass.
Component should match the design.
Behaviour should work as expected (up button increases number, down button decreases number).


samiwel:
[Numeric input component (2)](https://trello.com/c/xyexdzAA/207-numeric-input-component-2)

src/components/Forms/Number/index.js
```diff
@@ -0,0 +1,116 @@
...
+export const SpinnerButton = styled.button`
+  border: 0;
+  background: none;
+`;
+
+class NumberInput extends React.Component {
+  constructor() {
+    super();
+    this.state = {
+      value: "0"
```
WickyNilliams: Why do you need state? Why can't you use the prop value?

src/components/Forms/Number/index.js
```diff
@@ -0,0 +1,116 @@
...
+    };
+  }
+
+  componentWillMount() {
+    if (this.props.value) {
```
WickyNilliams: If state is needed, this should go in the constructor: 

```js
constructor(props) {
    super(props);
    this.state = {
      value: props.value || "0"
   };
}
```

src/components/Forms/Number/index.js
```diff
@@ -0,0 +1,116 @@
...
+  transform: scaleY(-1);
+  filter: FlipV;
+  -ms-filter: "FlipV";
+`;
+
+const StyledInput = styled(Input)`
+  &[type=number]::-webkit-inner-spin-button, 
+  &[type=number]::-webkit-outer-spin-button { 
+      -webkit-appearance: none;
+      -moz-appearance: none;
```
WickyNilliams: Will this work? I can't imagine `-moz-appearance` would be applied given the parent selector is webkit-specific

src/components/Forms/Number/index.js
```diff
@@ -0,0 +1,116 @@
...
+      value: e.value
+    });
+    if (this.props.onChange) {
+      this.props.onChange(e);
+    }
+  };
+
+  render() {
+    return (
+      <NumberWrapper>
```
WickyNilliams: This article explains how the HTML should be structured to make this accessible: http://heydonworks.com/practical_aria_examples/#button-controlled-input

src/components/Forms/Number/numberinput.test.js
```diff
@@ -0,0 +1,70 @@
...
+    let numberWithSpinner;
+
+    beforeEach(() => {
+      numberWithSpinner = mount(
+        <Number id="number" onChange={handleChange} value={defaultValue} />
+      );
+    });
+
+    it("should render correctly with spinner buttons", () => {
+      expect(numberWithSpinner).toMatchSnapshot("withSpinner");
```
WickyNilliams: You don't need a name for the snapshot, it is inferred from the test name

src/components/Forms/Number/numberinput.test.js
```diff
@@ -0,0 +1,70 @@
...
+    });
+
+    it("should render correctly without spinner buttons", () => {
+      expect(numberWithSpinner).toMatchSnapshot("withoutSpinner");
```
WickyNilliams: This is no different to the withSpinner test? And again, no need for name

src/components/Forms/Number/numberinput.test.js
```diff
@@ -0,0 +1,70 @@
...
+    });
+
+    it("should set the input type to number", () => {
+      expect(numberWithSpinner.find("input").node.type).toEqual("number");
```
WickyNilliams: This is covered by the snapshot: https://github.com/ONSdigital/eq-author/pull/189/files#diff-3f05d2a27f47aa791992cc76845804e6R115

src/components/Forms/Number/numberinput.test.js
```diff
@@ -0,0 +1,70 @@
...
+    });
+
+    it("should pass `defaultValue` prop to component when type=number", () => {
+      expect(numberWithSpinner.find("input").node.value).toEqual("0");
```
WickyNilliams: This is covered by the snapshot: https://github.com/ONSdigital/eq-author/pull/189/files#diff-3f05d2a27f47aa791992cc76845804e6R116

src/components/Forms/Number/index.js
samiwel: Good point

src/components/Forms/Number/index.js
samiwel: Added aria labels as described in the article to both the Number input and the Spinner buttons 👍

src/components/Forms/Number/index.js
```diff
@@ -0,0 +1,120 @@
...
+
+import arrowIcon from "./arrow.svg";
+
+const NumberWrapper = styled.div`
+  display: flex;
+  flex-direction: row;
+`;
+
+const InvertedVertically = styled.img`
+  -moz-transform: scaleY(-1);
```
hamishtaplin: no need to use vendor prefixes any more

src/components/Forms/Number/index.js
```diff
@@ -0,0 +1,120 @@
...
+`;
+
+const SpinnerButtonWrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+`;
+
+export const SpinnerButton = styled.button`
+  border: 0;
+  background: none;
```
hamishtaplin: Use the `background` property for the icon rather than wrapping an `img` with this. Semantically, The `img` tag should be used for content rather than UI.

src/components/Forms/Number/index.js
```diff
@@ -0,0 +1,120 @@
...
+                alt={`Add ${this.props.step}`}
+              />
+            </SpinnerButton>
+            <SpinnerButton
+              type="button"
+              title={`Subtract ${this.props.step}`}
+              onClick={this.handleDown}
+              aria-controls={this.props.id}
+            >
+              <img
```
hamishtaplin: As mentioned above, use the `background` in CSS for the icon and then the content of the `button` can be a text label.

src/components/Forms/Number/index.js
```diff
@@ -0,0 +1,120 @@
...
+  -moz-transform: scaleY(-1);
+  -o-transform: scaleY(-1);
+  -webkit-transform: scaleY(-1);
+  transform: scaleY(-1);
+  filter: FlipV;
+  -ms-filter: "FlipV";
+`;
+
+const StyledInput = styled(Input)`
+  &[type=number]::-webkit-inner-spin-button,
```
hamishtaplin: Probably worth looking at browser support for this and wether the [autoprefixer](https://github.com/postcss/autoprefixer) implementation in styled-components has this selector covered.

src/components/Forms/Number/numberinput.test.js
```diff
@@ -0,0 +1,59 @@
+import React from "react";
+import { mount } from "enzyme";
+import Number, { SpinnerButton } from "components/Forms/Number";
+
+const defaultValue = "0";
+
+const handleChange = jest.fn();
```
hamishtaplin: Assign mock fns in the `beforeEach` function.

src/components/Forms/Number/index.js
```diff
@@ -0,0 +1,120 @@
...
+    return (
+      <NumberWrapper>
+        <StyledInput
+          type="number"
+          value={this.props.value}
+          onChange={this.props.onChange}
+          ref={this.setNumberInput}
+          aria-live="assertive"
+          role="alert"
+          {...this.props}
```
hamishtaplin: We should look at implementing (and documenting in the story) the `max` and `min` attributes.

src/components/Forms/Number/index.js
```diff
@@ -0,0 +1,120 @@
...
+  
+  width: 4em;
+`;
+
+const SpinnerButtonWrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+`;
+
+export const SpinnerButton = styled.button`
```
hamishtaplin: Although [this is debatable](https://medium.com/simple-human/buttons-shouldnt-have-a-hand-cursor-b11e99ca374b), I would add `cursor: pointer` here to get the "hand" icon.

Also add hover style (something like `opacity: 0.8` will do)

samiwel: I have added these styles.

src/components/Forms/Number/index.js
```diff
@@ -0,0 +1,131 @@
...
+  height: 18px;
+  background: url(${arrowIcon}) no-repeat;
+  background-size: 18px 18px;
+  cursor: pointer;
+  &:hover {
+    opacity: 0.6;
+    transition: opacity .2s ease-in-out;
+  }
+
+  transform: scaleY(${props => (props["inverted"] ? -1 : 1)});
```
WickyNilliams: `props["inverted"]` -> `props.inverted`

Surprised eslint doesn't pick up this. We should configure it to do so

src/components/Forms/Number/index.js
```diff
@@ -0,0 +1,131 @@
...
+      this.props.min,
+      this.props.max
+    );
+
+    const newValue =
+      isNaN(enteredValue) || enteredValue === -0
+        ? this.props.min || 0
+        : enteredValue;
+
+    this.props.onChange({ name, value: toString(newValue) });
```
WickyNilliams: Weird to use `toString(number)` here but `number.toString()` above

src/components/Forms/Number/index.js
```diff
@@ -0,0 +1,131 @@
...
+
+class Number extends React.Component {
+  handleUp = () => {
+    const name = this.props.name || this.props.id;
+    const value = clamp(
+      parseInt(this.props.value, 10) + 1,
+      this.props.min,
+      this.props.max
+    ).toString();
+    this.props.onChange({ name, value });
```
WickyNilliams: Couldn't this method call to `handleChange` to reuse the clamp logic? So it would become:

```js
  handleUp = () => {
    const name = this.props.name || this.props.id;
    const value = parseInt(this.props.value, 10) + 1;

    this.handleChange({ name, value });
  };
```

Same for `handleDown`.

src/components/Forms/Number/index.js
```diff
@@ -0,0 +1,131 @@
...
+    this.props.onChange({ name, value: toString(newValue) });
+  };
+
+  render() {
+    return (
+      <StyledDiv>
+        <StyledInput
+          {...this.props}
+          type="number"
+          value={this.props.value}
```
WickyNilliams: Unnecessary if you're already spreading props

src/components/Forms/Number/index.js
```diff
@@ -0,0 +1,131 @@
...
+  handleChange = ({ name, value }) => {
+    const enteredValue = clamp(
+      parseInt(value, 10),
+      this.props.min,
+      this.props.max
+    );
+
+    const newValue =
+      isNaN(enteredValue) || enteredValue === -0
+        ? this.props.min || 0
```
WickyNilliams: use defaultProps for min?

src/components/Forms/Number/numberinput.test.js
```diff
@@ -0,0 +1,135 @@
...
+        expect(handleChange).toBeCalledWith({
+          name: "number",
+          value: "100"
+        });
+      });
+
+      it("should not go over the max when changed", () => {
+        numberWithMinMax.setProps({
+          value: "101"
+        });
```
WickyNilliams: Seems an odd way to do this? Why setProps to a value > max, instead of `simulate("change", { name : "number", value: "101" })`?

Same below

src/components/Forms/Number/numberinput.test.js
```diff
@@ -0,0 +1,135 @@
...
+        numberWithMinMax.find("input").simulate("change");
+        expect(handleChange).toBeCalledWith({
+          name: "number",
+          value: "100"
+        });
+      });
+
+      it("should not go under the min", () => {
```
WickyNilliams: If you reused the logic in handleChange as i mention above, one of these "shoudl not go under min" tests become redundant

src/components/Forms/Number/index.js
```diff
@@ -0,0 +1,131 @@
...
+
+export const SpinnerButton = styled.button`
+  border: none;
+  width: 18px;
+  height: 18px;
+  background: url(${arrowIcon}) no-repeat;
+  background-size: 18px 18px;
+  cursor: pointer;
+  &:hover {
+    opacity: 0.6;
```
WickyNilliams: Spoke with Hamish about this, could you invert the styles? So that `opacity: 1` on hover and `opacity: 0.6` otherwise

src/components/Forms/Number/index.js
```diff
@@ -0,0 +1,131 @@
...
+  display: flex;
+  flex-direction: column;
+`;
+
+export const SpinnerButton = styled.button`
+  border: none;
+  width: 18px;
+  height: 18px;
+  background: url(${arrowIcon}) no-repeat;
+  background-size: 18px 18px;
```
WickyNilliams: `background-size: contain` is probably a better choice

src/components/Forms/Number/index.js
samiwel: Not sure we should default min. It's feasible that a number field might be used without a min and max.

src/components/Forms/Number/numberinput.test.js
samiwel: That's cleaner, didn't realise you could do this.

src/components/Forms/Number/numberinput.test.js
samiwel: Agreed, but they're testing the same code in two different ways. One is clicking the spinner, the other is testing a direct change to the input. I think it's worth leaving in there just as a belts and braces check.

src/components/Forms/Number/index.js
WickyNilliams: But it's effectively being given a default value here? Given the use-case we're building for, it's very unlikely we'd ever want negative numbers. If a use-case did arise which required it, you could pass `undefined`/`null`/`Number.NEGATIVE_INFINITY`/`Number.MIN_SAFE_INTEGER` as min prop

src/components/Forms/Number/index.js
```diff
@@ -0,0 +1,124 @@
...
+
+const SpinnerButtonWrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+`;
+
+export const SpinnerButton = styled.button`
+  border: none;
+  width: 18px;
+  height: 18px;
```
hamishtaplin: If you bump the `height` up to `20px` then the buttons will be better aligned with the box.

PR #198, opened 2017-09-11
Radio answer component

bennpearson:
Added radio answer component

### How to review 
Check story and run tests.
Check agains trello task: https://trello.com/c/Jds54DeE/164-add-edit-radio-button-answer-component-3


bennpearson:
[Add/edit radio button answer component (3)](https://trello.com/c/Jds54DeE/164-add-edit-radio-button-answer-component-3)

src/components/AnswerEditor/index.js
```diff
@@ -5,50 +5,45 @@ import styled from "styled-components";
...
-    } = this.props;
-
-    const Answer = AnswerTypeMap[answer.type];
+  renderAnswer(answer) {
+    switch (answer.type) {
+      case TEXTFIELD:
+        return <TextAnswer {...this.props} />;
+      case CHECKBOX:
+        return <MultipleChoiceAnswer type={CHECKBOX} {...this.props} />;
+      case RADIO:
```
samiwel: These could be combined since they both render the same component, `MultipleChoiceAnswer`.
e.g. 

```js
case CHECKBOX:
case RADIO:
    return <MultipleChoiceAnswer type={answer.type} {...this.props} />;
```

WickyNilliams: i did consider this, but wasn't sure whether it was clearer as it is. But I think you're right, that's better

src/components/AnswerEditor/index.js
```diff
@@ -5,50 +5,44 @@ import styled from "styled-components";
...
-    const Answer = AnswerTypeMap[answer.type];
+  renderAnswer(answer) {
+    switch (answer.type) {
+      case TEXTFIELD:
+        return <TextAnswer {...this.props} />;
+      case CHECKBOX:
+      case RADIO:
+        return <MultipleChoiceAnswer type={answer.type} {...this.props} />;
+      default:
+        return null;
```
samiwel: Do we want to throw an error here rather than return null? Returning null will just render a blank page? If we don't get a valid answer we should probably error rather trying to render nothing?

hamishtaplin: @samiwel In theory, this will never happen due to PropTypes for `answer` guarding against an invalid `type`: https://github.com/ONSdigital/eq-author/blob/cea60fb9068ac454040f9bb039e1acddf3012683/src/custom-prop-types/index.js#L27

WickyNilliams: I've actually hit an issue because of this... Added radio answer on a PR branch, switched to another branch which doesn't have radio support yet, and 💥 it blew up. `Radio` is valid in terms of it being an answer type, but it is not yet supported.

So perhaps we should throw a meaningful error, because at the moment it's some obscure react error being thrown

src/components/Answers/MultipleChoiceAnswer/Option.js
```diff
@@ -1,44 +1,41 @@
...
-
-  &:hover {
-    color: ${colors.darkGrey};
-    transition: color .2s ease-in-out;
-  }
 `;
 
-const StyledCheckboxInput = styled(Input)`
-  border: 2px solid ${colors.borders};
+const borderSizes = {
```
hamishtaplin: `borderRadii`? Worth it just to use the word `radii` 🤣

PR #201, opened 2017-09-12
Remove type from questionpage

samiwel:
Removes references to `type` field on QuestionType.
Updates to point at latest schema.
Update mock data store with latest schema changes.

### How to review 
All tests should pass.

### DIFF
--- a/src/components/QuestionProperties/QuestionProperties.test.js
+++ b/src/components/QuestionProperties/QuestionProperties.test.js
@@ -12,9 +12,7 @@ const handleSubmit = jest.fn();
 const handleChange = jest.fn();
 const handleBlur = jest.fn();
 
-const question = {
-  type: "General"
-};
+const question = jest.fn();
 
 describe("QuestionProperties", () => {
   beforeEach(() => {
@@ -51,6 +49,6 @@ describe("QuestionProperties", () => {
 
   it("should pass question type to select", () => {
     const selectValue = wrapper.find(Select).prop("defaultValue");
-    expect(selectValue).toEqual(question.type);
+    expect(selectValue).toBeUndefined();
   });
 });
--- a/src/components/QuestionProperties/QuestionProperties.test.js
+++ b/src/components/QuestionProperties/QuestionProperties.test.js
@@ -12,15 +12,12 @@ const handleSubmit = jest.fn();
 const handleChange = jest.fn();
 const handleBlur = jest.fn();
 
-const question = jest.fn();
-
 describe("QuestionProperties", () => {
   beforeEach(() => {
     wrapper = shallow(
       <QuestionProperties
         onUpdate={handleUpdate}
         loading={false}
-        question={question}
         onSubmit={handleSubmit}
         onChange={handleChange}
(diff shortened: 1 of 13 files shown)


samiwel:
[Author API - Remove QuesitonPage.type (2)](https://trello.com/c/4QIMS7P0/97-author-api-remove-quesitonpagetype-2)

src/custom-prop-types/index.js
```diff
@@ -12,9 +12,7 @@ export default {
     title: PropTypes.string,
     navigation: PropTypes.bool
   }),
-  question: PropTypes.shape({
-    type: PropTypes.string.isRequired
-  }),
+  question: PropTypes.object,
```
WickyNilliams: This can probably be deleted. QuestionProperties uses it, but it's also going to be deleted I believe

samiwel: Agree this can be removed eventually, but I think it should be removed as part of the work to remove QuestionProperties. Since QuestionProperties still expects something to be passed in right now.

samiwel: Have removed the PropType but left the QuestionProperties component where it is for now.

WickyNilliams: Yeah, sorry, that's what i was getting at. Delete both? QuestionProperties exists solely to edit QuestionType. No QuestionType, no QuestionProperties component. Don't see much/any value in doing that cleanup separately

WickyNilliams: Oh I see you removed it :)

src/components/QuestionProperties/QuestionProperties.test.js
```diff
@@ -12,9 +12,7 @@ const handleSubmit = jest.fn();
 const handleChange = jest.fn();
 const handleBlur = jest.fn();
 
-const question = {
-  type: "General"
-};
+const question = jest.fn();
```
WickyNilliams: This is no longer needed (and in any case, passing a function instead of an object isn't correct?)

src/components/QuestionProperties/QuestionProperties.test.js
```diff
@@ -51,6 +49,6 @@ describe("QuestionProperties", () => {
 
   it("should pass question type to select", () => {
     const selectValue = wrapper.find(Select).prop("defaultValue");
-    expect(selectValue).toEqual(question.type);
+    expect(selectValue).toBeUndefined();
```
WickyNilliams: This whole test can be removed

src/graphql/updatePage.graphql
```diff
@@ -3,20 +3,16 @@ mutation UpdateQuestionPage(
...
     description: $description
     guidance: $guidance
-    type: $type
   ) {
     id
     title
     description
     guidance
-    type
-    mandatory
```
WickyNilliams: I assume `mandatory` was accidentally left in in another PR?

samiwel: Yeah, it's no longer in the API, wonder if it's something that got lost in a rebase. Anyways it shouldn't be there.

PR #210, opened 2017-09-25
Add edit number answer

samiwel:
This PR enables authors to add number answers to the questionnaire.
There was a mismatch between the types supported on the back-end GraphQL api and the front end.
The GraphQL API has both `Integer` and `PositiveInteger` types defined in the schema. However the Author front-end only has a concept of `Number`.
For now, I have mapped `Number` -> `Integer` in the back-end. We probably need to look at an answer property to constrain input to positive integers, but that needs thinking about separately.

image

### How to review 
All tests should pass. Should be able to add a number answer to the questionnaire.


samiwel:
[Add/edit number answer (1)](https://trello.com/c/uNCIkX74/68-add-edit-number-answer-1)

PR #211, opened 2017-09-25
Improve graphql usage

WickyNilliams:
**DO NOT MERGE UNTIL API & SCHEMA MERGED/DEPLOYED**

* improves tests for queries and mutations
* use input types
  * update mutations to have single argument and use input types
* use readFragment/writeFragment everywhere we can
* use `newId` field
  * alias to `id`, to limit scope of required changes
  * updated any code that treated id as integer

### How to review 
review code, test against associated PRs.


WickyNilliams:
[Improve GraphQL usage (5)](https://trello.com/c/8bkFSzmj/187-improve-graphql-usage-5)

.storybook/webpack.config.js
```diff
@@ -0,0 +1,6 @@
+const path = require("path");
```
samiwel: Does `path` need to be here?

src/graphql/answerFragment.graphql
```diff
@@ -1,12 +1,7 @@
-fragment Answer on Answer {
-  id
+fragment BasicAnswer on Answer {
```
samiwel: Why have we named this `BasicAnswer` when we're getting the option Ids on `MultipleChoiceAnswer`?

WickyNilliams: Hmmm good point, didn't think of it being an overloaded term. I needed a unique name that wouldn't clash with https://github.com/ONSdigital/eq-author/blob/20bbcf0ab428e3790ca74c84e465d675f0998a7a/src/graphql/fragments/answer.graphql. 

Any ideas on a naming convention? I was trying to signify you're requesting only IDs, and not the whole thing

.storybook/webpack.config.js
WickyNilliams: Nope! Forgot to remove it :)

src/graphql/answerFragment.graphql
samiwel: `AnswerFragment`?

Issue #191, opened 2017-09-04
Checkbox values being lost upon adding/removing options


samiwel:
### Expected behaviour
The values entered should be persisted when new options are added and other options are deleted.

### Actual behaviour
The entered labels are being lost.

### Steps to reproduce the behaviour
Enter some values into the options of a Checkbox answer, add new options or remove some options.

### Technical information
Did some preliminary investigation and I believe it's down to the focus logic https://github.com/ONSdigital/eq-author/blob/master/src/containers/QuestionnaireDesignPage/QuestionnaireDesignPage.js#L71

### Screenshot
![checkbox-values-being-lost](https://user-images.githubusercontent.com/4097796/30027134-808d5e10-9177-11e7-89ef-799fdbb1f825.gif)

WickyNilliams:
Having tested this, I'm fairly certain we've fixed it via #188

PR #213, opened 2017-09-26
Checkbox option test

samiwel:
This PR implements a functional test for issue #191 which was fixed indirectly whilst refactoring data management #188.

### How to review 

### DIFF
--- a/tests/pages/CheckboxAnswer.page.js
+++ b/tests/pages/CheckboxAnswer.page.js
@@ -1,0 +1,52 @@
+const CHECKBOX_ANSWER_LABEL_INPUT = "input[data-qa='optionLabel']";
+const BUTTON_ADD_OTHER_OPTION = "button[data-qa='addAnotherOption']";
+const PAUSE_DURATION = 300;
+
+class CheckboxAnswerPage {
+  getOptionLabelElements() {
+    browser.waitForVisible(CHECKBOX_ANSWER_LABEL_INPUT);
+    return browser.elements(CHECKBOX_ANSWER_LABEL_INPUT);
+  }
+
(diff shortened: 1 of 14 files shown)


tests/pages/CheckboxAnswer.page.js
```diff
@@ -0,0 +1,55 @@
+const CHECKBOX_ANSWER_LABEL_INPUT =
+  "input[type='text'][class*='CheckboxOption']";
+const BUTTON_ADD_OTHER_OPTION =
+  "div[class*='CheckboxAnswerWrapper'] > div:nth-child(2) > button";
```
hamishtaplin: I think it might be better to add specific selectors for these tests (use id's) as these are potentially fragile.

samiwel:
Discussed with @hamishtaplin that it's better to use Id's where possible for this test to make the selectors less brittle.
Explained that this was not possible when writing the test since it would have meant re-writing history (adding the Id's on a past commit and having to rebase the entire history in a different direction).
As a compromise, decided that I will change the test so that Id's are used instead of complex DOM selectors, but the test would no longer work on the past commits.
This should be fine since the test should be there in order to prevent regression rather than prove that something didn't work in the past.

samiwel:
The new functional test for issue 191 works fine when run locally using chrome headless, but for some reason it is failing when running on Travis.
I have made several changes to try and remedy this without success. For now, I think the best thing to do would be to merge this PR with the failing test skipped (see [0975268](https://github.com/ONSdigital/eq-author/pull/213/commits/097526864ef5647d3e8832f6910b3a856abf518f), until we find out why the test fails on Travis.

PR #219, opened 2017-09-28
142 scrolling behaviour

hamishtaplin:

[Trello card](https://trello.com/b/GOqjEepx/eq-author)

Adds `ScrollPanel` component that will add vertical scroll bars if it's children height exceeds the height of it's container.

Additionally, "add section" button is now sticky, should the navigation exceed the height of the browser it will "stick" to the bottom of the viewport.

Also, fixed a lingering issue with the `Column` component having `max-width: NaN` in CSS.

### How to review 

- add enough answers/other content to a page to invoke scrollbars on main content area.
- add enough pages/sections to invoke scrollbars on navigation

### DIFF
--- a/src/components/ScrollPane/index.js
+++ b/src/components/ScrollPane/index.js
@@ -7,7 +7,7 @@ const MainCanvas = styled.div`
   position: relative;
   background-color: rgba(0, 0, 0, 0);
   -webkit-background-clip: text;
-  transition: background-color 0.8s;
+  transition: background-color 0.2s;
 
   ::-webkit-scrollbar-thumb {
     border-radius: 0;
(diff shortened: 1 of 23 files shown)


src/components/ScrollPane/index.js
```diff
@@ -0,0 +1,37 @@
+import styled from "styled-components";
+
+const MainCanvas = styled.div`
```
WickyNilliams: `const ScrollPane`?

WickyNilliams:
image

There's a weird "ghost" of the "QUESTIONNAIRE STRUCTURE" heading on hover, when the sidebar overflows and `scrollTop > 0`

PR #220, opened 2017-09-28
Add link to section headers which takes author to first question.

samiwel:
A link has been added around the section title which takes the user to the first page when clicked.

### How to review 


samiwel:
[Make Section Titles Clickable (2)](https://trello.com/c/gGIvuVLA/177-make-section-titles-clickable-2)

src/components/QuestionnaireNav/SectionNav.js
```diff
@@ -65,23 +77,33 @@ const NavList = styled.ol`
   list-style: none;
 `;
 
-const SectionNav = ({ questionnaire, onAddPage, onDeletePage }) =>
+export const getFirstPage = section => first(section.pages) || { id: "-1" };
```
WickyNilliams: instead of returning a sort of "dummy" page, why not disable the link if section has no pages?

src/components/QuestionnaireNav/SectionNav.js
```diff
@@ -35,6 +40,13 @@ const SectionItem = styled.li`
   }
 `;
 
+const Link = styled(NavLink)`
+  text-decoration: none;
+  &:visited {
+    color: ${colors.text};
+  }
```
WickyNilliams: Not sure if it's related to this, but when i delete a page the section text briefly flashes blue. Weirdly only happens first time you delete a page...

kapture 2017-09-29 at 16 26 16

src/containers/QuestionnaireMetaPage/index.js
```diff
@@ -1,7 +1,7 @@
 import { compose } from "react-apollo";
 import { connect } from "react-redux";
 
-import getUrlParams from "utils/getUrlParams";
+import { getUrlParams } from "utils/URLUtils";
```
hamishtaplin: Can we make the filename consistent with the function naming? so `utils/UrlUtils`.

PR #221, opened 2017-09-29
Add delete section feature

samiwel:
deletesection1
Added delete button to section titles and implemented the Apollo GraphQL logic to perform the deletion and optimistic re-fetching of data.
If deleting the last section in a questionnaire, then a new section and page combination will be added similar to how page deletion works.

### How to review 
All tests and automated checks should pass.
Should be able to delete sections without error.

### Dependencies
This PR depends on #220 being merged first. This will prevent both the section and page delete buttons being displayed at the same time.


samiwel:
[Delete Section (2)](https://trello.com/c/5byDRIfK/215-delete-section-2)

src/components/QuestionnaireNav/SectionNav.js
```diff
@@ -65,23 +67,60 @@ const NavList = styled.ol`
   list-style: none;
 `;
 
-const SectionNav = ({ questionnaire, onAddPage, onDeletePage }) =>
+export const DeleteButton = styled.button`
```
WickyNilliams: Reuse code from PageNav?

src/components/QuestionnaireNav/SectionNav.js
```diff
@@ -65,23 +67,60 @@ const NavList = styled.ol`
...
+  color: ${colors.text};
+  border: none;
+  background: transparent;
+  font-size: 1.25em;
+  margin-right: 0.25em;
+  position: absolute;
+  right: 0;
+  top: 1em;
+  z-index: 3;
```
WickyNilliams: z-index issue:

image

src/containers/QuestionnaireNavContainer/index.js
```diff
@@ -11,5 +12,6 @@ export default compose(
   withRouter,
   withCreatePage,
   withDeletePage,
-  withCreateSection
+  withCreateSection,
+  withDeleteSection
```
WickyNilliams: Can you add a comment here about `withDeleteSection` needing to appear after `withCreateSection`. There used to be a similar one for deleting page, not sure where the note went (might be worth adding that back as well)

src/components/QuestionnaireNav/SectionNav.js
samiwel: @hamishtaplin Are you able to assist with this? I can't seem to get the tooltip text to appear outside of the scroll pane boundary. I've tried adjusting `z-index` on the Tooltip itself and also adjusting the `overflow-x: visible` on the scroll pane div but no joy.

src/components/QuestionnaireNav/PageNav.js
```diff
@@ -102,7 +102,7 @@ const NavList = styled.ol`
   list-style: none;
 `;
 
-export const DeleteButton = styled.button`
+export const SharedDeleteButtonStyles = css`
```
WickyNilliams: Can we do something like this, where we have an actual DeleteButton component and then extend is as necessary:

```js
// components/QuestionnaireNav/DeleteButton.js

export default styled.button`
  color: ${colors.text};
  border: none;
  background: transparent;
  font-size: 1.25em;
  margin-right: 0.25em;
  position: absolute;
  right: 0;
  top: 50%;
  z-index: 3;
  cursor: pointer;
  transform: translateY(-50%) translateX(50%);
  opacity: 0;
  transition: transform 0.1s ease-in, opacity 0.1s ease-in;
`;

// components/QuestionnaireNav/PageNav.js

const PageDeleteButton = DeleteButton.extend`
  ${StyledPageItem}:hover &,
  ${Link}:focus + &,
  &:focus {
    opacity: 1;
    transform: translateY(-50%) translateX(0);
  }
`;
```

hamishtaplin:
Can we look at the way transitions work when you delete a section? Currently, it's animating all the other sections back in again and it's pretty confusing.

src/components/QuestionnaireNav/SectionNavItem.js
```diff
@@ -0,0 +1,197 @@
...
+  saveRef = elem => {
+    this.elem = elem;
+  };
+
+  render() {
+    const { questionnaire, section, onDeletePage, className } = this.props;
+
+    return (
+      <li ref={this.saveRef} style={this.state.style} className={className}>
+        <div className="section-title-wrapper">
```
hamishtaplin: Why not a styled-component here?

PR #222, opened 2017-10-02
increase test coverage of non-component JS

hamishtaplin:

[Trello](https://trello.com/c/gXHdOcdB/127-test-coverage-for-non-component-javascript-2)

bumps the test coverage for non-component JS

### How to review 

- suggestions on other tests that could be written would be useful


src/apollo/__snapshots__/fragmentMatcher.test.js.snap
```diff
@@ -0,0 +1,14 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`apollo/fragmentMatcher should not change unexpectedly 1`] = `
+IntrospectionFragmentMatcher {
```
samiwel: I think the fragmentMatcher can be removed altogether.

samiwel:
[Test coverage for non-component javascript (2)](https://trello.com/c/gXHdOcdB/127-test-coverage-for-non-component-javascript-2)

src/apollo/__snapshots__/fragmentMatcher.test.js.snap
samiwel: Actually, I just tried this and you get a warning in the JS console. May as well leave as it.

PR #223, opened 2017-10-04
RTE integration

WickyNilliams:
This integrates the RTE into the designer. 

* HTML is sent to the server rather than draftjs data structure.
* It ensures no HTML is displayed in e.g. sidebar nav or web page title (`<title />` tag).
* Changes `SeamlessTextArea` to use new component `WrappingTextArea` which acts like a textarea crossed with input - it wraps multiple lines, but disallows carriage returns.

### How to review 
Run tests, review code, test features


samiwel:
Some weird z-index issue with the add answer selector if you enter some guidance text.

image

samiwel:
Question title was large and bold previously (a bit like the Heading text) but after RTE integration the question text is small and not bold.
See below (left: before RTE was introduced running in CloudFoundry, right: after RTE was introduced)
image

PR #224, opened 2017-10-04
Fix error observed on admin page when running with the mockAPI.

samiwel:
Fixes an error when navigating to the admin page when running with the mockAPI.
The fix involved specifying a reducer for the scalar GraphQL Date type and ensuring that time values do not make it into the localStorage data store.

### How to review 
Should no longer see the error on the admin page after creating a questionnaire when running with the mockAPI.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "homepage": "./",
   "scripts": {
     "start": "yarn && node scripts/start.js",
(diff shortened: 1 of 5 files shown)


samiwel:
@WickyNilliams be good to test the third merging option with this PR?

PR #231, opened 2017-10-10
185 Cascading menu component

hamishtaplin:

[Trello card](https://trello.com/c/yWibADHi/185-cascading-menu-component-8)

- Added `Menu` component wrapping [react-menu-list](https://github.com/StreakYC/react-menu-list) with styling
- Added a basic `QuestionnaireMenu` implementation

This PR doesn't implement the menu inside the RTE toolbar, however.

### How to review 

- review `Menu` components in the relevant story
- `QuestionnaireMenu` is not considered production-ready, just a reference implementation.

### DIFF
--- a/src/components/Menu/Menu.test.js
+++ b/src/components/Menu/Menu.test.js
@@ -1,0 +1,62 @@
+import React from "react";
+import { mount } from "enzyme";
+
+import {
+  Dropdown,
+  MenuItem,
+  MenuList,
+  SubMenuItem,
+  MenuScrollPane
+} from "components/Menu";
+
+let handleItemChosen, wrapper;
+
+describe("components/Menu", () => {
+  beforeEach(() => {
+    handleItemChosen = jest.fn();
+    wrapper = mount(
+      <Dropdown>
+        <MenuScrollPane>
+          <MenuList>
+            <MenuItem id="item-1" onItemChosen={handleItemChosen}>
+              1. Fusce Egestas Tristique Porta Sollicitudin
+            </MenuItem>
+            <MenuItem id="item-2" onItemChosen={handleItemChosen}>
+              2. Fusce Egestas Tristique Porta Sollicitudin
+            </MenuItem>
+            <MenuItem id="item-3" onItemChosen={handleItemChosen}>
+              3. Fusce Egestas Tristique Porta Sollicitudin
+            </MenuItem>
+            <SubMenuItem
+              menu={
+                <Dropdown>
+                  <MenuList>
+                    <MenuItem id="item-4.1" onItemChosen={handleItemChosen}>
+                      4.1. Fusce Egestas Tristique Porta Sollicitudin
+                    </MenuItem>
+                    <MenuItem id="item-4.2" onItemChosen={handleItemChosen}>
+                      4.2. Ligula Consectetur Ullamcorper Cursus Ligula
+                      Consectetur Ullamcorper Cursus
+                    </MenuItem>
+                  </MenuList>
+                </Dropdown>
+              }
+            >
+              4. Fusce Egestas Tristique Porta Sollicitudin
+            </SubMenuItem>
+          </MenuList>
+        </MenuScrollPane>
+      </Dropdown>
+    );
+  });
+
+  it("should render a Menu", () => {
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it("should handle items being chosen by calling onItemChosen with id prop", () => {
+    const menuItem = wrapper.find(MenuItem).first();
+    menuItem.simulate("click");
+    expect(handleItemChosen).toHaveBeenCalledWith(menuItem.props().id);
+  });
+});
--- a/src/components/Menu/MenuItem.js
+++ b/src/components/Menu/MenuItem.js
@@ -1,0 +1,60 @@
+import React from "react";
+import { MenuItem as RMLMenuItem } from "react-menu-list";
+import styled, { css } from "styled-components";
+import PropTypes from "prop-types";
+import { colors } from "constants/theme";
+
+export const highlightedClassName = "is-highlighted";
+export const menuItemStyles = css`
+  overflow: hidden;
+  /* stylelint-disable value-no-vendor-prefix */
+  display: -webkit-box;
+  -webkit-line-clamp: ${props => props.lines};
+  -webkit-box-orient: vertical;
+  /* stylelint-enable */
+  cursor: pointer;
+  user-select: none;
+  position: relative;
+  line-height: 1.3;
+  font-size: 0.8em;
+  padding: 0 1em;
+
+  &.${highlightedClassName} {
+    background: ${colors.lighterGrey};
+  }
+`;
+
+export const MenuItemInner = styled.div`
+  margin: 0.5em 0;
+  overflow: hidden;
+`;
+
+const StyledMenuItem = styled(RMLMenuItem)`
+  ${menuItemStyles};
+`;
+
+const MenuItem = ({ children, onItemChosen, id, ...otherProps }) => {
+  const handleItemChosen = e => onItemChosen(id);
+  return (
+    <StyledMenuItem
+      {...otherProps}
+      highlightedClassName={highlightedClassName}
+      onItemChosen={handleItemChosen}
+    >
+      <MenuItemInner>{children}</MenuItemInner>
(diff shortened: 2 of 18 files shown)


hamishtaplin:
[Cascading menu component (8)](https://trello.com/c/yWibADHi/185-cascading-menu-component-8)

src/components/QuestionnaireMenu/QuestionnaireMenu.test.js
```diff
@@ -0,0 +1,27 @@
...
+
+const answer = { id: "4", title: "Answer" };
+const page = { id: "2", title: "Page", answers: [answer] };
+const section = { id: "3", title: "Section", pages: [page] };
+const questionnaire = { id: "1", title: "Questionnaire", sections: [section] };
+
+describe("components/QuestionnaireMenu", () => {
+  beforeEach(() => {
+    handleItemChosen = jest.fn();
+    wrapper = mount(
```
WickyNilliams: Why mount?

src/components/QuestionnaireMenu/index.js
```diff
@@ -0,0 +1,87 @@
...
+      </MenuScrollPane>
+    </MenuList>
+  </Dropdown>
+);
+
+SectionMenu.propTypes = {
+  questionnaire: CustomPropTypes.questionnaire
+};
+
+const QuestionnaireMenu = ({ ...props }) => <SectionMenu {...props} />;
```
WickyNilliams: What's the reason for this vs just exporting `SectionMenu`?

src/components/QuestionnaireMenu/index.js
```diff
@@ -0,0 +1,87 @@
...
+);
+
+PageMenu.propTypes = {
+  pages: PropTypes.arrayOf(CustomPropTypes.page).isRequired
+};
+
+const AnswerMenu = ({ answers, ...otherProps }) => (
+  <Dropdown maxWidth={"10em"}>
+    <MenuList>
+      {answers &&
```
WickyNilliams: lodash's `map` will handle undefined/null, so you don't need `answers &&`

src/components/QuestionnaireMenu/index.js
```diff
@@ -0,0 +1,87 @@
...
+  MenuList,
+  SubMenuItem,
+  MenuScrollPane
+} from "components/Menu";
+
+const PageMenu = ({ pages, ...otherProps }) => (
+  <Dropdown maxWidth={"25em"}>
+    <MenuList>
+      {pages &&
+        pages.map((page, i) => {
```
WickyNilliams: lodash's `map` will handle undefined/null, so you don't need `pages &&`

src/components/Menu/Menu.test.js
```diff
@@ -0,0 +1,62 @@
...
+
+let handleItemChosen, wrapper;
+
+describe("components/Menu", () => {
+  beforeEach(() => {
+    handleItemChosen = jest.fn();
+    wrapper = mount(
+      <Dropdown>
+        <MenuScrollPane>
+          <MenuList>
```
WickyNilliams: In some places you have `Dropdown > MenuList > MenuScrollPane`. Here you have `Dropdown > MenuScrollPane > MenuList`. Others you have only `Dropdown > MenuList`.

I wonder if we can fold `Dropdown` into `MenuList` itself, to save having to remember to pair them. And then if we can always have a `MenuScrollPane`? What happens if there's a scroll pane and no `max-height`?

public/index.html
```diff
@@ -24,4 +24,6 @@
   <body>
     <div id="root"></div>
   </body>
+
+
```
WickyNilliams: ??

src/components/Menu/Menu.test.js
```diff
@@ -0,0 +1,62 @@
...
+  SubMenuItem,
+  MenuScrollPane
+} from "components/Menu";
+
+let handleItemChosen, wrapper;
+
+describe("components/Menu", () => {
+  beforeEach(() => {
+    handleItemChosen = jest.fn();
+    wrapper = mount(
```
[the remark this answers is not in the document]
hamishtaplin: to snapshot the styles, also the second test relies on `mount`.

src/components/Menu/Menu.test.js
hamishtaplin: Inconsistencies with the order aside, I was copying the structure that `react-menu-list` uses as I think it's easier to compose. Eg. bits that are scrollable with bits that aren't:

```<Dropdown>
  <MenuList>
    <MenuItem>Next page</MenuItem>
    <MenuItem>Previous page</MenuItem>
  </MenuList>
  <MenuScrollPane>
    <MenuList>
      <MenuItem>
        1. Fusce Egestas Tristique Porta Sollicitudin
      </MenuItem>
      <MenuItem>
        2. Fusce Egestas Tristique Porta Sollicitudin
      </MenuItem>
      <MenuItem>
        3. Fusce Egestas Tristique Porta Sollicitudin
      </MenuItem>
    </MenuList>
  </MenuScrollPane>
</Dropdown>```

File: src/components/Menu/MenuItem.js [hunk 8]
```diff
@@ -0,0 +1,60 @@
+import React from "react";
+import { MenuItem as RMLMenuItem } from "react-menu-list";
+import styled, { css } from "styled-components";
+import PropTypes from "prop-types";
+import { colors } from "constants/theme";
+
+export const highlightedClassName = "is-highlighted";
```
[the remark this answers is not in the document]
hamishtaplin: Both `MenuItem` and `SubMenuItem` pass it down to the underlying `react-menu-list` components as it's added to the elements `classList` when highlighted.

src/components/Menu/MenuList.js
```diff
@@ -0,0 +1,8 @@
+import { MenuList } from "react-menu-list";
+import PropTypes from "prop-types";
+
+MenuList.propTypes = {
+  children: PropTypes.node.isRequired
+};
```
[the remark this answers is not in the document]
hamishtaplin: I think I just made a wrapper for each component in `react-menu-list` and then forgot about this. It's convenient to import them all from the same place so I guess this (or the `index.js) could just re-export the `react-menu-list` component?

src/components/Menu/Menu.test.js
WickyNilliams: I'm just wondering if there are certain pairings that always happen e.g. is `MenuList` always the single and direct descendent of `MenuScrollPane`? If so, we could introduce ScrollableMenuList or something which wraps those two components.

What happens if you have a scroll pane where the contents is smaller than max-height? What happens if you have a scroll pane and do not set a max-height?

src/components/QuestionnaireMenu/QuestionnaireMenu.test.js
hamishtaplin: to snapshot the styles, otherwise they aren't covered

src/components/Menu/Menu.test.js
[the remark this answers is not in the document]
WickyNilliams: Why does the other test rely on it?

src/components/QuestionnaireMenu/index.js
hamishtaplin: We do it like this throughout the app though (consistency), isn't it better to be explicit about the fact the object might be undefined? Happy to use lodash where savings can be made but this is pretty simple.

src/components/QuestionnaireMenu/index.js
hamishtaplin: I think I must have removed some stuff and then not noticed it could be simplified.

src/components/Menu/MenuList.js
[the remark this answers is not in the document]
WickyNilliams: I think re-export the component is probably better - less code for us to look after!

src/components/Menu/Menu.test.js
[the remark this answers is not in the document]
hamishtaplin: `shallow` doesn't render the actual `react-menu-list` component, which is needed for the `click` event to work in that test. Otherwise, there's no way of testing that `onItemChosen` gets called.

src/components/Menu/Menu.test.js
hamishtaplin: ok cool, I will look at some options

PR #240, opened 2017-10-19
214 toast component

WickyNilliams:
Adds Toast and ToastList components. 

### How to review 
Review code, run tests, play with stories

### DIFF
--- a/src/components/Toast/Transition.js
+++ b/src/components/Toast/Transition.js
@@ -5,13 +5,14 @@ import PropTypes from "prop-types";
 const ToastTransition = styled(CSSTransition).attrs({
   classNames: "toast"
 })`
-  transition: transform ${props => props.timeout}ms ease-out,
-    opacity ${props => props.timeout}ms ease-out;
-  transform: translateY(0);
+  transition-property: transform, opacity;
+  transition-duration: ${props => props.timeout * 0.66}ms;
 
   &.toast-enter {
     opacity: 0;
     transform: translateY(50%);
+    transition-timing-function: cubic-bezier(0.73, 0.02, 0.32, 1.49);
+    transition-delay: ${props => props.timeout * 0.33}ms;
   }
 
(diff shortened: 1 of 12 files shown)


src/components/Toast/Transition.js
```diff
@@ -0,0 +1,43 @@
+import styled from "styled-components";
+import { CSSTransition } from "react-transition-group";
+import PropTypes from "prop-types";
+
+const ToastTransition = styled(CSSTransition).attrs({
+  classNames: "toast"
+})`
+  transition: transform ${props => props.timeout}ms ease-out,
+    opacity ${props => props.timeout}ms ease-out;
```
hamishtaplin: Can we play around with this just to improve the sense of the toasts moving out of the way to accommodate new ones? Using delays/durations as discussed.

src/components/Toast/story.js
```diff
@@ -0,0 +1,127 @@
...
+            <div>Hello</div>
+          </Toast>
+          <Toast id="2">
+            <div>World</div>
+          </Toast>
+        </ToastList>
+      </StoryInner>
+    </StoryContainer>
+  ))
+  .add("Interactive", () => <Story />);
```
hamishtaplin: Worth adding a "disappears on use" story to satisfy the relevant AC?

samiwel:
Toast component looks good to me in general. Although I guess it wouldn't be a major issue in practice, wonder if we should limit the maximum number of toasts shown on the screen and queue up any extra ones. Otherwise you run the risk of having something like this. 
Too much toast?

WickyNilliams:
@samiwel i can't imagine that happening in practice. In any case, I think logic for something like queuing is better placed outside of the Toast component itself, perhaps in a container component (e.g. `mapStateToProps` could select only the 5 oldest notifications at a time). This keeps Toast dumb but flexible

PR #242, opened 2017-10-20
Remove the properties panel from questionnaire design page.

samiwel:
This PR removes the properties panel from the right hand side of the Questionnaire Design page.
The panel will be added back once a decision has been made about which properties should be displayed.

### How to review 
Observe that the properties panel is no longer visible.


samiwel:
[Hide the properties panel (0.5)](https://trello.com/c/xdQ3OLnV/280-hide-the-properties-panel-05)

PR #243, opened 2017-10-20
275 Add fullstory service to production environment

hamishtaplin:

[Trello card](https://trello.com/c/r3EFP5Kc/275-add-fullstory-service)

Adds the Javascript snippet to enable FullStory service to the production environment by way of an environment variable `REACT_APP_USE_FULLSTORY`.

### How to review 

- check service is not enabled in development environments by attempting to access the `_fs_namespace` global in the browser console.
- check it is enabled in prod through the same method


samiwel:
[Add FullStory service](https://trello.com/c/r3EFP5Kc/275-add-fullstory-service)

public/index.html
```diff
@@ -26,6 +26,25 @@
     -->
     <title>Author</title>
   </head>
+  <script type="text/javascript">
+    if ("%REACT_APP_USE_FULLSTORY%" === "true") {
```
WickyNilliams: Does this HTML get put through uglify?

WickyNilliams: As a note for future: we should probably reconsider this approach when we have more time. Manually embedding scripts in HTML means we lose out on a lot of niceties of the build process etc. 

We should consider something like having multiple entrypoints output by webpack - one for main app, and one for various services we want to integrate with. We can then have webpack/uglify strip code for any services we don't use in any environment

WickyNilliams: Actually, it seems this code _is_ run through uglify. So the `if` statement is completely removed in production. However I notice in dev that the `%REACT_APP_USE_FULLSTORY%` value isn't replaced at all. Unsure why.

hamishtaplin: @WickyNilliams Did you restart the app after adding the relevant env var? Caught me out a couple of times.

If it's being stripped out in prod then it's pointless anyway—will look at other options.

WickyNilliams: To be clear, I meant that the if statement is removed, but its contents are kept. e.g.

```js
// this
if(foo) {
  alert("foo!");
}

// becomes this
alert("foo!");
```

PR #244, opened 2017-10-23
Disable Fullstory on Travis

hamishtaplin:
Describe what you have changed and why, link to other PRs or Issues as appropriate.

### How to review 
Describe the steps required to test the changes (include screenshots if appropriate).

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,8 @@
     "start": "yarn && node scripts/start.js",
     "start:mockAPI": "REACT_APP_USE_MOCK_API=true node scripts/start.js",
     "build": "node scripts/build.js",
-    "build:mockAPI": "REACT_APP_USE_MOCK_API=true node scripts/build.js",
+    "build:mockAPI":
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -7,7 +7,8 @@
     "start": "yarn && node scripts/start.js",
     "start:mockAPI": "REACT_APP_USE_MOCK_API=true node scripts/start.js",
     "build": "node scripts/build.js",
-    "build:mockAPI": "REACT_APP_USE_MOCK_API=true node scripts/build.js",
+    "build:mockAPI":
+      "REACT_APP_USE_MOCK_API=true REACT_APP_USE_FULLSTORY=false node scripts/build.js",
```
WickyNilliams: I think the logic should be inverted.

There is only one env where we actually want it enabled (real production deploy), and many envs where we don't (netlify, travis). Therefore it should be off by default (remove `REACT_APP_USE_FULLSTORY="true"` from `.env.production`) and should be opted into where necessary.

I assume this would be configured in a docker file. @samiwel thoughts?

hamishtaplin:
Closing this as it's the wrong approach.

PR #245, opened 2017-10-23
Disable Fullstory env vars in dotfiles

hamishtaplin:

Sets `REACT_APP_USE_FULLSTORY` to default `"false"` everywhere so that it can be set to `"true"` in the only deploy environment we want it in: ECS.

### How to review 

Fullstory should not run in Netlify preview build or on Travis.


samiwel:
[Add FullStory service](https://trello.com/c/r3EFP5Kc/275-add-fullstory-service)

PR #246, opened 2017-10-23
Test and coverage in a single step on Travis CI.

samiwel:
I think the issue we were having with invalid snapshots (due to a bug in the istanbul library) has been resolved in the latest versions.
A quick win for improving the Travis build times would be to combine the test and coverage tasks into a single step.
This should hopefully shave about a minute off our build times.

### How to review 
Build should pass/fail in expected ways.
Tests should run as expected and coverage report should be generated in a single pass.


samiwel:
[Investigate how to improve Travis build times](https://trello.com/c/pd4OsKM4/228-investigate-how-to-improve-travis-build-times)

PR #253, opened 2017-10-30
Display toasts in author and allow undo.

samiwel:
This PR adds toasts and undo capability to the author application.
It uses the Redux store to raise and dismiss toast messages whenever a destructive action, such as deleting a Questionnaire, Section, Page or Answer.
Note: A toast is not displayed when deleting options, as per the acceptance criteria, because it was felt that options are easy enough to add back in.

### How to review 
This PR requires some changes to the API. See the Pull request [here](https://github.com/ONSdigital/eq-author-api/pull/44).
One running, it should be possible to delete items of the questionnaire and observe that the toast message is displayed at the bottom of the screen.
There is a limited window where the user can click on the undo button in the toast message, before the toast disappears. Hovering/focusing over the toast message ensures that the timeout is paused.
Clicking the undo button in the toast should restore the deleted element to its original place.


samiwel:
[Undo Deletes (8)](https://trello.com/c/2ch8mhgt/205-undo-deletes-8)

src/containers/enhancers/withDeleteSection.js
```diff
@@ -1,9 +1,10 @@
 import { graphql } from "react-apollo";
 import deleteSectionMutation from "graphql/deleteSection.graphql";
-import { remove } from "lodash";
+import { remove, find } from "lodash";
 import fragment from "graphql/questionnaireFragment.graphql";
 import getNextSection from "utils/getNextOnDelete";
 import { getLink } from "utils/UrlUtils";
+import pluralize from "pluralize";
```
WickyNilliams: This lib seems a bit much since, you can effectively do the same with:

src/actionCreators/undelete/undeleteAnswer.js
```diff
@@ -0,0 +1,65 @@
...
+    update: createUpdate(context)
+  });
+
+export const undeleteAnswer = (id, context) => {
+  return (dispatch, getState, client) => {
+    const undelete = createUndelete(createMutate(client, query));
+
+    dispatch(undeleteRequest());
+
+    return undelete(id.replace("Answer", ""), context)
```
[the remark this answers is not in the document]
samiwel: It should never happen in the testing that I've done. However, you make a good point. I've removed the string replacement altogether in favour of just using the `answerId` which is on the context anyway.

src/containers/ToastContainer/index.js
```diff
@@ -0,0 +1,86 @@
...
+  dismissToast: PropTypes.func.isRequired,
+  undoToast: PropTypes.func.isRequired
+};
+
+export default compose(
+  connect(null, UndeleteQuestionnaireActions),
+  connect(null, UndeleteSectionActions),
+  connect(null, UndeletePageActions),
+  connect(null, UndeleteAnswerActions),
+  connect(mapStateToProps, ToastActionCreators)
```
WickyNilliams: How about:

src/containers/enhancers/withDeleteSection.js
```diff
@@ -51,11 +74,17 @@ export const mapMutateToProps = ({ ownProps, mutate }) => ({
...
+        handleDeletion(ownProps, ownProps.questionnaireId, sectionId).then(
+          () => res
+        )
       )
-    );
+      .then(res =>
+        displayToast(ownProps, ownProps.questionnaireId, sectionId).then(
+          () => res
+        )
+      );
```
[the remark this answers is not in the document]
samiwel: Yeah this way is so much cleaner. The reason for returning `Promise.resolve` was specifically for chaining the result. Storing reference to the original promise cleans this up quite a bit.

src/containers/enhancers/withDeleteAnswer.js
```diff
@@ -32,10 +32,14 @@ export const mapMutateToProps = ({ ownProps, mutate }) => ({
...
-    return mutate({
+    const mutation = mutate({
       variables: { input: answer },
       update
-    }).then(res => displayToast(ownProps, pageId, answerId).then(() => res));
+    });
+
+    return mutation.then(res =>
+      displayToast(ownProps, pageId, answerId).then(() => mutation)
+    );
```
WickyNilliams: These can be written as:

```js
mutation
  .then(() => displayToast(ownProps, questionnaireId))
  .then(() => mutation);
```

src/containers/enhancers/withDeleteAnswer.js
```diff
@@ -16,15 +16,28 @@ export const deleteUpdater = (pageId, answerId) => (proxy, result) => {
   });
 };
 
-export const mapMutateToProps = ({ mutate }) => ({
+export const displayToast = (ownProps, pageId, answerId) => {
+  ownProps.raiseToast(`Answer${answerId}`, "Answer deleted", "undeleteAnswer", {
```
WickyNilliams: Is there any reason that the entity's ID is used for the toast ID? The entity's ID might be something you want to keep in the action's context, but the ID for the toasts themselves should be independent of the entities' (and generated in the raiseToast action creator)

samiwel: Only that it's unique and predictable.

WickyNilliams: It's typically the actionCreator that generates an ID, see here: https://redux.js.org/docs/basics/ExampleTodoList.html#actionsindexjs. I guess it's fine for now though

src/actionCreators/undelete/undeleteAnswer.test.js
```diff
@@ -0,0 +1,88 @@
...
+      expect(proxy.writeFragment).toHaveBeenCalledWith({
+        id: "QuestionPage1",
+        fragment,
+        data: {
+          answers: [{ id: 1 }]
+        }
+      });
+    });
+  });
+});
```
WickyNilliams: I don't think these tests are testing the right things. Right now, I could change all the actions that get dispatched and nothing would fail.

There are no assertions that certain action have been dispatched, and in what order, etc. Stuff like `"should return a thunk function"` would be implicitly tested throughout the rest of the tests (if it wasn't a thunk, then all the tests would fail _"x is not a function"_). 

So we should think about the real requirements here, and how we can prove them. Perhaps worth looking into redux testing strategies - whether unit tests are good enough, whether integration tests are preferable, or both.

PR #254, opened 2017-11-03
Add properties header to right-hand side panel.

samiwel:
This PR adds a "Properties" header to the right hand side panel.
The word "properties" has been dropped from subsequent sections.
The side panel has been re-introduced into the application in preparation for some of the upcoming stories.
Snapshots have been updated.

### How to review 
A visual inspection of the app to see that the properties heading is displayed above all accordion sections.

### DIFF
--- a/src/components/Accordion/index.js
+++ b/src/components/Accordion/index.js
@@ -13,12 +13,11 @@ const StyledPanelTitle = styled(PanelTitle)`
   user-select: none;
   position: relative;
   padding: 0.65em 4px;
-  font-size: 0.6em;
+  font-size: 0.7em;
 
   & button {
     color: ${colors.text};
     font-size: 1em;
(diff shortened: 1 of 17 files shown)


samiwel:
[Properties header for right side panel (1)](https://trello.com/c/jkGKyCj0/282-properties-header-for-right-side-panel-1)

src/components/Accordion/index.js
```diff
@@ -13,12 +13,11 @@ const StyledPanelTitle = styled(PanelTitle)`
   user-select: none;
   position: relative;
   padding: 0.65em 4px;
-  font-size: 0.6em;
+  font-size: 0.7em;
```
hamishtaplin: This should be more like `0.8em` (roughly `13px`)

PR #256, opened 2017-11-06
Add toggle button component, story and test.

samiwel:
toggle-buttons1
This PR introduces a new `ToggleButton` presentational component which is needed for some of the upcoming UI changes to the create a questionnaire page and the properties panel.

### How to review 
See the `ToggleButton` component in the storybook.

### DIFF
--- a/src/components/ToggleSwitch/index.js
+++ b/src/components/ToggleSwitch/index.js
@@ -1,0 +1,169 @@
+import React from "react";
+import styled from "styled-components";
+import PropTypes from "prop-types";
+import { uniqueId } from "lodash";
+import { Input } from "components/Forms";
+
+const knobColors = {
+  on: "#008AD2"
+};
+
+const backgroundColors = {
+  on: "#78C3E9",
+  off: "#E8E8E8"
+};
+
+const SharedDefaultProps = {
+  width: 2.6,
+  height: 1
+};
+
+const glowScale = 1.5;
+
+export const HiddenInput = styled(Input)`
+  opacity: 0;
+  position: absolute;
+
+  &:focus {
+    border: none;
+  }
+`;
+
+export const ToggleSwitchBackground = styled.div`
+  cursor: pointer;
+  height: ${props => props.height}em;
+  width: ${props => props.width}em;
+  background: ${props =>
+    props.checked ? backgroundColors.on : backgroundColors.off};
+  border-radius: 1em;
+  content: "";
+  transition: background 100ms ease-in-out;
+  position: relative;
+`;
+
+ToggleSwitchBackground.propTypes = {
+  width: PropTypes.number,
+  height: PropTypes.number
+};
+
+ToggleSwitchBackground.defaultProps = SharedDefaultProps;
+
+const ToggleSwitchKnob = styled.div`
+  display: inline-block;
+  height: ${props => props.size}em;
+  width: ${props => props.size}em;
+  background: ${knobColors.on};
+  content: "";
+  position: relative;
+  top: ${props => -((props.size - props.height) / 2)}em;
+  left: ${props => (props.checked ? props.width - props.size : 0)}em;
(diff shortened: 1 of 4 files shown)


src/components/ToggleSwitch/index.js
```diff
@@ -0,0 +1,112 @@
...
+    label: PropTypes.string.isRequired,
+    name: PropTypes.string.isRequired
+  };
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      checked: props.checked || false
+    };
+    this.id = this.generateUniqueId();
```
WickyNilliams: I'd allow `id` to be passed as a prop, and only generate one if `id` is omitted. Otherwise tests can be non-deterministic. I had to take this approach with `Tooltip` component

src/components/ToggleSwitch/index.js
```diff
@@ -0,0 +1,112 @@
...
+  render() {
+    return (
+      <FlexDiv {...this.props}>
+        <label htmlFor={this.id}>{this.props.label}</label>
+        <Switch checked={this.state.checked} htmlFor={this.id}>
+          <HiddenInput
+            id={this.id}
+            type="checkbox"
+            onChange={this.handleToggle}
+            defaultChecked={this.state.checked}
```
WickyNilliams: and if you get rid of state, use `checked` instead of `defaultChecked`

samiwel:
[Create survey (removing the meta page) (8)](https://trello.com/c/YLRWMPzj/279-create-survey-removing-the-meta-page-8)

src/components/ToggleSwitch/index.js
```diff
@@ -0,0 +1,112 @@
...
+  position: absolute;
+  left: -99999px;
+`;
+
+const switchPositon = {
+  true: -ballSize,
+  false: nudgeLeft
+};
+
+const Switch = styled.label`
```
hamishtaplin: Not sure this should be a `label` as we already have a label elsewhere. Also there is no text inside to label the `input`.

src/components/ToggleSwitch/index.js
```diff
@@ -0,0 +1,112 @@
...
+    display: inline-block;
+    height: ${ballSize}em;
+    width: ${ballSize}em;
+    background: ${colors.on};
+    content: "";
+    position: relative;
+    top: ${nudgeTop}em;
+    left: ${props => switchPositon[props.checked]}em;
+    border-radius: 50%;
+    box-shadow: 1px 1px 2px #555555;
```
hamishtaplin: Can we use the dropshadow from the design which is:

```box-shadow: 0 0 1px 0 rgba(0,0,0,0.12), 0 1px 1px 0 rgba(0,0,0,0.24);```

File: src/components/ToggleSwitch/index.js [hunk 5]
```diff
@@ -0,0 +1,112 @@
...
+
+    transition: left 0.2s ease-in-out;
```
hamishtaplin: transition is a bit slow, I'd go with `100ms`

src/components/ToggleSwitch/index.js
```diff
@@ -0,0 +1,112 @@
...
+};
+
+const ballSize = 1.4;
+const railWidth = 2.6;
+const railHeight = 1;
+const nudgeTop = (ballSize - railHeight) / 2;
+const nudgeLeft = -railWidth;
+
+export const HiddenInput = styled.input`
+  visibility: hidden;
```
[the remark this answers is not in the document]
hamishtaplin: Take a look at how material design implement this without having to use a hidden `input`: https://material-components-web.appspot.com/switch.html

src/components/ToggleSwitch/index.js
samiwel: @hamishtaplin is there some other element I could use that, when clicked would toggle the checkbox?

PR #258, opened 2017-11-15
Piped values

WickyNilliams:
Allows piping of previous answers into current page.

### How to review 
Review code, run tests, pipe some values, test in preview


hamishtaplin:
Can we go with having `disabled` answers for now rather than omitting them?

Also, 'Page Title' for unentered titles.

src/components/Aux/index.js
```diff
@@ -0,0 +1 @@
+export default props => props.children;
```
samiwel: Where is this used other than the test? Not clear what the purpose of this component is.

src/components/QuestionPageEditor/MetaEditor.js
```diff
@@ -6,21 +6,24 @@ import CustomPropTypes from "custom-prop-types";
 import withEntityEditor from "components/withEntityEditor";
 import pageFragment from "graphql/fragments/page.graphql";
 import styled from "styled-components";
-
+import { compose, withApollo } from "react-apollo";
```
samiwel: Probably not worth making this change now. But going forward wondering if we should split out the stateless component and the apollo stuff and move the apollo compose logic into the `containers` package.

src/components/QuestionPageEditor/MetaEditor.js
```diff
@@ -31,9 +34,18 @@ const GuidanceRichTextEditor = styled(RichTextEditor)`
 
 export class StatelessMetaEditor extends React.Component {
   render() {
-    const { page, onChange, onUpdate } = this.props;
+    const { page, onChange, onUpdate, client } = this.props;
```
samiwel: Again, strictly speaking, the stateless component probably shouldn't know anything about how to use Apollo client. But I think this is something for another day.

src/components/Aux/index.js
WickyNilliams: Ah... I was using it, but after some refactoring I didn't need it any more. This component allows you to take advantage of React v16 handling arrays as return types - you can pass it multiple components and have it render them without a wrapper.

Is it worth removing, do you think? Or just leave for future use?

src/components/QuestionPageEditor/MetaEditor.js
WickyNilliams: There was a reason I had to do it this way, that I can't quite remember now. I'm thinking it was because i didn't want to run the query eagerly, as the `graphql` HoC does, but rather when the toolbar is opened

src/components/Aux/index.js
samiwel: Ah okay fair enough, maybe just leave it for now. Not doing any harm being there.

src/components/QuestionPageEditor/MetaEditor.js
samiwel: I guess if it helps make the implementation easier then fair enough. Wonder if eventually moving to a thunk-like approach similar to how toasts are raised would allow lazy loading of the query data in a less coupled way.

WickyNilliams: Yea, I can imagine eventually we'll move more and more apollo stuff into actions

PR #264, opened 2017-11-16
Fixing visual glitches with new delete/undo toast messages.

### LINKED ISSUE #259: Weird ghosting issue with Your Questionnaires title when scrolling down the page.
### Expected behaviour
No ghosting of text

### Actual behaviour
weird-ghosting-issue

Go to the questionnaires list and scroll down a bit.

### LINKED ISSUE #260: Unable to click delete questionnaire icon when toast is displayed.
### Expected behaviour
Should be able to click the delete icon.

### Actual behaviour
The toast takes up 100% of the width when displayed and prevents user from clicking on a layer behind.

Delete a questionnaire.
Observe toast is displayed.
Try to delete the questionnaire behind the toast.

### Technical information
The toast container spans 100% of the screen width. It should just occupy the width of the toast message.
As more toasts are displayed, it prevents user from clicking on more than one questionnaire.
It probably affects clicking on other items (such as answers) in the builder screen too.

### Screenshot
unable-to-delete-while-toast-displayed

### LINKED ISSUE #261: z-index of toasts is too low
### Expected behaviour

toasts are on a higher z-index than other content. 

### Actual behaviour

toasts are obscured by RTE

kapture 2017-11-16 at 9 50 28

samiwel:
This PR fixes a few visual glitches with the toast.
- Fixes #259 
- Fixes #260
- Fixes #261  - In addition to z-index had to remove the transparent background on toast messages to prevent text from bleeding through.

### How to review 
Re-test the above defects to confirm they are fixed.

### DIFF
--- a/src/components/Menu/__snapshots__/Menu.test.js.snap
+++ b/src/components/Menu/__snapshots__/Menu.test.js.snap
@@ -116,6 +116,11 @@ exports[`components/Menu should render a Menu 1`] = `
   width: 10px;
 }
 
+.c2 >:first-child {
+  -webkit-backface-visibility: hidden !important;
+  backface-visibility: hidden !important;
+}
+
 .c1 {
   overflow-y: auto;
   overflow-x: hidden;
--- a/src/components/ScrollPane/__snapshots__/index.test.js.snap
+++ b/src/components/ScrollPane/__snapshots__/index.test.js.snap
@@ -37,6 +37,11 @@ exports[`ScrollPane should render 1`] = `
   width: 10px;
 }
 
+.c0 >:first-child {
+  -webkit-backface-visibility: hidden !important;
+  backface-visibility: hidden !important;
+}
+
 <div
   className="c0"
 >
--- a/src/components/ScrollPane/index.js
+++ b/src/components/ScrollPane/index.js
@@ -33,6 +33,10 @@ const ScrollPane = styled.div`
   ::-webkit-scrollbar {
     width: 10px;
   }
+
+  > :first-child {
+    backface-visibility: hidden !important;
+  }
 `;
 
 export default ScrollPane;
--- a/src/containers/ToastContainer/index.js
+++ b/src/containers/ToastContainer/index.js
@@ -36,8 +36,9 @@ const ToastOuterContainer = styled.div`
   position: absolute;
   bottom: 0;
   text-align: center;
-  width: 100%;
+  min-width: 10em;
   margin-bottom: ${props => (props.hasMargin ? "0.5em" : "0")};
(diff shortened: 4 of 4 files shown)


samiwel:
[Visual glitches with toast component.](https://trello.com/c/c36fiTTs/304-visual-glitches-with-toast-component)

src/components/Toast/index.js
```diff
@@ -2,14 +2,11 @@ import React from "react";
...
 import { colors } from "constants/theme";
-import { transparentize } from "polished";
 import timer from "utils/timer";
 import { invoke } from "lodash";
 
-const bgColor = transparentize(0.1, colors.darkGrey);
-
 const StyledToast = styled.div`
-  background-color: ${bgColor};
+  background-color: ${colors.darkGrey};
```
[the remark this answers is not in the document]
samiwel: @WickyNilliams Text is visible through the toast if opacity is added back in. I can revert this if it's required for the design?

hamishtaplin:
@WickyNilliams @samiwel With regards the ghosting, using `backface-visibility: hidden` on the offending element also works.

We could apply that to the first-child of the `ScrollPane` component so it's always set.

src/components/App/index.js
```diff
@@ -37,6 +37,13 @@ injectGlobal`
   button {
     font-family: inherit;
   }
+  
+  #toast {
```
hamishtaplin: Can this CSS be in a component, rather than a global?

public/index.html
```diff
@@ -47,6 +47,5 @@
   </script>
   <body>
     <div id="root"></div>
-    <div id="toast"></div>
```
WickyNilliams: I assumed the plan was to keep all this stuff (portal etc) in place, and add `ToastWrapper` directly to `ToastContainer` component? e.g.

```js
<ToastWrapper>
  <ToastOuterContainer>
    <ToastInnerContainer>
      <ToastList>
        {map(toasts, (toast, id) => (
          // etc
        ))}
      </ToastList>
    </ToastInnerContainer>
  </ToastOuterContainer>
</ToastWrapper>;
```

samiwel: Isn't the benefit of using a portal that it's possible to render things somewhere else on the page? i.e. away from where the component is mounted? In this case, I don't see that there's any benefit of using portal since `ToastContainer` is rendered in-place in `BaseLayout`. So here I'm just wrapping it with `ToastWrapper` to apply the global styles in a component. Or am I misunderstanding something?

WickyNilliams:
BTW: you've reverted the changes to fix the ghosting issue.

PR #265, opened 2017-11-17
Display questionnaire properties in side panel and remove meta page.

samiwel:
This PR makes the front-end changes for moving the questionnaire properties into the properties panel.
The survey Id field has been removed.
The description field is now optional.
Checkboxes have been replaced with toggle switches.
Questionnaire properties can be changed from the sidebar.
The route and link to the meta page have been removed.
New summary property introduced to toggle on and off the summary. (Publisher changes to enable this to follow....)

Also did a bit of tidying up e.g. removed the unused "upload" button from questionnaire list.
Fixed up sizing of labels in answer properties.

### How to review 

### Dependencies
This PR depends on the following changes:
 - https://github.com/ONSdigital/eq-author-graphql-schema/pull/18
 - https://github.com/ONSdigital/eq-author-api/pull/46
 - https://github.com/ONSdigital/eq-publisher/pull/31


src/components/AnswerProperties/AnswerProperties.test.js
```diff
@@ -53,6 +51,10 @@ describe("Answer Properties", () => {
       wrapper = mount(answerProperties());
     });
 
+    it("should render when mounted", () => {
+      expect(wrapper).toMatchSnapshot();
```
WickyNilliams: Why mount?

src/components/PropertiesPanel/index.js
```diff
@@ -62,6 +64,14 @@ class PropertiesPanel extends React.Component {
         <PropertiesPaneBody>
           <ScrollPane>
             <Accordion>
+              <AccordionPanel
+                id={getIdForObject(this.props.page)}
```
WickyNilliams: `this.props.questionnaire`? Otherwise it has the same `id` as PagePropeties' Panel.

src/components/ToggleSwitch/index.js
```diff
@@ -132,20 +133,22 @@ class ToggleSwitch extends React.Component {
...
   render() {
-    const { checked, onChange, width, height, size } = this.props;
+    const { checked, onChange, width, height, size, onBlur } = this.props;
     return (
       <FlexInline width={width}>
         <HiddenInput
           id={this.id}
           type="checkbox"
           onChange={onChange}
+          onBlur={onBlur}
```
WickyNilliams: Why use onBlur here? Surely onChange is enough?

src/components/ToggleSwitch/index.js
```diff
@@ -132,20 +133,22 @@ class ToggleSwitch extends React.Component {
   };
 
   handleToggle = () => {
+    document.getElementById(this.id).focus();
```
WickyNilliams: You should use `ref`s for this. But I wonder why you need to re-focus anyway?

src/components/ToggleSwitch/index.js
```diff
@@ -132,20 +133,22 @@ class ToggleSwitch extends React.Component {
...
     this.props.onChange({
-      name: this.props.name,
+      name: this.id,
```
WickyNilliams: I wonder if we can use the `withChangeHandler` HoC for this?

src/containers/QuestionnairePropertiesContainer/index.js
```diff
@@ -0,0 +1,12 @@
+import { compose } from "react-apollo";
+
+import withUpdateQuestionnaire from "containers/enhancers/withUpdateQuestionnaire";
+import withEntityEditor from "components/withEntityEditor";
+import questionnaireFragment from "graphql/fragments/questionnaire.graphql";
+import StatelessQuestionnaireProperties from "components/QuestionnaireProperties";
+
+export default compose(withUpdateQuestionnaire)(
```
WickyNilliams:
```jscompose(
  withUpdateQuestionnaire,
  withEntityEditor("questionnaire", questionnaireFragment)
)(StatelessQuestionnaireProperties)

File: src/containers/QuestionnairesPage/QuestionnairesPage.js [hunk 7]
```diff
@@ -27,9 +30,6 @@ const Questionnaires = props => {
           >
             Create
           </LinkButton>
-          <Button id="btn-load-questionnaire" secondary disabled>
-            Upload
-          </Button>
```
WickyNilliams: Haha was wondering when we'd finally remove that :D

src/components/AnswerProperties/AnswerProperties.test.js
samiwel: Good spot, I added it in to test what was being rendered, but it's not required, so I'll remove.

src/components/ToggleSwitch/index.js
samiwel: The reason for adding this `focus()` is because the ToggleSwitch renders a hidden `checkbox` input. Since the click event is handled by the Toggle Switch (and not the checkbox), the focus is needed to ensure the `onBlur` event fires as expected.

samiwel: I will switch to using `ref` however.

samiwel:
[Create survey (removing the meta page) (8)](https://trello.com/c/YLRWMPzj/279-create-survey-removing-the-meta-page-8)

PR #266, opened 2017-11-22
Add custom 404 page design.

samiwel:
image
This PR adds the custom design to the 404 page not found page.

### How to review 
All existing checks and tests should pass.
Navigate to a route which does not exist e.g. http://localhost:3000/#/notfound should display the 404 page as per the design.

Note: This PR does not deal with Error cases coming back from queries, so if you put in a valid route URL, but the Id used is invalid/does not exist, then this will not display the 404 page.
There is another story to deal with error handling, and we might want to do this a bit differently.


samiwel:
[404 Page (2)](https://trello.com/c/5iQGGSyS/243-404-page-2)

src/containers/NotFoundPage/index.js
```diff
@@ -1,14 +1,40 @@
...
-      title="404—Page not found"
-      docTitle="Page not found"
-      hasNav={false}
-    />
+    <BaseLayout docTitle={pageTitle} hasNav={false}>
+      <Grid>
+        <Column cols={6} offset={3}>
+          <CenteredPane>
+            <Pencil />
+            <p>404 - Sorry, the page you were looking for was not found</p>
```
hamishtaplin: This should be `h1` as there's none present otherwise

PR #270, opened 2017-11-29
Piping fixes

WickyNilliams:
Fixes some issues with piping:

* disables Date Range piping, as this is more complex than first imagined
* handles piped values where the answer has since been deleted
* handles cases where an answer without a label has been piped 

### How to review 
review code, test fixes as outlined above


WickyNilliams:
[Piping (13)](https://trello.com/c/wVMdktbS/92-piping-13)

PR #273, opened 2017-12-01
Add Firebase authentication

WickyNilliams:
Adds auth with firebase. Treats most routes as private, requiring auth before they can be accessed.

**You need to add your own firebase credentials to run the app. The README was updated for this PR, and explains what is needed**

### How to review 
Have a gander at code, check it works, ensure auth "flow" feels right


src/components/Header/index.js
```diff
@@ -39,6 +43,30 @@ export const UtilityBtns = styled(ButtonGroup)`
   align-items: center;
 `;
 
+const UserAvatar = styled.img`
+  width: 2em;
```
hamishtaplin: Can we bump this up to `2.5em`? Was a bit bigger in the mockups and makes the tooltip look a bit tidier.

WickyNilliams:
[Integrate user sign in](https://trello.com/c/itrglzyg/296-integrate-user-sign-in)

src/components/UserProfile/index.js
```diff
@@ -0,0 +1,32 @@
...
+
+export const LogoutButton = styled(Button)`
+  padding-left: 0;
+  padding-right: 0;
+`;
+
+const UserProfile = ({ user, onSignOut }) => (
+  <Tooltip content="Sign Out">
+    <LogoutButton clear onClick={onSignOut}>
+      <UserAvatar src={user.photoURL} alt="" />
```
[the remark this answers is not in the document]
WickyNilliams: Left empty alt, supplemented with `role="presentation"`, discussed

README.md
```diff
@@ -130,6 +130,30 @@ REACT_APP_SECRET_CODE=abcdef
...
+REACT_APP_FIREBASE_PROJECT_ID=ABC REACT_APP_FIREBASE_API_KEY=DEF REACT_APP_FIREBASE_MESSAGING_SENDER_ID=GHI yarn start
+```
+
+Or they can be added to an `.env.local` file in the root of the repo:
+
+```
+REACT_APP_FIREBASE_PROJECT_ID="ABC"
+REACT_APP_FIREBASE_API_KEY="DEF"
+REACT_APP_FIREBASE_MESSAGING_SENDER_ID="GHI"
+```
```
samiwel: To test this, do I need to create my own firebase app, or is there already an eq-author app that we have the project_id and api_keys for?

samiwel: Got it working by creating own app 👍

src/redux/auth/reducer.js
```diff
@@ -0,0 +1,30 @@
...
+        verifiedStatus: true
+      };
+    default:
+      return state;
+  }
+};
+
+export const getUser = state => get(state, "auth.user");
+export const isSignedIn = state => Boolean(getUser(state));
+export const verifiedAuthStatus = state => get(state, "auth.verifiedStatus");
```
samiwel: Nice use of these utility methods 👍

samiwel:
Probably just need to think about how we manage the firebase app used in staging. i.e. is there some ONS account we should be creating the app under, can we add the developers as admins to that app, and we need to raise a PR request to eq-compose so that the firebase project id and api keys are passed in as env vars.

PR #276, opened 2017-12-14
324 deeper auth integration

WickyNilliams:
Adds deeper authentication integration. Specifically:

* display's name in header
* identifies users with full story (when built with `REACT_APP_USE_FULLSTORY="true"`)
* supply user's name to API on creation of questionnaire
* display "created by" in questionnaires list

### How to review 

Review code, run tests, play with app


WickyNilliams:
[Deeper authentication integration](https://trello.com/c/g11m6BHp/324-deeper-authentication-integration)

hamishtaplin:
Can we tidy this up?

screenshot 2018-01-18 12 00 07

hamishtaplin:
Icon/text aren't aligned centrally

screenshot 2018-01-18 12 05 22

samiwel:
Now that we're moving over to email authentication by default, wonder if it makes sense to default to disable auth in development mode for better DX?

WickyNilliams:
@samiwel i think it's better if we "dog food" what the users are doing. Otherwise we might miss issues with auth. You only need to sign in once, so not a lot of hassle for dev

PR #278, opened 2018-01-05
Move tabs into the main canvas and update snapshots.

samiwel:
Move the tabs into the main editing canvas. Previously the tabs occupied their own toolbar which was taking up precious screen real-estate.
This PR moves the tabs down into the main editing canvas. Routing tab does not exist at the moment.
Updated the relevant snapshots.

### How to review 
All checks and tests should pass.
Visually inspect that tabs are now appearing in the main editing canvas and matches the design on the Trello card.

### DIFF
--- a/src/components/BaseLayout/__snapshots__/BaseLayout.test.js.snap
+++ b/src/components/BaseLayout/__snapshots__/BaseLayout.test.js.snap
@@ -28,25 +28,6 @@ exports[`components/BaseLayout should render a Nav 1`] = `
           }
         }
       />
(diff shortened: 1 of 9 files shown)


samiwel:
[Move tabs to main editing canvas](https://trello.com/c/tVZGWKSG/253-move-tabs-to-main-editing-canvas)

hamishtaplin:
As discussed:

1. move tabs so they align with the `MainCanvas`
2. include (disabled for now) "routing" tab
3. adjust type/spacing to match design

PR #279, opened 2018-01-08
Separate panels on main editing canvas.

samiwel:
Visual change to separate the panels on the main editing canvas.
Also the add answer button has been re-named to "Add another answer", centered and now spans the entire width of the section as per the new design.

### How to review 
Ensure that the visual changes match the design attached to the trello card.
All existing checks and tests should continue to pass.

### DIFF
--- a/src/components/AnswerTypeSelector/index.js
+++ b/src/components/AnswerTypeSelector/index.js
@@ -6,6 +6,7 @@ import styled from "styled-components";
 import { colors } from "constants/theme";
 import AnswerTypeGrid from "./AnswerTypeGrid";
 import addIcon from "./icon-add.svg";
+import CustomPropTypes from "../../custom-prop-types";
 
 export const AddAnswerBtn = styled.button`
   appearance: none;
@@ -30,7 +31,8 @@ export const AddAnswerBtn = styled.button`
 
 export default class AnswerTypeSelector extends React.Component {
   static propTypes = {
-    onSelect: PropTypes.func.isRequired
+    onSelect: PropTypes.func.isRequired,
+    answers: PropTypes.arrayOf(CustomPropTypes.answer).isRequired
   };
 
   state = {
@@ -56,7 +58,7 @@ export default class AnswerTypeSelector extends React.Component {
   render() {
     const trigger = (
       <AddAnswerBtn type="button" id="add-answer-btn">
-        Add an answer
+        Add {this.props.answers.length === 0 ? "an" : "another"} answer
       </AddAnswerBtn>
     );
 
--- a/src/components/EditorSurface/CanvasSection.js
+++ b/src/components/EditorSurface/CanvasSection.js
@@ -17,6 +17,7 @@ export const BasicSection = styled.div`
   background-color: white;
   position: relative;
   box-shadow: ${shadow};
+  margin-bottom: 1em !important;
 `;
 
 const FocusableSection = styled(BasicSection)`
--- a/src/components/AnswerTypeSelector/AnswerTypeGrid.js
+++ b/src/components/AnswerTypeSelector/AnswerTypeGrid.js
@@ -21,6 +21,7 @@ const Menu = styled.div`
   background-color: white;
   box-shadow: rgba(0, 0, 0, 0.16) 0 5px 20px 0;
   width: 340px;
+  text-align: initial;
 `;
 
 const Header = styled.div`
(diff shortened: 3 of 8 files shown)


samiwel:
[Separate panels on main editing canvas](https://trello.com/c/VSfd7dx3/344-separate-panels-on-main-editing-canvas)

src/components/MainCanvas/index.js
```diff
@@ -5,6 +5,14 @@ const MainCanvas = styled.div`
...
   margin: 2em auto 1em;
   position: relative;
+
+  & .first-canvas-section {
+    margin-top: 2em;
+  }
+
+  & .last-canvas-section {
+    padding: 1em;
+  }
```
WickyNilliams: This style should be set on BasicCanvas, no need for manually-added class then.

src/components/MainCanvas/index.js
```diff
@@ -5,6 +5,14 @@ const MainCanvas = styled.div`
   max-width: 50em;
   margin: 2em auto 1em;
   position: relative;
+
+  & .first-canvas-section {
```
WickyNilliams: Seems that MainCanvas already had margin top. Adjacent margins collapse, so this is redundant

src/components/Popout/index.js
```diff
@@ -11,6 +11,7 @@ const ESC_KEY_CODE = 27;
 const Container = styled.div`
   position: relative;
   display: inline-block;
+  width: 100%;
```
WickyNilliams: Doesn't seem like the right place to add this style... not all Popups will be 100% width. You could put `text-align:center` on BasicCanvas (or add a prop `textAlign` prop to it, which defaults to `"left"`), and it will give the same effect. You can also then remove all the flex stuff added to `AnswerTypeSelector`.

src/components/AnswerTypeSelector/index.js
```diff
@@ -56,7 +62,7 @@ export default class AnswerTypeSelector extends React.Component {
   render() {
     const trigger = (
       <AddAnswerBtn type="button" id="add-answer-btn">
-        Add an answer
+        Add another answer
```
WickyNilliams: Should this have different wording depending on current number of answers? "Add _another_ answer" doesn't make sense on a brand new page

src/components/EditorSurface/CanvasSection.js
```diff
@@ -17,6 +17,7 @@ export const BasicSection = styled.div`
   background-color: white;
   position: relative;
   box-shadow: ${shadow};
+  margin-top: 1em;
```
WickyNilliams: @hamishtaplin do we tend to favour margin-bottom over top?

src/components/Popout/index.js
```diff
@@ -2,17 +2,26 @@
...
+
 const Container = styled.div`
   position: relative;
   display: inline-block;
+  ${props => props.stretched && stretched};
 `;
 
+Container.propTypes = {
+  stretched: PropTypes.bool
+};
```
WickyNilliams: I don't think we need this. 100% width of the container makes the position of the popout weird. 

Here's a git patch of changes i would suggest making (all deletions 🎉 ):

```diff
diff --git a/src/components/AnswerTypeSelector/index.js b/src/components/AnswerTypeSelector/index.js
index 0d8c53f..1643e37 100644
--- a/src/components/AnswerTypeSelector/index.js
+++ b/src/components/AnswerTypeSelector/index.js
@@ -16,7 +16,6 @@ export const AddAnswerBtn = styled.button`
   padding: 0;
   color: ${colors.text};
   font-size: 0.75em;
-  width: 100%;
 
   &::before {
     vertical-align: middle;
diff --git a/src/components/Popout/index.js b/src/components/Popout/index.js
index c09a91d..71ba0a6 100644
--- a/src/components/Popout/index.js
+++ b/src/components/Popout/index.js
@@ -2,26 +2,17 @@
 import React from "react";
 import { findDOMNode } from "react-dom";
 import PropTypes from "prop-types";
-import styled, { css } from "styled-components";
+import styled from "styled-components";
 import uncontrollable from "uncontrollable";
 import { TransitionGroup, CSSTransition } from "react-transition-group";
 
 const ESC_KEY_CODE = 27;
 
-const stretched = css`
-  width: 100%;
-`;
-
 const Container = styled.div`
   position: relative;
   display: inline-block;
-  ${props => props.stretched && stretched};
 `;
 
-Container.propTypes = {
-  stretched: PropTypes.bool
-};
-
 const Layer = styled.div`
   position: absolute;
   bottom: 0;
@@ -122,7 +113,7 @@ class Popout extends React.Component {
 
   render() {
     return (
-      <Container stretched>
+      <Container>
         {React.cloneElement(this.props.trigger, {
           onClick: this.handleToggleOpen,
           "aria-haspopup": true,
```

samiwel: Reason for stretching the width of this container was because of the A/C on the card that the button should span the entire width. I'll discuss with Kalun, but I agree about the popout location being weird, so I've applied the changes above for the time being 👍

PR #282, opened 2018-01-12
Add modal dialog component

samiwel:
Add a new modal component.
Add storybook stories to demonstrate functionality and implement some basic tests covering the acceptance criteria.

### How to review 
Review code and test that component displays correctly in Storybook.
Test that the component meets the acceptance criteria.


samiwel:
[Add modal dialog component (5)](https://trello.com/c/rHL9xkNx/345-add-modal-dialog-component-5)

src/components/Modal/dialog.js
```diff
@@ -0,0 +1,68 @@
+import React from "react";
+import styled, { css } from "styled-components";
+import PropTypes from "prop-types";
+import { colors } from "constants/theme";
+import Button from "components/Button";
+
+const sizes = {
```
hamishtaplin: I think it would be more flexible to allow the content to dictate the width (as it does height) rather than preset sizes.

src/components/Modal/story.js
```diff
@@ -0,0 +1,129 @@
...
+    this.setState({ showDialog: false });
+  };
+
+  render() {
+    return (
+      <Background>
+        <Button primary onClick={this.handleOpen}>
+          Trigger modal
+        </Button>
+        <Modal
```
hamishtaplin: Rather than pass props in and let the modal figure out what to do, it's more flexible to use composition. The modal itself should just be the window that pops up and has a close button. Not sure it should even be concerned with the other buttons, although this can be a it trickier. For example:

```javascript
<Modal>
  <ModalHeader>
    <ModalTitle>I am a title</ModalTitle>
    <p>I can be some arbitrary content</p>
  </ModalHeader>
  <ModalBody>
    <div>Anything can go in here.</div>
  </ModalBody>
  <ModalFooter>
    <BtnGroup horizontal>
      <button onClick={confirmHandler}>Confirm</button>
      <button onClick={cancelHandler}>Cancel</button>
    </BtnGroup>
  </ModalFooter>
</Modal>
```

Things like internal layout of the header can be done with further components, if needed. For example, we won't need an icon most of the time.

This is a good article on this approach: https://devbutze.com/p/93

src/components/Modal/index.js
```diff
@@ -0,0 +1,182 @@
...
+  };
+
+  componentWillMount() {
+    ReactModal.setAppElement("body");
+  }
+
+  primaryActionButton = primaryButton => {
+    if (primaryButton) {
+      // eslint-disable-next-line react/no-find-dom-node
+      findDOMNode(primaryButton).focus();
```
hamishtaplin: It's clunky, but the title should receive focus, not the button if we are shooting for this being as accessible as possible.

src/components/Modal/dialog.js
```diff
@@ -0,0 +1,68 @@
...
+  `,
+  large: css`
+    width: 900px;
+  `
+};
+
+const CloseButton = styled(Button)`
+  position: absolute;
+  top: 4px;
+  right: 4px;
```
WickyNilliams: em?

src/components/Modal/icon.js
```diff
@@ -0,0 +1,19 @@
+import PropTypes from "prop-types";
+import styled from "styled-components";
+
+const icons = {
+  move: require("./icons/move.min.svg"),
```
WickyNilliams: why use require instead of importing above?

src/components/Modal/index.js
WickyNilliams: Not sure this is true. w3c seem to recommend focus should be situation-dependent. For the type of actions we'll show this in response to, they say: https://www.w3.org/TR/wai-aria-practices/#dialog_modal

src/components/Modal/modal.test.js
```diff
@@ -0,0 +1,99 @@
+import React from "react";
+import { shallow, mount } from "enzyme";
+import Modal from "components/Modal";
+import Dialog from "components/Modal/dialog";
+
+const createWrapper = (props = {}, render = shallow, children = undefined) => {
```
WickyNilliams: not really necessary to change, but FYI: children is just another prop, you don't need to separate it out like that. These two are equivalent:

```js
const props = { children: <b>hello world</b> };

const header1 = <h1 {...props} />;
const header2 = <h1 children={props.children} />
const header3 = <h1>{props.children}</h1>
```

src/components/Modal/index.js
```diff
@@ -0,0 +1,182 @@
...
+  }
+};
+
+const collapsed = css`
+  padding: 0;
+  margin: 0;
+`;
+
+const flex = css`
+  display: flex;
```
WickyNilliams: Seems a little unnecessary to have a single property/value pair extracted?

src/custom-prop-types/index.js
```diff
@@ -74,5 +74,20 @@ export default {
...
+    size: PropTypes.oneOf(["small", "medium", "large"]),
+    isOpen: PropTypes.bool.isRequired,
+    onClose: PropTypes.func.isRequired,
+    primaryAction: PropTypes.func.isRequired,
+    primaryActionText: PropTypes.string.isRequired,
+    secondaryAction: PropTypes.func,
+    secondaryActionText: PropTypes.string,
+    tertiaryAction: PropTypes.func,
+    tertiaryActionText: PropTypes.string
+  }).isRequired
```
WickyNilliams: Is there any need to extract these? I can't imagine they'll be used anywhere outside of the modal

src/components/Modal/story.js
samiwel: Made some changes based on the above, but rather than introduce ModalHeader, ModalBody and ModalFooter etc. I've separated the concerns so that Modal just blurs out the background and renderers its children.

src/components/Modal/index.js
samiwel: @WickyNilliams To me, it seems sensible that focus should go on the least destructive action. But this is contrary to the acceptance criteria for this task which says that the primary action should receive focus. Perhaps the A/C needs to be re-visited?

package.json
```diff
@@ -173,14 +177,22 @@
...
   "babel": {
-    "presets": ["react-app"]
+    "presets": [
+      "react-app"
+    ]
   },
-  "moduleRoots": ["./src"]
+  "moduleRoots": [
+    "./src"
+  ]
```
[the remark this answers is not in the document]
samiwel: I've added a `.prettierignore` file which should exclude `package.json` from  being processed by prettier in the post-commit hook.

src/components/Modal/alert.js
```diff
@@ -0,0 +1,22 @@
+import styled from "styled-components";
```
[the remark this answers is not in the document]
samiwel: Changed the file names to be upper case 👍

src/components/Modal/index.js
samiwel: Have inlined these in the refactor. 👍

src/components/Dialog/index.js
```diff
@@ -0,0 +1,131 @@
...
+  static propTypes = {
+    children: PropTypes.node.isRequired,
+    onClose: PropTypes.func.isRequired,
+    primaryAction: PropTypes.func.isRequired,
+    primaryActionText: PropTypes.string.isRequired,
+    secondaryAction: PropTypes.func,
+    secondaryActionText: PropTypes.string,
+    tertiaryAction: PropTypes.func,
+    tertiaryActionText: PropTypes.string,
+    size: PropTypes.oneOf(["small", "medium", "large"])
```
WickyNilliams: `Object.keys(sizes)`? 

Though as @hamishtaplin said, I think it should be left to the content to decide dimensions (perhaps there should be some sensible min-height/width as default on here)

src/components/Dialog/index.js
```diff
@@ -0,0 +1,131 @@
...
+          {secondaryAction && (
+            <Action secondary onClick={secondaryAction}>
+              {secondaryActionText}
+            </Action>
+          )}
+          {tertiaryAction && (
+            <Action tertiary onClick={tertiaryAction}>
+              {tertiaryActionText}
+            </Action>
+          )}
```
WickyNilliams: A third button seems excessive? Can't imagine any scenarios where we'd want it?

src/components/Modal/index.js
WickyNilliams: I made an assumption that primary was the best to focus when we were writing the A/C. So yeah, I think we should change to match recommendations

src/components/Modal/story.js
```diff
@@ -0,0 +1,164 @@
...
+    icon: "move",
+    primaryAction: action("Primary action"),
+    primaryActionText: "Primary action",
+    secondaryAction: action("Secondary action"),
+    secondaryActionText: "Secondary action"
+  };
+
+  constructor(props) {
+    super(props);
+    this.state = {
```
WickyNilliams: if you're not using the props in the constructor, you can omit the constructor and have state as a class property. e.g.

```js
class Foo extends React.Component {
  state = {
     foo: "bar"
  };
}
```

src/components/Dialog/index.js
samiwel: I agree, can't see us ever having a need for a "Yes", "No", "Maybe"?

src/components/Dialog/Dialog.test.js
```diff
@@ -0,0 +1,111 @@
...
+        size: "medium"
+      })
+    ).toMatchSnapshot();
+  });
+
+  it("should render a large sized dialog", () => {
+    expect(
+      createWrapper({
+        ...props,
+        size: "large"
```
WickyNilliams: Don't need these size props anymore

src/components/Modal/index.js
```diff
@@ -0,0 +1,53 @@
...
+  render() {
+    const { children, isOpen, onClose } = this.props;
+    return (
+      <StyledModal
+        isOpen={isOpen}
+        onRequestClose={onClose}
+        shouldCloseOnOverlayClick
+        style={reactModalStyleOverride}
+      >
+        <Centered onClick={onClose}>{children}</Centered>
```
WickyNilliams: Why does this close on click?

src/components/Modal/story.js
```diff
@@ -0,0 +1,148 @@
...
+  .add("Modal - Should grow with content", () => (
+    <StatefulModal
+      heading="Large Dialog Example"
+      subheading=""
+      description="The modal should change size based on its content."
+      icon="delete"
+      primaryActionText="Save"
+      tertiaryAction={action("Tertiary action")}
+      tertiaryActionText="Delete"
+      secondaryAction={undefined}
```
WickyNilliams: The secondary button still shows, despite it being undefined here. I guess that's only a problem with the wrapper used in this story though?

samiwel: Did a little tweaking to the story, should be fixed 👍

src/components/Modal/story.js
hamishtaplin: I agree with separating the Modal from the Dialog but the Dialog itself should still be composable to allow this component to be flexible.

hamishtaplin:
There's something weird going on with the stories as the Modal doesn't dismiss when you navigate away from that story.

WickyNilliams:
Latest changes look very good!

src/components/Modal/index.js
```diff
@@ -0,0 +1,82 @@
...
+};
+
+// eslint-disable-next-line no-unused-expressions
+injectGlobal`
+.ReactModalPortal > div {
+    opacity: 0;
+}
+
+.ReactModalPortal .ReactModal__Overlay {
+  transition: opacity 200ms ease-out;
```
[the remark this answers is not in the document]
samiwel: Agreed this will be left for now due to use of `injectGlobal`

PR #283, opened 2018-01-19
Introduce 'Add question page' button to main canvas

samiwel:
Introduce new button at the bottom of the main canvas that allows user to add a question place.
Upon clicking the button, a new page is created and the user is navigated to the new page.
image

### How to review 
Visual inspection of the new button.
Review the code.


src/components/IconButton/index.js
```diff
@@ -0,0 +1,26 @@
...
+    color: ${colors.highlight};
+  }
+
+  &:hover svg path,
+  &:focus svg path {
+    fill: ${colors.highlight};
+  }
+`;
+
+export default IconButton;
```
WickyNilliams: We already have an IconButton component?

src/components/IconButtons/AddQuestionPageButton.js
```diff
@@ -0,0 +1,12 @@
...
+import { AddPageIcon } from "./icons";
+
+export default props => {
+  return (
+    <IconButton clear {...props}>
+      <AddPageIcon />
+      Add question page
+    </IconButton>
+  );
+};
```
WickyNilliams: Why is this in a folder IconButton_s_? Why not put it as a sibling of QuestionnaireDesignPage, since it'll only ever be used there

samiwel:
[Quick access to add a question page (2)](https://trello.com/c/61nXgVNR/333-quick-access-to-add-a-question-page-2)

src/containers/QuestionnaireDesignPage/QuestionnaireDesignPage.js
```diff
@@ -89,6 +99,13 @@ export class QuestionnaireDesignPage extends Component {
                   onUpdateSection={onUpdateSection}
                 />
               </MainCanvas>
+              <Centered>
+                <AddQuestionPageButton
+                  onClick={function() {
```
WickyNilliams: extract method `handleAddPageClick`?

src/components/IconButton/index.js
samiwel: Yeah, unfortunately it doesn't do what we need here. Probably need to re-think the naming of the other IconButton / remove it in favour of this IconButton going forward.

src/components/IconButtons/AddQuestionPageButton.js
samiwel: The idea behind this folder is that it holds all of the concrete implementations of IconButtons that are used in Author going forward. So eventually the "Add an answer" button should also live here.

src/components/IconButton/index.js
hamishtaplin: imo if we are introducing a new component, then any existing components should be refactored to use that new one.

src/components/IconButtons/AddQuestionPageButton.js
hamishtaplin: This isn't the right way to go, imo. There should be a canonical `IconButton` component which can be implemented, rather than a bunch of implementations in a folder—we'll likely never use the same button twice so as Nick says, they can live with the component that implements them. The `IconButton` component should be written to accept anything it needs (icon, text. variants of colour/size, etc.) as props. This is how the existing `IconButton` works, we just need to change the implementation.

hamishtaplin:
Distance between icon and button is inconsistent with the two buttons:

screenshot 2018-01-19 15 52 56

Also, the colours are pure black instead of the shade used elsewhere.

Spacing is also no consistent with designs:

screenshot 2018-01-19 16 00 34

This would be helped by upping the font size to `14px` while we're at it.

samiwel:
I have updated most of the other icon buttons to use the new IconButton so that they are consistent across the app. However,I have not been able to remove the old IconButton completely since it is used a lot in the Draft JS popup toolbar and changing it over to use the new IconButton yields some weird display issues. Fixing these is beyond the scope of this task so I think it's better to look at that separately in a tech debt task.

src/components/AnswerTypeSelector/index.js
```diff
@@ -57,9 +35,16 @@ export default class AnswerTypeSelector extends React.Component {
...
     const trigger = (
-      <AddAnswerBtn type="button" id="add-answer-btn">
+      <IconButton
+        id="add-answer-button"
+        title={`Add ${this.props.answers.length === 0
+          ? "an"
+          : "another"} answer`}
+        icon={addIcon}
+        clear
+      >
```
WickyNilliams: You use `title` and `children` for the same thing here. `children` seems like a better choice?

src/components/IconButton/IconButton.test.js
```diff
@@ -0,0 +1,42 @@
...
+    };
+    wrapper = createWrapper(props);
+  });
+
+  it("should render correctly", () => {
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it("should render icon only", () => {
+    wrapper = createWrapper({ ...props, iconOnly: true }, mount);
```
WickyNilliams: Why mount?

src/components/IconButton/index.js
```diff
@@ -0,0 +1,69 @@
...
+  };
+
+  static defaultProps = {
+    highlightOnHover: true,
+    iconOnly: false
+  };
+
+  render() {
+    const { icon, title, iconOnly, ...otherProps } = this.props;
+    const Text = iconOnly ? <VisuallyHidden>{title}</VisuallyHidden> : title;
```
WickyNilliams: No need to capitalise the identifier

src/components/IconButton/index.js
```diff
@@ -0,0 +1,69 @@
...
+    highlightOnHover: PropTypes.bool
+  };
+
+  static defaultProps = {
+    highlightOnHover: true,
+    iconOnly: false
+  };
+
+  render() {
+    const { icon, title, iconOnly, ...otherProps } = this.props;
```
WickyNilliams: As mentioned above, children is the more natural choice than title

src/components/IconButton/index.js
```diff
@@ -0,0 +1,69 @@
...
+    const Text = iconOnly ? <VisuallyHidden>{title}</VisuallyHidden> : title;
+    return (
+      <Tooltip content={title}>
+        <div>
+          <StyledButton {...otherProps}>
+            <SVG src={icon} uniqueHash={otherProps.uniqueHash} />
```
WickyNilliams: What is uniqueHash for?

samiwel: It's an implementation detail of the `react-inlinesvg` package. I didn't want to use it here, but needed a way of mocking it for a test. It generates a random SHA hash during render otherwise. I was going to just spread `otherProps` but thought it was better to be explicit about what props were being used.

WickyNilliams:
Seems a bit much to have a tooltip as well as the regular text? 

image

src/components/IconButton/IconButton.test.js
samiwel: Codecov doesn't register coverage of props used within styled component CSS unless the component is rendered using mount.

samiwel:
Yeah, I agree about the tooltip being a bit overkill when you have the button text right underneath. I'll restrict it so that tooltip only renders if `iconOnly` is set to true. I think that makes a bit more sense since the tooltip is useful in order to explain what the icon does.

src/components/AnswerTypeSelector/index.js
samiwel: I've removed the `title` prop.

src/components/IconButton/IconButton.test.js
hamishtaplin: Yeah I had that problem elsewhere

hamishtaplin:
Tooltips shouldn't be a concern of the button, imo. They're so simple to implement that keeping them out of the button reduces complexity and sits better with our "prefer composition" approach.

hamishtaplin:
Sorry to be pedantic but icon/text not aligned.

screenshot 2018-01-24 09 33 22

PR #284, opened 2018-01-23
230 focus styles

WickyNilliams:
* Adds focus styles to all inputs on editor canvas
* Updates "primary" blue colours under Kalun's direction
* Removes `withSeamlessness` HoC, since it was only used in one component. The other component which used it - `SeamlessInput` - wasn't used itself, so that was deleted too.
* Rename SeamlessTextArea to WrappingInput (since it behaves like a regular input more than a textarea)

### How to review 
Play around with focus on editor canvas, ensure it works as expected


WickyNilliams:
[Focus styles (3)](https://trello.com/c/Vo0BEaHO/230-focus-styles-3)

src/components/WrappingInput/index.js
```diff
@@ -0,0 +1,90 @@
...
+import styled, { css } from "styled-components";
+import AutoResizeTextArea from "react-textarea-autosize";
+import withChangeHandler from "components/Forms/withChangeHandler";
+import { colors } from "constants/theme";
+import { transparentize } from "polished";
+
+const ENTER_KEY = 13;
+
+/**
+ * 1.75em  = 28px/16px
```
hamishtaplin: Are we better offloading this to either a) some helper functions or (preferably) b) forgetting about px values of font sizes and just approximating them with some rounder numbers?

Really, we should have a more formal type scale with value we re-use.

WickyNilliams: worth noting: i didn't add/change these values in this PR, i simply renamed the file they were in. 

The values as shown originally came from Dan, so i noted them for future reference, possibly at your request

samiwel:
Not necessarily something that needs to be fixed under this PR, but the new focus styles seem to be overlapping the delete button on Answers, looks a bit weird.
image

src/components/WrappingInput/index.js
samiwel: Probably worth deferring this change until Kalun has finished looking at the type scale.

PR #285, opened 2018-01-24
Implement new sidebar navigation design.

samiwel:
The navigation bar has been redesigned for accessibility purposes.
This PR is a first stab at updating the styling to match the new design.

### How to review 
Check to make sure that it looks okay and behaves correctly.


samiwel:
[Implement new sidebar design (5)](https://trello.com/c/ffsjO1cV/356-implement-new-sidebar-design-5)

src/containers/QuestionnaireDesignPage/QuestionnaireDesignPage.js
```diff
@@ -21,6 +21,13 @@ const Centered = styled.div`
   margin-bottom: 4em;
 `;
 
+const StyledScrollPane = styled(ScrollPane)`
```
hamishtaplin: Is there a more descriptive name we can give this? How does it differ from a regular `ScrollPane`?

src/constants/theme.js
```diff
@@ -10,7 +10,12 @@ export const colors = {
   orange: "#F5A623",
   borders: "#EAEAEA",
   white: "#FFFFFF",
-  highlight: "#008ad2"
+  highlight: "#008ad2",
+  textInverted: "#E1E1E1",
+  textInvertedHover: "#FFFFFF",
+  navBackground: "#4A4A4A",
```
hamishtaplin: Is there any need to keep these colours here? They're not used in more than one context (nav) and so aren't re-usable.

samiwel: I get what you mean, but I think it's better to keep theme colours together so they can be changed more easily without  being scattered across the app. I'm not sure if there are other places where these colours will be used once Kalun has finished his re-design?

src/containers/QuestionnaireDesignPage/QuestionnaireDesignPage.js
samiwel: Agreed, is `NavigationScrollPane` better? This might need to change actually. Kalun has asked us to investigate if the scroll bar can be placed inside, yet another pane to give the illusion of the inner scroll bar.

src/components/QuestionnaireNav/PageNav.js
```diff
@@ -151,15 +145,12 @@ export class PageNavItem extends React.Component {
     });
   };
 
+  handleFocus = _ => {
+    document.querySelector('[class*="NavigationScrollPane"]').scrollLeft = 0;
```
WickyNilliams: Hmmm... this doesn't seem like the right way to do this.

src/components/QuestionnaireNav/PageNav.test.js
```diff
@@ -83,4 +83,42 @@ describe("PageNav", () => {
...
+      jest.restoreAllMocks();
+    });
+
+    it("should set the scrollLeft to 0", () => {
+      querySelector.mockReturnValue(mockScrollPane);
+      wrapper.instance().handleFocus();
+      expect(querySelector).toHaveBeenCalledWith(
+        '[class*="NavigationScrollPane"]'
+      );
+      expect(mockScrollPane.scrollLeft).toEqual(0);
```
WickyNilliams: This is not a good test. A better test would be `simulate("focus")` on the delete button and then check `scrollLeft === 0`. No need to worry about *how* that is done (querySelector), only that it is done

src/components/QuestionnaireNav/SectionNavItem.js
```diff
@@ -109,6 +119,10 @@ class SectionNavItem extends React.Component {
     saveSectionItemRef(section.id, elem);
   };
 
+  handleDeleteFocus = _ => {
+    document.querySelector('[class*="NavigationScrollPane"]').scrollLeft = 0;
```
WickyNilliams: Again, this raises eyebrows. We should reconsider this approach

src/components/QuestionnaireNav/SectionNavItem.test.js
```diff
@@ -71,4 +71,27 @@ describe("SectionNavItem", () => {
...
+    });
+
+    afterEach(() => {
+      jest.restoreAllMocks();
+    });
+
+    it("should set the scrollLeft to 0", () => {
+      querySelector.mockReturnValue(mockScrollPane);
+      wrapper.instance().handleDeleteFocus();
+      expect(querySelector).toHaveBeenCalledWith(
```
WickyNilliams: As above

src/components/QuestionnaireNav/index.js
```diff
@@ -8,40 +8,79 @@ import plusIcon from "./icon-plus.svg";
...
+    margin-left: 0.5em;
+  }
+`;
+
+const SettingsLink = styled.a`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+
+  & > div {
```
WickyNilliams: Instead of descendent selector, create a new styled-component for this.

src/components/QuestionnaireNav/index.js
```diff
@@ -77,7 +116,29 @@ class QuestionnaireNav extends Component {
...
+          </li>
+          <li>
+            <SettingsLink href="/">
+              <div>Settings</div>
+              <svg width="16" height="16" viewBox="0 0 24 24">
+                <path
+                  fill="#fff"
+                  d="M12,15.5A3.5,3.5 0 0,1 8.5,12A3.5,3.5 0 0,1 12,8.5A3.5,3.5 0 0,1 15.5,12A3.5,3.5 0 0,1 12,15.5M19.43,12.97C19.47,12.65 19.5,12.33 19.5,12C19.5,11.67 19.47,11.34 19.43,11L21.54,9.37C21.73,9.22 21.78,8.95 21.66,8.73L19.66,5.27C19.54,5.05 19.27,4.96 19.05,5.05L16.56,6.05C16.04,5.66 15.5,5.32 14.87,5.07L14.5,2.42C14.46,2.18 14.25,2 14,2H10C9.75,2 9.54,2.18 9.5,2.42L9.13,5.07C8.5,5.32 7.96,5.66 7.44,6.05L4.95,5.05C4.73,4.96 4.46,5.05 4.34,5.27L2.34,8.73C2.21,8.95 2.27,9.22 2.46,9.37L4.57,11C4.53,11.34 4.5,11.67 4.5,12C4.5,12.33 4.53,12.65 4.57,12.97L2.46,14.63C2.27,14.78 2.21,15.05 2.34,15.27L4.34,18.73C4.46,18.95 4.73,19.03 4.95,18.95L7.44,17.94C7.96,18.34 8.5,18.68 9.13,18.93L9.5,21.58C9.54,21.82 9.75,22 10,22H14C14.25,22 14.46,21.82 14.5,21.58L14.87,18.93C15.5,18.67 16.04,18.34 16.56,17.94L19.05,18.95C19.27,19.03 19.54,18.95 19.66,18.73L21.66,15.27C21.78,15.05 21.73,14.78 21.54,14.63L19.43,12.97Z"
+                />
+              </svg>
```
WickyNilliams: Why are we embedding the SVG directly here? We don't use this approach elsewhere

src/constants/theme.js
WickyNilliams: i'm not convinced we should put these in theme either. Make them usefully-named constants where they're used, sure, but `textInvertedHover: "#FFFFFF"` isn't really of use here.

WickyNilliams:
I wonder if we should rename all the `QuestionnaireNav` to something else. We've all been referring to it in various forms as the "sidebar", i wonder whether a name incorporating that would be better suited

samiwel:
`NavigationSidebar`?

src/components/QuestionnaireNav/index.js
samiwel: This is just place holder. There's a separate card to do the home button.

src/components/QuestionnaireNav/PageNav.js
samiwel: Do you have ideas for a different approach to solving this?

src/components/QuestionnaireNav/SectionNavItem.js
samiwel: This is the only way I could get it to work. Bare in mind, that this is a temporary stop-gap solution. Eventually the delete button is being removed from the side bar.

src/components/QuestionnaireNav/index.js
WickyNilliams: But seems odd to inline it, instead of import the file?

WickyNilliams: but you may as well extract out of here?

src/components/QuestionnaireNav/index.js
hamishtaplin: Yeah we need to avoid selectors like that as much as possible, they are very fragile.

src/components/QuestionnaireNav/index.js
```diff
@@ -8,40 +8,79 @@ import plusIcon from "./icon-plus.svg";
...
+  padding: 1em;
+  margin: 0;
+  background: ${colors.navBackground};
+  position: sticky;
+  top: 0;
+  left: 0;
+  z-index: 9999;
+  border-bottom: 1px solid #c3c3c3;
+
+  & > li {
```
hamishtaplin: Create a separate component for this.

src/components/QuestionnaireNav/index.js
```diff
@@ -8,40 +8,79 @@ import plusIcon from "./icon-plus.svg";
...
+    text-decoration: none;
+    display: flex;
+    align-items: center;
+  }
+
+  & a:hover {
+    color: ${colors.textInvertedHover};
+  }
+
+  & li:nth-child(2) a svg {
```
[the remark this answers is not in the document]
hamishtaplin: This can probably be solved with a more component-based approach, even if it's a bigger refactor.

src/components/QuestionnaireNav/index.js
samiwel: Have moved the SVG's out into their own files.

src/constants/theme.js
samiwel: I've inlined these colours in the component that use them. I've also removed `textInvertedHover` in favour of re-using ${colors.white}

src/components/QuestionnaireNav/PageNav.js
samiwel: I will leave this in for the time being. It's a quick fix until Hamish's PR get's merged, then it can be removed.

src/components/QuestionnaireNav/index.js
samiwel: This selector has been removed altogether.

src/components/QuestionnaireNav/index.js
```diff
@@ -8,40 +8,79 @@ import plusIcon from "./icon-plus.svg";
...
+  }
+
+  & a,
+  & a:link,
+  & a:visited {
+    color: ${colors.textInverted};
+    text-decoration: none;
+    display: flex;
+    align-items: center;
+  }
```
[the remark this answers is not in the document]
samiwel: I've split these out now.

src/components/QuestionnaireNav/SectionNavItem.js
WickyNilliams: Hmm... so we have to keep this until we can delete sections elsewhere?

WickyNilliams:
Sections still a bit too tall, compared to the design:

image

In the design it looks as though page and section items are same height

src/components/NavigationSidebar/NavigationHeader.js
```diff
@@ -0,0 +1,75 @@
...
+  z-index: 9999;
+  border-bottom: 1px solid #c3c3c3;
+
+  & > li {
+    display: flex;
+    align-items: center;
+  }
+`;
+
+const StyledLink = styled.a`
```
WickyNilliams: This should extend react-router's NavLink component, rather than be a regular anchor. As it stands, this causes a full page reload, which breaks the user experience

src/components/NavigationSidebar/PageNav.js
```diff
@@ -198,7 +153,6 @@ const PageNav = ({ section, questionnaire, onDelete }) => (
             pageId={page.id}
             sectionId={section.id}
             questionnaireId={questionnaire.id}
-            onDelete={onDelete}
             index={pages.length - i}
```
WickyNilliams: I believe we no longer need the `index` prop. Same goes for `pageNumber` above it

src/components/NavigationSidebar/NavigationHeader.js
samiwel: Good shout

src/components/NavigationSidebar/NavigationHeader.js
```diff
@@ -0,0 +1,74 @@
...
+    vertical-align: middle;
+  }
+`;
+
+const NavigationHeader = () => {
+  return (
+    <IconList>
+      <li>
+        <StyledLink to="/">
+          <SVG uniqueHash="home-icon" src={homeIcon} />
```
WickyNilliams: This needs some text

src/components/NavigationSidebar/index.js
```diff
@@ -3,53 +3,68 @@ import styled from "styled-components";
...
+  }
+
+  &:hover {
+    &::-webkit-scrollbar-thumb {
+      background: ${colors.lightGrey};
+    }
+  }
+
+  ::-webkit-scrollbar-track {
+    border-radius: 0.5em;
```
WickyNilliams: The has meant that the scrollbars have an inconsistent look. Either the rounded edge should be dropped from here, or added to every other scroll bar.

samiwel: As per the design it should be rounded but agree that it should be consistent. I've reverted the rounded edges for now, I guess it's for Kalun to decide if we need to re-visit scroll bar appearance across the app.

src/components/NavigationSidebar/NavigationHeader.js
samiwel: Added some text, but made it visually hidden as per the design.

PR #286, opened 2018-01-24
341 Delete page

hamishtaplin:

[Trello card](https://trello.com/c/E2iEO5GW/341-delete-page-5)

- adds toolbar to editor
- clicking delete button opens modal confirmation

### How to review 

- navigate to or add a question
- click delete
- confirm modal opens
- confirm "delete" button deletes that question and navigates as expected
- confirm action is cancelable
- confirm navigation via back/prev buttons closes modal


src/components/QuestionPageEditor/DeleteIcon.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+
+const icon = () => (
```
samiwel: You can use `react-inlinesvg` here rather than have a separate component for the DeleteIcon.

hamishtaplin:
[Delete page (5)](https://trello.com/c/E2iEO5GW/341-delete-page-5)

hamishtaplin:
[Delete confirmation](https://trello.com/c/N8socS2A/337-delete-confirmation)

WickyNilliams:
Are we keeping the toast pop up?

src/components/QuestionPageEditor/DeleteConfirmDialog.js
```diff
@@ -0,0 +1,76 @@
...
+
+import getTextFromHTML from "utils/getTextFromHTML";
+
+import iconAlert from "./icon-alert.svg";
+
+const Dialog = styled(ModalDialog)`
+  width: 30em;
+  position: relative;
+  top: -7em;
+`;
```
WickyNilliams: Why is this dialog special in terms of positioning? Shouldn't every dialog be the same?

src/components/QuestionPageEditor/DeleteConfirmDialog.js
```diff
@@ -0,0 +1,76 @@
...
+
+  &::before {
+    content: url(${iconAlert});
+    display: inline-block;
+    margin: 0 0.5em 0 0;
+  }
+`;
+
+const DeleteConfirmModal = ({ page, onDeletePage, onClose, ...otherProps }) => (
+  <Dialog onClose={onClose} {...otherProps}>
```
WickyNilliams: It's confusing calling this just Dialog as we already have a component called that, and this is actually a ModalDialog, not just a Dialog

src/components/QuestionPageEditor/DeleteConfirmDialog.js
```diff
@@ -0,0 +1,76 @@
...
+        <Description>
+          <Alert>
+            All edits, properties and routing settings will also be removed.
+          </Alert>
+        </Description>
+      </Message>
+      <DialogIcon icon="page" />
+    </DialogHeader>
+    <ButtonGroup horizontal align="right">
+      <Button secondary onClick={onClose} id="btn-cancel">
```
WickyNilliams: Since these `id`s are used for tests, I wonder if we should come up with a consistent approach, a data attribute like `data-test="whatever"` or something? I think there are babel/webpack plugins available to strip that kind of stuff in production builds

src/components/QuestionPageEditor/QuestionPageEditor.test.js
```diff
@@ -74,4 +82,13 @@ describe("Question Page Editor", () => {
     wrapper.find(AnswerTypeSelector).simulate("select", "Textfield");
     expect(mockMutations.onAddAnswer).toHaveBeenCalledWith("Textfield");
   });
+
+  it("should behave...", () => {
```
WickyNilliams: how should it behave? haha

src/components/QuestionPageEditor/QuestionPageToolbar.js
```diff
@@ -0,0 +1,83 @@
...
+
+import { colors } from "constants/theme";
+
+import Tooltip from "components/Tooltip";
+import DeleteConfirmDialog from "./DeleteConfirmDialog";
+
+import PropTypes from "prop-types";
+import CustomPropTypes from "custom-prop-types";
+
+import SVG from "react-inlinesvg";
```
WickyNilliams: I somehow missed this being added as a dep. Why not use a webpack loader for this kind of stuff? Saves having to make a http request at run time to get the image.

src/components/QuestionPageEditor/QuestionPageToolbar.js
```diff
@@ -0,0 +1,83 @@
...
+
+export const DeleteButton = styled.button.attrs({ type: "button" })`
+  background: none;
+  border: none;
+  padding: 0.25rem;
+  margin: 0;
+
+  .lid {
+    transform-origin: 50% 50%;
+    transition: all 200ms ease-out;
```
WickyNilliams: For some reason, the lid gets all blurry/pixelly on hover on my external monitors (this whole image looks particularly blurry on my mac's screen, but it's extra pronounced on my external monitor)

image

src/components/QuestionPageEditor/DeleteConfirmDialog.js
```diff
@@ -0,0 +1,76 @@
...
+  margin-bottom: 1em;
+
+  &::before {
+    content: url(${iconAlert});
+    display: inline-block;
+    margin: 0 0.5em 0 0;
+  }
+`;
+
+const DeleteConfirmModal = ({ page, onDeletePage, onClose, ...otherProps }) => (
```
WickyNilliams: The file is called `DeleteConfirmDialog.js`, but the component is `DeleteConfirmModal`?

src/components/QuestionPageEditor/DeleteConfirmDialog.js
```diff
@@ -0,0 +1,76 @@
...
+      </Button>
+      <Button tertiary autoFocus onClick={onDeletePage} id="btn-delete">
+        Delete
+      </Button>
+    </ButtonGroup>
+  </Dialog>
+);
+
+DeleteConfirmModal.propTypes = {
+  page: CustomPropTypes.page,
```
WickyNilliams: This should probably just take a title or something rather than whole page object? Since we'll likely use the same Dialog for deleting Sections too?

src/components/QuestionPageEditor/DeleteConfirmDialog.js
```diff
@@ -0,0 +1,76 @@
...
+          </Alert>
+        </Description>
+      </Message>
+      <DialogIcon icon="page" />
+    </DialogHeader>
+    <ButtonGroup horizontal align="right">
+      <Button secondary onClick={onClose} id="btn-cancel">
+        Cancel
+      </Button>
+      <Button tertiary autoFocus onClick={onDeletePage} id="btn-delete">
```
WickyNilliams: Seems odd that tertiary styles are used. Is "tertiary" a misnomer?

src/components/QuestionPageEditor/DeleteConfirmDialog.js
```diff
@@ -0,0 +1,76 @@
...
+        </Subheading>
+        <Description>
+          <Alert>
+            All edits, properties and routing settings will also be removed.
+          </Alert>
+        </Description>
+      </Message>
+      <DialogIcon icon="page" />
+    </DialogHeader>
+    <ButtonGroup horizontal align="right">
```
WickyNilliams: I notice you're not using `DialogButtons` component - is this because they're not fit for purpose? If so, should we modify them so that they are, instead of ditching them?

src/components/QuestionPageEditor/QuestionPageToolbar.test.js
```diff
@@ -0,0 +1,50 @@
...
+  });
+
+  it("should render", () => {
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it("should display Modal when delete button is clicked", () => {
+    const deleteBtn = wrapper.find(DeleteButton);
+    deleteBtn.simulate("click");
+    expect(wrapper.state().showDeleteConfirmDialog).toBe(true);
```
WickyNilliams: You're better off finding the Modal component and testing the prop `isOpen` is set to `true` than the state. A component's props are public, state is private. When i switch to use my `ModalTrigger` for instance, these tests will break. Whereas testing props should not.

src/components/QuestionPageEditor/QuestionPageToolbar.test.js
```diff
@@ -0,0 +1,50 @@
...
+  });
+
+  it("should offload closing of Modal to the Dialog itself", () => {
+    const deleteConfirmDialog = wrapper.find(DeleteConfirmDialog);
+    deleteConfirmDialog.simulate("close");
+    expect(wrapper.state().showDeleteConfirmDialog).toBe(false);
```
WickyNilliams: As above

src/containers/QuestionnaireDesignPage/index.js
```diff
@@ -1,19 +1,23 @@
 import { compose } from "react-apollo";
 import { connect } from "react-redux";
 
+import * as ToastActionCreators from "redux/toast/actions";
```
WickyNilliams: What are these used for?

src/components/Dialog/DialogIcon/index.js
```diff
@@ -3,10 +3,12 @@ import styled from "styled-components";
 
 import moveIcon from "./icons/icon-dialog-move.svg";
 import deleteIcon from "./icons/icon-dialog-delete.svg";
+import pageIcon from "./icons/icon-dialog-page.svg";
 
 const icons = {
   move: moveIcon,
-  delete: deleteIcon
+  delete: deleteIcon,
+  page: pageIcon
```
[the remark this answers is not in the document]
hamishtaplin: yeah definitely

src/components/QuestionPageEditor/DeleteConfirmDialog.js
hamishtaplin: Yes, I'll change the main component

src/components/QuestionPageEditor/DeleteConfirmDialog.js
hamishtaplin: Will fix

src/components/QuestionPageEditor/DeleteConfirmDialog.js
hamishtaplin: DeleteConfirmModalDialog?

src/components/QuestionPageEditor/DeleteConfirmDialog.js
hamishtaplin: No, they don't work for this due to it having a different primary action button (called tertiary but we can deal with this when Kalun looks at buttons).

This is why I recommended we stick to composition as much as possible instead of trying to guess how all modals will work in future. We already have the abstractions required to render some buttons in a group, with some variants of layout. I don't think we need the `DialogButton` component if we're just going to have to change it all the time, what's it for? It's pretty trivial to implement buttons as it is.

src/components/QuestionPageEditor/DeleteConfirmDialog.js
hamishtaplin: Yeah that would be great tbh, I did wonder that when I was adding this. Data attributes are much better as you don't have to worry about duplicates.

src/components/QuestionPageEditor/DeleteConfirmDialog.js
hamishtaplin: Yeah it's odd. Defer until Kalun looks at buttons, imo.

src/components/QuestionPageEditor/DeleteConfirmDialog.js
hamishtaplin: Agreed.

src/components/QuestionPageEditor/QuestionPageEditor.test.js
hamishtaplin: It should just _behave_

src/components/QuestionPageEditor/QuestionPageToolbar.js
hamishtaplin: I didn't realise it did that. Out of scope for this possibly? Can be a tech debt thing.

src/components/QuestionPageEditor/QuestionPageToolbar.js
```diff
@@ -0,0 +1,83 @@
...
+import iconDelete from "./icon-delete.svg";
+
+export const Toolbar = styled.div`
+  display: flex;
+  justify-content: flex-end;
+  border-bottom: 1px solid ${colors.borders};
+  margin: 0 -1em 1em;
+`;
+
+export const DeleteButton = styled.button.attrs({ type: "button" })`
```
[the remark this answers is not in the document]
hamishtaplin: I can't remember, will look into it.

src/components/QuestionPageEditor/QuestionPageToolbar.js
hamishtaplin: I'll take another look.

src/components/QuestionPageEditor/QuestionPageToolbar.js
```diff
@@ -0,0 +1,83 @@
...
+    .lid {
+      transform: translateY(-1px) rotate(6deg);
+    }
+
+    .lid,
+    .body {
+      fill: ${colors.blue};
+    }
+  }
+`;
```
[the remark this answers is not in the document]
hamishtaplin: Yeah, I noticed that too.

src/components/QuestionPageEditor/DeleteConfirmDialog.js
WickyNilliams: I mean the styled component on the inside i.e: `const Dialog = styled(ModalDialog)`. I think the "confirm" part is implied by it being a Dialog. Not too fussed on naming otherwise

src/components/QuestionPageEditor/DeleteConfirmDialog.js
WickyNilliams: After discussion in slack, I'm happy to ditch DialogButton if it's not flexible enough for our uses (and it's easy enough to build your own button groups, as here)

src/components/QuestionPageEditor/QuestionPageToolbar.js
WickyNilliams: Created a card for this: https://trello.com/c/dRuA9hJV/152-improve-usage-of-svg-in-components

src/containers/QuestionnaireDesignPage/index.js
WickyNilliams: Cool, cheers. I thought they weren't being used, but couldn't tell if they were passed down the tree somewhere.

hamishtaplin: Oh actually yeah they are used so that the toast pops up after deletion.

WickyNilliams: As discussed, will look breaking this implicit dependency, and have withDeletePage/withDeleteSection import the toast action creators directly.

WickyNilliams: For future readers: plan failed. Just added a comment to describe what's going on

PR #287, opened 2018-01-29
Integrate Cypress as e2e continuous testing tool.

samiwel:
Following a spike, a decision was taken by the team to migrate from using chimp for end to end testing to using cypress.

This change removes previous chimp configuration and tests and replaces it with cypress dependencies and tests.

### How to review 
All tests should pass on travis and when run locally.


samiwel:
[Integrate cypress as e2e tool (5)](https://trello.com/c/xpCwBujo/353-integrate-cypress-as-e2e-tool-5)

cypress/integration/author_spec.js
```diff
@@ -0,0 +1,212 @@
...
+  it("Can create checkboxes", () => {
+    cy.get("#add-answer-button").click();
+    cy.contains("Checkbox").click();
+    cy.get("#MultipleChoiceAnswer1").within(() => {
+      cy
+        .get("button")
+        .contains("Add another option")
+        .click();
+    });
+    cy.get("#Option1").within(() => {
```
samiwel: Is there any way of re-working this to not have an explicit option Id embedded in the test?If the test was run more than once, without deleting the mockDataStore, it will probably fail? Can we avoid having a manual step to delete the mockDataStore.

cypress/integration/author_spec.js
```diff
@@ -0,0 +1,212 @@
...
+      cy.get("textarea#label").type("Checkbox label");
+      cy.get("#description").type("Checkbox description");
+    });
+  });
+
+  it("Can delete checkboxes", () => {
+    cy.get("#Option2").within(() => {
```
samiwel: Same as above.

cypress/integration/author_spec.js
```diff
@@ -0,0 +1,212 @@
...
+    });
+    cy.get("[aria-label='Delete answer']").click();
+    cy.get('[id *="MultipleChoiceAnswer"]').should("not.exist");
+  });
+
+  it("Can create radio buttons", () => {
+    cy.get("#add-answer-button").click();
+    cy.contains("Radio").click();
+    cy.get("#MultipleChoiceAnswer2").within(() => {
+      cy.get("#Option3, #Option4");
```
samiwel: Same here.

cypress/integration/author_spec.js
```diff
@@ -0,0 +1,212 @@
...
+      5: "Currency",
+      6: "Number",
+      7: "Date"
+    };
+    for (var i = 3; i <= 7; i++) {
+      cy.get("#add-answer-button").click();
+      cy.contains(QuestionTypes[i]).click();
+
+      cy.get("#label").type(QuestionTypes[i] + " label");
+      if (i < 7) {
```
samiwel: Rather than have conditional logic in here which does something different for Date, does it make more sense to split testing date into it's own test case?

cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
samiwel: We can delete this file. It's not used.

cypress/integration/author_spec.js
```diff
@@ -0,0 +1,212 @@
...
+
+    cy.get('[class *="MultipleChoiceAnswer__Options"]').within(() => {
+      cy.get('[class *="Option__StyledOption"]').should("have.length", 2);
+    });
+    cy.get("[aria-label='Delete answer']").click();
+    cy.get('[id *="MultipleChoiceAnswer"]').should("not.exist");
+  });
+
+  it("Can create and delete the different types of textbox", () => {
+    var QuestionTypes = {
```
samiwel: Should be called `AnswerTypes`?

samiwel: Potentially this should be an Array, rather than an Object.

cypress/integration/author_spec.js
```diff
@@ -0,0 +1,223 @@
...
+      cy
+        .get("[aria-label='Delete section']")
+        .first()
+        .click();
+
+      cy.get("[aria-label='Delete section']").should("have.length", 1);
+    });
+  });
+
+  it("Can create checkboxes", () => {
```
samiwel: This test doesn't seem to be asserting anything. We should always have some kind of an assertion in our tests. Maybe assert that the number of options `should('equal', 2)`?

cypress/integration/author_spec.js
```diff
@@ -0,0 +1,223 @@
...
+      });
+
+    cy.get('[class *="MultipleChoiceAnswer__Options"]').within(() => {
+      cy.get('[class *="Option__StyledOption"]').should("have.length", 1);
+    });
+    cy.get("[aria-label='Delete answer']").click();
+    cy.get('[id *="MultipleChoiceAnswer"]').should("not.exist");
+  });
+
+  it("Can create radio buttons", () => {
```
samiwel: Same as above, missing an assertion for this test.

cypress/integration/author_spec.js
```diff
@@ -0,0 +1,223 @@
...
+
+  it("Can create and delete date ranges", () => {
+    cy.get("#add-answer-button").click();
+    cy.contains("Date range").click();
+    cy.get('[id *="BasicAnswer"]');
+    cy.get("[aria-label='Delete answer']").click();
+    cy.get('[id *="BasicAnswer"]').should("not.exist");
+  });
+
+  it("can edit section title", () => {
```
samiwel: Missing assertion. Perhaps assert that the new section title is displayed on the Navigation bar?

cypress/integration/author_spec.js
```diff
@@ -0,0 +1,223 @@
...
+        .type("s")
+        .click();
+
+      cy.get("[aria-label='description']").click();
+    });
+
+    cy.get("#questionnaire-nav").contains("s");
+  });
+
+  it("can edit page title", () => {
```
samiwel: Same as above.

cypress/integration/author_spec.js
```diff
@@ -0,0 +1,223 @@
...
+    });
+  });
+
+  it("should create a new section when deleting only section", () => {
+    cy.hash().then(prevHash => {
+      cy.get("#questionnaire-nav [aria-label='Delete section']").click();
+      cy.hash().should("not.equal", prevHash);
+    });
+  });
+  // it('can undo a delete action', () => {
```
samiwel: We shouldn't leave commented out code in here. Better to delete it.

cypress/integration/author_spec.js
```diff
@@ -0,0 +1,223 @@
...
+      cy.get('[class *="Option__StyledOption"]').should("have.length", 2);
+    });
+    cy.get("[aria-label='Delete answer']").click();
+    cy.get('[id *="MultipleChoiceAnswer"]').should("not.exist");
+  });
+
+  it("Can create and delete the different types of textbox", () => {
+    const answerTypes = ["Text", "Textarea", "Currency", "Number"];
+
+    for (var answer of answerTypes) {
```
samiwel: Prefer using `let` instead of `var`. https://stackoverflow.com/questions/762011/whats-the-difference-between-using-let-and-var-to-declare-a-variable

WickyNilliams:
@samiwel not directly related to the changes here, but we should probably come up with a consistent way of identifying sections and elements on the page. Hamish started using `data-test` attributes the other day in a couple of his unit tests. Perhaps we should extend this to the rest of the app, to make these tests simpler/clearer/robust

cypress/integration/author_spec.js
```diff
@@ -0,0 +1,239 @@
...
+      cy.get('[class *="Option__StyledOption"]').should("have.length", 2);
+    });
+    cy.get("[aria-label='Delete answer']").click();
+    cy.get('[id *="MultipleChoiceAnswer"]').should("not.exist");
+  });
+
+  it("Can create and delete the different types of textbox", () => {
+    const answerTypes = ["Text", "Textarea", "Currency", "Number"];
+
+    for (let answer of answerTypes) {
```
WickyNilliams: We tend to prefer using functional approaches to iteration in JS. So forEach is a better choice here:

samiwel:
Yeah, agree about standardising our approach for identifying elements for tests. Annoying that we have to resort to using Id and className matching. Better approach to use a common `data` attributes sometimes use className Worth having that discussion as team at some point.

cypress/integration/author_spec.js
```diff
@@ -0,0 +1,227 @@
...
+    cy.get("#add-answer-button").click();
+    cy.contains("Date").click();
+    cy.get('[id *="BasicAnswer"]');
+    cy.get("[aria-label='Delete answer']").click();
+    cy.get('[id *="BasicAnswer"]').should("not.exist");
+  });
+
+  it("Can create and delete date ranges", () => {
+    cy.get("#add-answer-button").click();
+    cy.contains("Date range").click();
```
WickyNilliams: It might be worth extracting a helper function `addAnswer(answerType)`, since you have this kind of code in multiple places:

```js
cy.get("#add-answer-button").click();
cy.contains(answerType).click();
```

cypress/integration/author_spec.js
```diff
@@ -0,0 +1,227 @@
...
+  });
+
+  it("can create a new section", () => {
+    cy.get("#questionnaire-nav").within(() => {
+      cy
+        .get("button")
+        .contains("Create new section")
+        .click();
+
+      cy.hash().should("match", /questionnaire\/\d+\/design\/\d+\/\d+/);
```
WickyNilliams: The requirement with creating a page/section is that you are taken to that new page/section after creation. This test only asserts that you're still on _some_ page in the designer, but not a _different_ page. As a result, this test would pass even if the app was doing the wrong thing - which isn't good.

Therefore, you should assert here that you've moved to a new section/page, by comparing the hash to the previous hash (as is done in the test `"should create a new page when deleting only page in section"`). If you want a challenge you can take this quite far by checking that only pageId changes when creating a new page, and sectionId and pageId change when creating a new section.

.travis.yml
```diff
@@ -18,15 +18,13 @@ install:
  - yarn upgrade eq-author-graphql-schema
 
 before_script:
- - yarn global add chimp
- - yarn global add serve
+ - yarn global add cypress serve
```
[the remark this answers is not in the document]
WickyNilliams: @samiwel what do you think?

.travis.yml
```diff
@@ -39,6 +37,7 @@ after_success:
 env:
   global:
     - NODE_PATH=src/
+    - CYPRESS_baseUrl=http://localhost:3000
```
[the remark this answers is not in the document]
WickyNilliams: @samiwel what do you think

package.json
```diff
@@ -10,7 +10,7 @@
     "build:mockAPI": "REACT_APP_USE_MOCK_API=true node scripts/build.js",
     "build:status": "node scripts/build-status-page.js",
     "test": "node scripts/test.js",
-    "smoketest:author": "chimp config/chimp.author.js",
+    "smoketest:author": "cypress run -s cypress/integration/author_spec.js --browser chrome",
     "smoketest:storybook": "chimp config/chimp.storybook.js",
```
WickyNilliams: Should remove all references to chimp

WickyNilliams:
Commenting here as it's not directly related to any specific change: I think the README should be updated to explain how to run smoketests locally

.travis.yml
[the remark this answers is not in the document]
samiwel: We could possibly remove since `serve` is listed in the `devDependencies`. Interestingly, we also have a `serve` entry in the scripts section of `package.json`.

cypress/integration/author_spec.js
samiwel: Agree, I would just add to NIck's comment that simply checking for inequality, whilst it would pass is perhaps not the right approach.

.travis.yml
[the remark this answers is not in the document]
samiwel: Definitely makes sense to add it to `.env.development` since it will mean you don't have to manually specify it when running locally. I'm not sure which .env file would be applied on Travis, my guess would be `.env.production` since the Cypress test suite runs against a served production optimised build of Author. We probably *don't* want to add it to .env.production so maybe leaving it as a Travis env variable is the best approach?

cypress/integration/author_spec.js
SamGodwin2: @WickyNilliams Just looking back at that test it's asserted in the next line with `cy.get("[aria-label='Delete section']").should("have.length", 2);` as if the new section hasn't been created this will still equal 1. The hash match only checks to make sure its on the right page for that assertion to happen.

.travis.yml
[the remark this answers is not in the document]
WickyNilliams: Sounds good!

.travis.yml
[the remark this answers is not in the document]
WickyNilliams: Yeah would be good to add to `.env.development` for easing local testing. I think Travis would be using development build, since we just run:

```js
REACT_APP_ENABLE_AUTH=false yarn build:mockAPI
```

This should probably be changed to set `NODE_ENV=production`.

I previously added this line to `run_smoketests_author.sh` to use test vars - so we could add `CYPRESS_baseUrl` there and it would work. 

```sh
# Read test env vars
export $(egrep -v '^#' .env.test | xargs)
```

cypress/integration/author_spec.js
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  it("Should redirect to the sign-in page", () => {
+    cy.visit("/");
+    cy.url().should("include", "sign-in");
+  });
+
+  it("Should allow user to log in as guest", () => {
+    cy.contains("Sign in as Guest").click();
+    cy.get("h1").should("contain", "Your Questionnaires");
```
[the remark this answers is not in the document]
SamGodwin2: @WickyNilliams Looked into trying to change this however Travis and local machines have different root urls and we couldn't figure a way around it.

.travis.yml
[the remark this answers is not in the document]
samiwel: Ah yeah, I see that line now. Interesting though since the `build:mockAPI` task sets `NODE_ENV` to `production` because it runs the `scripts/build.js` script.

cypress/integration/author_spec.js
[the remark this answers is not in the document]
WickyNilliams: I think you can get at the config values like this:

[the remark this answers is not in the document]
WickyNilliams: I think you can get at the config values like this:

```js
Cypress.env("baseUrl")
```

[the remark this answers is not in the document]
samiwel: @WickyNilliams I think the tricky thing here was deciding where to set `REACT_APP_BASE_NAME`. When running locally (i.e. in development mode) base name is not set. When running on Travis base name is set to "/eq-author" as per the production config. So that's all fine and well for the React app, but in order to check the `hash()`to assert you're back at the root, you need to take `REACT_APP_BASE_NAME` into account. Since cypress runs in its own node process it doesn't have access to this env var so it needs to be set somewhere, but deciding where to do that was tricky. i.e. it shouldn't be in run_author_smoke_test.sh or the package.json script since they might be run locally. I guess it could go into the .travis config, but it just doesn't seem right.

cypress/integration/author_spec.js
```diff
@@ -0,0 +1,327 @@
...
+      .then(() => {
+        expect(regexCapturesBefore.questionnaireId).to.equal(
+          regexCapturesAfter.questionnaireId
+        );
+        expect(regexCapturesBefore.sectionId).to.equal(
+          regexCapturesAfter.sectionId
+        );
+        expect(regexCapturesBefore.pageId).to.not.equal(
+          regexCapturesAfter.pageId
+        );
```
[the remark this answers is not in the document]
WickyNilliams: Using the above approach allows you to simplify this test to:

```js
  it("can create a new page", () => {
    let prevHash;

    cy.hash().then(hash => {
      prevHash = hash;
    });

    cy.get("#btn-add-page").click();
    cy.get('[class *="PageNav__Link-"]').should("have.length", 2);

    cy.hash().then(hash =>
      assertHash(prevHash, hash, {
        questionnaireId: true,
        sectionId: true,
        pageId: false
      })
    );
  });
```

which is a lot clearer than before :)

cypress/utility_functions/cy_test_functions.js
```diff
@@ -0,0 +1,50 @@
...
+    .contains("Create")
+    .click();
+}
+
+function addAnswerType(answerType) {
+  cy.get("#add-answer-button").click();
+  cy.contains(answerType).click();
+}
+
+function hashApplication(prevHash, currentHash, equality) {
```
WickyNilliams: Can you give this a more descriptive name? The name doesn't tell me what the function does. `assertHash` or something similar would be preferable

cypress/utility_functions/cy_test_functions.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  if (equality.pageId) {
+    expect(prevRegexCapture[3], "pageId").to.equal(currentRegexCapture[3]);
+  } else {
+    expect(prevRegexCapture[3], "pageId").not.to.equal(currentRegexCapture[3]);
+  }
+}
+
+export { createQuestionnaire, addAnswerType, hashApplication };
```
WickyNilliams: Can you export the individual functions, instead of an object. e.g. 

```js
export function addAnswerType() {
```

cypress/integration/author_spec.js
```diff
@@ -1,35 +1,10 @@
...
-      questionnaireId: regexCapture[1],
-      sectionId: regexCapture[2],
-      pageId: regexCapture[3]
-    };
-  }
+import {
+  createQuestionnaire,
+  addAnswerType,
+  hashApplication
+} from "../utility_functions/cy_test_functions.js";
```
WickyNilliams: Can you rename this file and folder? Perhaps name the folder `utils` (as we have elsewhere), and the file `index.js` (so it's easy to import then with `import { whatever } from "../utils"`)

PR #289, opened 2018-02-01
Tweak codecov config - 90% target, 0.25% threshold

WickyNilliams:
We have a high level code coverage, and codecov is currently configured to expect a constant upwards trend. It is more reasonable and practical to have a lower bound, combined with a threshold, so small dips are permissible but we keep coverage high.

### How to review 
Ensure configuration is reasonable and codecov passes

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -2,8 +2,8 @@ coverage:
   status:
     project:
       default:
-        target: auto
-        threshold: null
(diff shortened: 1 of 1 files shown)


samiwel:
[Revisit codecov configuration](https://trello.com/c/ux6FODs0/367-revisit-codecov-configuration)

PR #291, opened 2018-02-05
Fixed the focus style issue on the add question button.

SamGodwin2:
To change the style of the nav bar to bring the focus style of the add question button into line with the rest of the nav bar focus styles.

### How to review 
Should pass all tests and the focus style of the add question button should look the same as the page link focus styles. 
Also this blue line should no longer exist.
image

### DIFF
--- a/src/components/NavigationSidebar/SectionNavItem.js
+++ b/src/components/NavigationSidebar/SectionNavItem.js
@@ -22,6 +22,8 @@ export const AddPageBtn = styled.button`
   color: ${textInverted};
   font-weight: bold;
   font-size: 0.75em;
+  margin-bottom: 0.35em;
+  width: 100%;
 
   &::before {
     font-size: 1.25em;


src/components/NavigationSidebar/SectionNavItem.js
```diff
@@ -22,6 +22,8 @@ export const AddPageBtn = styled.button`
   color: ${textInverted};
   font-weight: bold;
   font-size: 0.75em;
+  margin-bottom: 5px;
```
samiwel: 0.25em?

src/components/NavigationSidebar/SectionNavItem.js
```diff
@@ -22,6 +22,8 @@ export const AddPageBtn = styled.button`
...
+  padding-right: 218.89px;
```
samiwel: delete this and make `width: 100%`

PR #292, opened 2018-02-06
375 eq author 288 non existent survey link doesnt re direct to a 404 error issue

SamGodwin2:
### What is the context of this PR?
Fixes #288 by redirecting to a 404 error if the url tries to point towards a non-existent questionnaire, section or page.  

### How to review 
Program should pass all tests and if a url is entered in the format http://localhost:3000/#/questionnaire/(random_characters)/design/58/94 you should be redirected to a 404 error rather than the program crashing

PR #294, opened 2018-02-08
Revert "375 eq author 288 non existent survey link doesnt re direct to a 404 error issue"

samiwel:
Reverts ONSdigital/eq-author#292


samiwel:
Reverting this PR since it causes conflicting issues with the Cypress tests. We will re-apply this fix after merging the cypress tests.

PR #297, opened 2018-02-14
Remove duplicate and redundant DOM element Ids

SamGodwin2:
Removed duplicate/redundant id's and test dependancies on on id references.

### How to review 
No tests should depend on ID references and redundant Id's should all be removed while all tests should still pass on jest and cypress.


SamGodwin2:
[Fix duplicate IDs (3)](https://trello.com/c/Rdczwraa/373-fix-duplicate-ids-3)

src/components/QuestionPageEditor/MetaEditor.js
```diff
@@ -58,9 +58,10 @@ export class StatelessMetaEditor extends React.Component {
             controls={titleControls}
             size="large"
             fetchAnswers={fetchAnswers}
+            datatest="txt-question-title"
```
hamishtaplin: I would either camelCase this (`dataTest`) or separate with hyphens (`data-test`).

cypress/utils/index.js
```diff
@@ -1,7 +1,7 @@
 export function setQuestionnaireSettings(name) {
   cy.get(`[data-test="questionnaire-settings-form"]`).within(() => {
     cy
-      .get("#title")
+      .get("[data-test='txt-Questionnaire-title']")
```
hamishtaplin: Why is this capitalised?

package-lock.json
```diff
@@ -0,0 +1,539 @@
...
+      "resolved": "https://registry.npmjs.org/zen-observable/-/zen-observable-0.7.1.tgz",
+      "integrity": "sha512-OI6VMSe0yeqaouIXtedC+F55Sr6r9ppS7+wTbSexkYdHbdt4ctTuPNXP/rwm7GTVI63YBc+EBT0b0tl7YnJLRg=="
+    },
+    "zen-observable-ts": {
+      "version": "0.4.4",
+      "resolved": "https://registry.npmjs.org/zen-observable-ts/-/zen-observable-ts-0.4.4.tgz",
+      "integrity": "sha512-SNVY1sWWhoe7FwFmHpD9ERi+7Mhhj3+JdS0BGy2UxLIg7cY+3zQbyZauQCI6DN6YK4uoKNaIm3S7Qkqi1Lr+Fw=="
+    }
+  }
+}
```
WickyNilliams: Why has this file been added? This is an artefact of using `npm` but we should be using `yarn`

src/components/RichTextEditor/index.js
```diff
@@ -132,7 +132,8 @@ class RichTextEditor extends React.Component {
     className: PropTypes.string,
     multiline: PropTypes.bool,
     size: PropTypes.oneOf(Object.keys(sizes)),
-    fetchAnswers: PropTypes.func
+    fetchAnswers: PropTypes.func,
+    dataTest: PropTypes.string
```
WickyNilliams: I don't think `dataTest` is a very good name for this. `testId`, or something more descriptive, would be better

hamishtaplin:
What happened to 1 and 2? 🤔

src/components/Answers/MultipleChoiceAnswer/__snapshots__/Option.test.js.snap
```diff
@@ -285,21 +281,22 @@ exports[`Option should render a checkbox 1`] = `
               className="c1 c2"
             >
               <Option__DummyInput
-                id="label"
+                id={undefined}
```
samiwel: Why the id undefined here? We should either set the Id or remove the prop if it's not required.

src/components/Answers/MultipleChoiceAnswer/__snapshots__/Option.test.js.snap
```diff
@@ -391,16 +391,20 @@ exports[`Option should render a checkbox 1`] = `
             className="c2"
           >
             <withChangeHandler(WrappingTextArea)
-              id="description"
+              data-test="option-description"
+              id={undefined}
```
samiwel: Same as above.

samiwel:
Looks a lot better, there's a few places where id is undefined in the snapshots. We should prob remove the Id props if they're not being set.

src/components/Answers/MultipleChoiceAnswer/__snapshots__/Option.test.js.snap
SamGodwin2: Looks like the issue is due to the mount function as it renders all prop types, possibly soved by updating to a newer version of jest-enzyme? https://github.com/FormidableLabs/enzyme-matchers/issues/184

WickyNilliams: We don't use `id` in `Option` anymore, so it can be removed from the component's propTypes and the test, which should clear up this

WickyNilliams: Though tbh, not sure why DummyInput is rendering `id` at all, it's not passed to it

samiwel: Seems like there's a fix for enzyme-to-json package that may fix this, worth upgrading perhaps? https://github.com/adriantoine/enzyme-to-json/releases/tag/v3.2.1

SamGodwin2: @WickyNilliams Yeah tried removing it from the proptypes and the test, it still keeps appearing as an undefined prop in DummyInput though.

PR #298, opened 2018-02-15
Upgrade to Apollo client v2

samiwel:
This PR removes dependency on the old v1 client and replaces with new dependencies required for Apollo v2.
Steps have been taken to migrate existing setup of redux store and Apollo client to the new way of doing things in v2.

### How to review 
Review code with reference to the [Apollo v2 migration guide](https://github.com/apollographql/apollo-client/blob/master/docs/source/2.0-migration.md).
Spin up the app both using the mock ApI and using the real API to check that the application behaves normally and that I haven't missed any key bits of functionality.
All existing automated tests and checks should pass. Cypress tests should run successfully.


samiwel:
[Upgrade to Apollo v2 (1 day)](https://trello.com/c/Tv5rRzC6/372-upgrade-to-apollo-v2-1-day)

package.json
```diff
@@ -108,6 +108,12 @@
     "whatwg-fetch": "^2.0.3"
   },
   "dependencies": {
+    "apollo-cache-inmemory": "latest",
```
WickyNilliams: don't use latest

package.json
```diff
@@ -108,6 +108,12 @@
...
+    "apollo-client": "latest",
+    "apollo-client-preset": "^1.0.8",
+    "apollo-link": "latest",
+    "apollo-link-http": "latest",
```
WickyNilliams: same with all of these

src/apollo/createSchemaLink.js
```diff
@@ -0,0 +1,15 @@
+import typeDefs from "eq-author-graphql-schema";
+import { makeExecutableSchema, addMockFunctionsToSchema } from "graphql-tools";
+import { SchemaLink } from "apollo-link-schema";
+
+export const createSchema = (resolvers, fn = addMockFunctionsToSchema) => {
```
WickyNilliams: Can you called this `createExecutableSchema`?

Also, I don't think this should be exported, but rather called by the default export. I  can't imagine a scenario where you'd call one and not the other, so may as well have just one function to do it all.

src/apollo/createSchemaLink.test.js
```diff
@@ -0,0 +1,29 @@
...
+    it("should pass resolvers to addMockFunctionsToSchema", () => {
+      const resolvers = { String: noop };
+      const addMockFunctionsToSchema = jest.fn();
+
+      createSchema(resolvers, addMockFunctionsToSchema);
+
+      expect(addMockFunctionsToSchema).toHaveBeenCalledWith(
+        expect.objectContaining({
+          mocks: resolvers
+        })
```
WickyNilliams: I wonder if it's better to write an integration test for this? You could dreate a full `ApolloClient` with the `SchemaLink`, and run a simple `query()` against it. That way you know everything works together.

src/index.js
```diff
@@ -1,28 +1,37 @@
...
 import configureStore from "redux/configureStore";
-import createClient from "apollo/createClient";
 import App from "containers/App";
 import Raven from "raven-js";
+import getIdForObject from "utils/getIdForObject";
+import fragmentMatcher from "./apollo/fragmentMatcher";
+import createApolloClient from "apollo/createApolloClient";
+import createApolloCache from "apollo/createApolloCache";
+import createSchemaLink, { createSchema } from "./apollo/createSchemaLink";
+import createHttpLink from "./apollo/createHttpLink";
```
WickyNilliams: Can you change the imports to be consistent? Some are relative (`./apollo/`), some are absolute (`apollo/`). We tend to favour absolute

src/index.js
```diff
@@ -1,28 +1,37 @@
...
 import createHistory from "history/createHashHistory";
 import configureStore from "redux/configureStore";
-import createClient from "apollo/createClient";
 import App from "containers/App";
 import Raven from "raven-js";
+import getIdForObject from "utils/getIdForObject";
+import fragmentMatcher from "./apollo/fragmentMatcher";
+import createApolloClient from "apollo/createApolloClient";
+import createApolloCache from "apollo/createApolloCache";
+import createSchemaLink, { createSchema } from "./apollo/createSchemaLink";
```
WickyNilliams: You should move these into the `if` statement below, and change to `require`. That way all code associated with the mock API is omitted from any build which doesn't use it (e.g. production)

src/index.js
```diff
@@ -1,28 +1,37 @@
...
 if (process.env.REACT_APP_USE_MOCK_API === "true") {
-  const mockNetworkInterface = require("./apollo/createMockNetworkInterface")
-    .default;
-  const mockResolvers = require("./tests/utils/MockResolvers").default;
-  networkInterface = mockNetworkInterface(mockResolvers);
+  const resolvers = require("./tests/utils/MockResolvers").default;
+  link = createSchemaLink(createSchema(resolvers));
 } else {
-  networkInterface = require("./apollo/createRealNetworkInterface").default;
+  link = createHttpLink(process.env.REACT_APP_API_URL);
```
WickyNilliams: After you've done the above, to be consistent, can you require `createHttpLink` here rather than import at the top. This also means that all the real API stuff is omitted from any build using the mock API

PR #299, opened 2018-02-19
218-improve-webpack-configs

SamGodwin2:
The webpack config files for both dev and prod had a large amount of duplicated code, this PR is intended to export as much of this common code as possible to both clean up the files and allow for expansion if necessary. 

### How to review 
App should be able to be built in both prod and dev while having a minimum amount of duplicated code, and all smoketests should pass on it.

### DIFF
--- a/config/webpack.config.common.js
+++ b/config/webpack.config.common.js
@@ -1,0 +1,126 @@
+/* eslint-disable import/unambiguous */
+"use strict";
+
+const path = require("path");
+const eslintFormatter = require("react-dev-utils/eslintFormatter");
+const ModuleScopePlugin = require("react-dev-utils/ModuleScopePlugin");
+const paths = require("./paths");
+
+module.exports.resolve = {
+  // This allows you to set a fallback for where Webpack should look for modules.
+  // We placed these paths second because we want `node_modules` to "win"
+  // if there are any conflicts. This matches Node resolution mechanism.
+  // https://github.com/facebookincubator/create-react-app/issues/253
+  modules: [paths.appSrc, "node_modules", paths.appNodeModules].concat(
+    // It is guaranteed to exist because we tweak it in `env.js`
+    process.env.NODE_PATH.split(path.delimiter).filter(Boolean)
+  ),
+  // These are the reasonable defaults supported by the Node ecosystem.
+  // We also include JSX as a common component filename extension to support
+  // some tools, although we do not recommend using it, see:
+  // https://github.com/facebookincubator/create-react-app/issues/290
+  extensions: [".js", ".json", ".jsx"],
+  alias: {
+    // Support React Native Web
+    // https://www.smashingmagazine.com/2016/08/a-glimpse-into-the-future-with-react-native-for-web/
+    "react-native": "react-native-web"
+  },
+  plugins: [
+    // Prevents users from importing files from outside of src/ (or node_modules/).
+    // This often causes confusion because we only process files within src/ with babel.
+    // To fix this, we prevent you from importing files out of src/ -- if you'd like to,
+    // please link the files into your node_modules/ and let module-resolution kick in.
+    // Make sure your source files are compiled, as they will not be processed in any way.
+    new ModuleScopePlugin(paths.appSrc)
+  ]
+};
+
+module.exports.loaders = {
+  eslint: {
+    // TODO: Disable require.ensure as it's not a standard language feature.
+    // We are waiting for https://github.com/facebookincubator/create-react-app/issues/2176.
+    // { parser: { requireEnsure: false } },
+
+    // First, run the linter.
+    // It's important to do this before Babel processes the JS.
+    test: /\.(js|jsx)$/,
+    enforce: "pre",
+    use: [
+      {
+        options: {
+          formatter: eslintFormatter
+        },
+        loader: require.resolve("eslint-loader")
+      }
+    ],
+    include: paths.appSrc
+  },
+
+  fileExclusions: {
+    // ** ADDING/UPDATING LOADERS **
+    // The "file" loader handles all assets unless explicitly excluded.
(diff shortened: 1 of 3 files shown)


SamGodwin2:
[Improve webpack configs (1)](https://trello.com/c/1bZndkHm/218-improve-webpack-configs-1)

config/commonWebpackConfigs.js
```diff
@@ -0,0 +1,236 @@
...
+
+module.exports.ExtractTextPlugin = require("extract-text-webpack-plugin");
+module.exports.getClientEnvironment = require("./env");
+module.exports.webpack = require("webpack");
+module.exports.HtmlWebpackPlugin = require("html-webpack-plugin");
+module.exports.InterpolateHtmlPlugin = require("react-dev-utils/InterpolateHtmlPlugin");
+
+// This is the development configuration.
+// It is focused on developer experience and fast rebuilds.
+// The production configuration is different and lives in a separate file.
```
WickyNilliams: This comment is not relevant to this file. Should be placed in dev file

config/commonWebpackConfigs.js
```diff
@@ -0,0 +1,236 @@
...
+  // A missing `test` is equivalent to a match.
+  test: [/\.bmp$/, /\.gif$/, /\.jpe?g$/, /\.png$/],
+  loader: require.resolve("url-loader"),
+  options: {
+    limit: 10000,
+    name: "static/media/[name].[hash:8].[ext]"
+  }
+};
+
+module.exports.prodBabelCompiler =
```
WickyNilliams: Why have a production-specific option in the common config?

config/commonWebpackConfigs.js
```diff
@@ -0,0 +1,236 @@
...
+// "style" loader normally turns CSS into JS modules injecting <style>,
+// but unlike in development configuration, we do something different.
+// `ExtractTextPlugin` first applies the "postcss" and "css" loaders
+// (second argument), then grabs the result CSS and puts it into a
+// separate file in our build process. This way we actually ship
+// a single CSS file in production instead of JS code injecting <style>
+// tags. If you use code splitting, however, any async bundles will still
+// use the "style" loader inside the async code so CSS from them won't be
+// in the main CSS file.
+module.exports.prodCssCompiler = {
```
WickyNilliams: Same as above.

config/commonWebpackConfigs.js
```diff
@@ -0,0 +1,236 @@
...
+      }
+    }
+  ] // Note: this won't work without `new ExtractTextPlugin()` in `plugins`.
+};
+
+module.exports.gqlCompiler = {
+  test: /\.(graphql|gql)$/,
+  exclude: /node_modules/,
+  loader: "graphql-tag/loader"
+  // ** STOP ** Are you adding a new loader?
```
WickyNilliams: This comment no longer makes sense at this position. As it sounds like a fairly important message, it should be added back where it makes sense

config/commonWebpackConfigs.js
```diff
@@ -0,0 +1,236 @@
...
+          { removeTitle: true },
+          { convertColors: { shorthex: false } },
+          { convertPathData: false }
+        ]
+      }
+    }
+  ]
+};
+
+module.exports.devCssCompiler =
```
WickyNilliams: Why have dev-specific stuff in common?

config/commonWebpackConfigs.js
```diff
@@ -0,0 +1,236 @@
+/* eslint-disable import/unambiguous */
+"use strict";
+
+const autoprefixer = (module.exports.autoprefixer = require("autoprefixer"));
```
WickyNilliams: Seems a little pointless to export all these things from node_modules (since you're not doing anything with them or configuring them), when they could just be required directly in the dev/prod configs. 

For example, `path` module is part of node standard library, there's no need to import it only to immediately export. It adds a layer of indirection that isn't beneficial

config/commonWebpackConfigs.js
```diff
@@ -0,0 +1,236 @@
...
+    // Prevents users from importing files from outside of src/ (or node_modules/).
+    // This often causes confusion because we only process files within src/ with babel.
+    // To fix this, we prevent you from importing files out of src/ -- if you'd like to,
+    // please link the files into your node_modules/ and let module-resolution kick in.
+    // Make sure your source files are compiled, as they will not be processed in any way.
+    new ModuleScopePlugin(paths.appSrc)
+  ]
+};
+
+module.exports.eslintLoader = {
```
WickyNilliams: I wonder if these should all be under a single object, `loaders` or something. That way, in the prod/dev configs you can do:

```js
const { loaders } = require("./commonWebpackConfigs");

// then later...

module: {
    strictExportPresence: true,
    rules: [
      loaders.eslint,
      loaders.file,
      // etc...
    ]
  },
```

config/commonWebpackConfigs.js
```diff
@@ -0,0 +1,236 @@
...
+      options: {
+        formatter: eslintFormatter
+      },
+      loader: require.resolve("eslint-loader")
+    }
+  ],
+  include: paths.appSrc
+};
+
+module.exports.fileCompiler = {
```
WickyNilliams: Why is this suffixed with "compiler", while others are "loader"? Seems like an arbitrary distinction, and they're all loaders in webpack terminology. If we went with my above suggestion of having a `loaders` object, then the suffixes can be dropped

config/commonWebpackConfigs.js
```diff
@@ -0,0 +1,126 @@
+/* eslint-disable import/unambiguous */
```
samiwel: Sorry to be pedantic but any chance we can rename this file `webpack.config.common.js` for consistency.

PR #302, opened 2018-02-23
Fix missing "created by" on Questionnaires list page

WickyNilliams:
The user's display name wasn't being supplied when creating a questionnaire, so it would be displayed as "Unknown" when listing questionnaires.

Unfortunately it's difficult/impossible to write a unit test for this. I added a test to our cypress suite to cover this instead.

The basic fix was to pull the user from the store in the `QuestionnairesPage` container. To get the test passing with mock API I needed to also modify the `MockDataStore` as it didn't support createdBy. This test passes with both mock and real API.

### How to review 
Tests pass, code looks good

### DIFF
--- a/cypress/integration/author_spec.js
+++ b/cypress/integration/author_spec.js
@@ -18,8 +18,21 @@ describe("eq-author", () => {
 
   it("can create a questionnaire", () => {
     cy.get("[data-test='create-questionnaire']").click();
-    setQuestionnaireSettings("Title");
+    setQuestionnaireSettings("My Questionnaire Title");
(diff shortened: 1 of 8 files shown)


WickyNilliams:
I wonder if we should modify the schema to make `createdBy` non-nullable. It was added as nullable to ensure temporary backwards compatibility (no longer necessary)

WickyNilliams:
["created by" not populated when questionnaire created](https://trello.com/c/ZpeWe7PE/384-created-by-not-populated-when-questionnaire-created)

PR #305, opened 2018-02-27
Fixing broken storybook components.

SamGodwin2:
Fixed the broken storyboard components.

### How to review 
All storyboard components should work as expected.

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -6,11 +6,7 @@ import { setOptions } from "@storybook/addon-options";
 import App from "components/App";
 
 addDecorator(StoryRouter());
-addDecorator(story =>
-  <App>
-    {story()}
-  </App>
-);
+addDecorator(story => <App>{story()}</App>);
 
 function requireAll(requireContext) {
   return requireContext.keys().map(requireContext);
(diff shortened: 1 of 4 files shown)


src/components/AnswerTypeSelector/story.js
```diff
@@ -10,13 +10,17 @@ const Wrapper = styled.div`
...
 
-const Decorator = storyFn =>
-  <Wrapper>
-    {storyFn()}
-  </Wrapper>;
+const Decorator = storyFn => <Wrapper>{storyFn()}</Wrapper>;
+
+const answers = {
+  length: 1
+};
```
WickyNilliams: This should be an array. I would just inline it below:

```js
<AnswerTypeSelector
  answers={[]}
  onSelect={action("Answer type selected")}
/>
```

.storybook/config.js
```diff
@@ -5,12 +5,8 @@ import { setOptions } from "@storybook/addon-options";
 
 import App from "components/App";
 
-addDecorator(StoryRouter());
-addDecorator(story =>
-  <App>
-    {story()}
-  </App>
-);
+// addDecorator(StoryRouter());
```
WickyNilliams: I think I commented this out when we were trying to fix RTE story... can probably be reinstated?

PR #308, opened 2018-02-27
Implement toggle chip component.

samiwel:
image

 - Refactoring of initial component created by Sam G.
 - Fix linting errors and update snapshots.
 - Add test for covering click to focus functionality.

### How to review 
Launch storybook an examine the ToggleChip component. Check that it behaves as expected and matches the design.
All existing tests and checks should pass.

### DIFF
--- a/src/components/ToggleChip/ToggleChip.story.js
+++ b/src/components/ToggleChip/ToggleChip.story.js
@@ -1,0 +1,59 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import styled from "styled-components";
+import ToggleChip from "components/ToggleChip";
+import { uniqueId } from "lodash";
+
+const Wrapper = styled.div`
+  padding: 1em;
+  width: 50%;
+`;
+
+class StatefulToggleChip extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      value: false
+    };
+  }
+
+  handleChange = ({ value }) => {
+    this.setState({
+      value
+    });
+  };
+
+  render() {
+    return (
+      <ToggleChip
+        id={uniqueId()}
+        onChange={this.handleChange}
+        checked={this.state.value}
+        {...this.props}
+      />
+    );
+  }
+}
+
+storiesOf("ToggleChip", module)
+  .add("Single", () => (
+    <Wrapper>
+      <StatefulToggleChip>Lorem ipsum dolor sit amet.</StatefulToggleChip>
+    </Wrapper>
+  ))
+  .add("Multiple", () => (
+    <Wrapper>
+      <StatefulToggleChip>Lorem ipsum dolor.</StatefulToggleChip>
+      <StatefulToggleChip>Lorem ipsum dolor sit.</StatefulToggleChip>
+      <StatefulToggleChip>Lorem ipsum.</StatefulToggleChip>
+      <StatefulToggleChip>Lorem ipsum dolor sit amet.</StatefulToggleChip>
+      <StatefulToggleChip>Lorem ipsum dolor sit.</StatefulToggleChip>
+      <StatefulToggleChip>Lorem ipsum.</StatefulToggleChip>
+      <StatefulToggleChip>Lorem.</StatefulToggleChip>
+      <StatefulToggleChip title="Lorem ipsum dolor sit amet, consectetur adipisicing elit. Earum quos saepe voluptatem. Ab consectetur dolore ea eaque excepturi praesentium voluptate.">
+        Lorem ipsum dolor sit amet, consectetur adipisicing elit. Earum quos
+        saepe voluptatem. Ab consectetur dolore ea eaque excepturi praesentium
+        voluptate.
+      </StatefulToggleChip>
+    </Wrapper>
+  ));
--- a/src/components/ToggleChip/index.js
+++ b/src/components/ToggleChip/index.js
@@ -57,18 +57,15 @@ const Field = styled.div`
   margin-bottom: 0.2em;
 `;
 
-const borderSize = 2;
-
 export const Label = styled.label`
   display: flex;
   align-items: center;
   user-select: none;
   border-radius: 2.75em;
   font-size: 0.8em;
-  padding: 0 0.8em 0 2.5em;
+  padding: 0.5em 0.8em 0.5em 2.5em;
   cursor: pointer;
-  border: ${borderSize}px solid;
-  height: calc(1.875rem + ${2 * borderSize}px);
+  border: 2px solid;
   ${props => (props.checked ? labelStyles.checked : labelStyles.unchecked)};
 `;
 
--- a/src/components/ToggleChip/index.js
+++ b/src/components/ToggleChip/index.js
@@ -1,0 +1,152 @@
+import React from "react";
+import PropTypes from "prop-types";
+import styled, { css } from "styled-components";
+import { colors } from "constants/theme";
+import checkedIcon from "./checked.svg";
+
+const labelStyles = {
+  checked: css`
+    color: ${colors.white};
+    background: #008dd0;
+    border-color: #008dd0;
+
+    &:hover {
+      background: #007ab3;
+    }
+  `,
+  unchecked: css`
+    color: ${colors.text};
+    background: #f8f8f8;
+    border-color: rgba(0, 141, 208, 0);
+
+    &:hover {
+      background: #eeeded;
+    }
+  `
+};
+
+const focusStyles = {
+  checked: css`
+    color: ${colors.white};
+    background: #008dd0;
+    border-color: #005e8b;
+  `,
+  unchecked: css`
+    color: ${colors.text};
+    background: #eeeded;
+    border-color: #008dd0;
+  `
+};
+
+const checkboxStyle = {
+  checked: css`
+    border-color: white;
+    background: white url(${checkedIcon}) no-repeat center;
+    background-size: contain;
+  `,
+  unchecked: css`
+    background-color: rgba(255, 255, 255, 0);
+    border-color: #979797;
+  `
+};
+
+const Field = styled.div`
+  display: inline-block;
+  position: relative;
+  margin-right: 0.2em;
+  margin-bottom: 0.2em;
+`;
+
+const borderSize = 2;
+
+export const Label = styled.label`
+  display: flex;
+  align-items: center;
+  user-select: none;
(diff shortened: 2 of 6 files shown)


src/components/ToggleChip/index.js
```diff
@@ -0,0 +1,180 @@
...
+  ${props => (props.checked ? bodyStyle.checked : bodyStyle.unchecked)};
+`;
+
+const Truncated = styled.div`
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+`;
+
+const checkboxSize = 1.2;
```
WickyNilliams: Seems a little unnecessary to extract this var

src/components/ToggleChip/index.js
```diff
@@ -0,0 +1,180 @@
...
+  };
+
+  inputRef = input => {
+    this.hiddenInput = input;
+  };
+
+  render() {
+    const { children, checked, onChange, width } = this.props;
+    return (
+      <FlexInline role="switch" aria-checked={checked} width={width}>
```
WickyNilliams: Not convinced we need the role/aria stuff here, since there is a genuine checkbox below (which should convey the same semantics)

src/components/ToggleChip/index.js
```diff
@@ -0,0 +1,180 @@
...
+          id={this.id}
+          type="checkbox"
+          onChange={onChange}
+          checked={checked}
+          ref={this.inputRef}
+        />
+        <ToggleChipBody
+          checked={checked}
+          onClick={this.handleToggle}
+          title={children}
```
WickyNilliams: `children` propType is `node`, so this won't reliably work. 

I also just did some reading, and it seems like the `title` attribute is not a great choice as some screen readers announce it, whilst others do not. Not sure how best to handle this from an accessibility point of view. Might be worth asking around

src/components/ToggleChip/index.js
```diff
@@ -0,0 +1,180 @@
...
+    size: PropTypes.number,
+    children: PropTypes.node
+  };
+
+  constructor(props) {
+    super(props);
+    this.id = this.props.id || this.generateUniqueId();
+  }
+
+  generateUniqueId = () => {
```
hamishtaplin: Should components be generating their own id's? I'm not so sure

src/components/ToggleChip/index.js
hamishtaplin: Agreed. This should just be a label/input as per example from prototype. No need to hide the input away and then create a mock one.

src/components/ToggleChip/toggleChip.test.js
```diff
@@ -0,0 +1,56 @@
...
+    wrapper = createWrapper(props);
+  });
+
+  it("should render", () => {
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it("should render checked", () => {
+    let wrapper = createWrapper(
+      Object.assign({}, props, { id: "chiptoggle-1", checked: true })
```
hamishtaplin: Use destructuring here.

src/components/ToggleChip/toggleChip.test.js
```diff
@@ -0,0 +1,56 @@
...
+      target: {
+        name: props.name,
+        checked: true
+      }
+    });
+  });
+
+  it("should focus checkbox when clicked", () => {
+    wrapper = createWrapper(props, mount);
+    wrapper.find("ToggleChip__ToggleChipBody").simulate("click");
```
hamishtaplin: Don't use CSS classes as selectors in tests.

src/components/ToggleChip/index.js
```diff
@@ -0,0 +1,180 @@
...
+  align-items: center;
+  border-radius: 2.75em;
+  font-size: 0.8em;
+  padding: 0 0.8em;
+  cursor: pointer;
+  border: 2px solid;
+  height: calc(1.875rem + 4px);
+  max-width: 16em;
+  transition: background-color ${durationInMillis / 2}ms ease-in-out,
+    border-top-color ${durationInMillis / 2}ms ease-in-out,
```
hamishtaplin: I think you can just transition `border-color` instead of doing them all separately

src/components/ToggleChip/toggleChip.test.js
samiwel: This is actually a component name rather than a CSS class, but I'm going to re-work this anyway 👍

src/components/ToggleChip/index.js
samiwel: This was copied over from the ToggleSwitch I believe. It's just a way of generating unique Id's for checkboxes if multiple instances are added and no Ids are specified.

I'm not convinced that every checkbox needs an Id. Rather a name is probably more important. So we can probably remove Id generation here.

src/components/ToggleChip/index.js
samiwel: Ah cool, I thought so but WebStorm wouldn't auto-complete for me, so wasn't sure.

src/components/ToggleChip/index.js
samiwel: To prevent screen reader from reading the value twice aria-hidden has been added to Truncated. That's the advice that was given on some thread I was reading on SO, will try dig out the link.

src/components/ToggleChip/index.js
samiwel: Was just a convenience thing for adjusting the size during development. I can inline.

src/components/ToggleChip/toggleChip.test.js
hamishtaplin: Oh yeah, I didn't even know you could do that. 👍

samiwel:
[Toggle Chip Component (3)](https://trello.com/c/UgVntR1k/370-toggle-chip-component-3)

src/components/ToggleChip/index.js
```diff
@@ -0,0 +1,156 @@
...
+  border: 2px solid;
+  height: calc(1.875rem + 4px);
+  transition: background-color ${durationInMillis / 2}ms ease-in-out,
+    border-color ${durationInMillis / 2}ms ease-in-out;
+  ${props => (props.checked ? labelStyles.checked : labelStyles.unchecked)};
+`;
+
+const Truncated = styled.span`
+  display: inline-block;
+  max-width: ${props => props.width || 16}em;
```
WickyNilliams: use `defaultProps` for this, and the prop should probably be called `maxWidth`

src/components/ToggleChip/index.js
```diff
@@ -0,0 +1,156 @@
...
+    background-color: rgba(255, 255, 255, 0);
+    border-color: #979797;
+  `
+};
+
+const Field = styled.div`
+  display: inline-flex;
+  flex-direction: row-reverse;
+  justify-content: center;
+  align-items: center;
```
WickyNilliams: I think all this flexbox stuff can be dropped in favour of `display: inline-block`

src/components/ToggleChip/index.js
```diff
@@ -0,0 +1,156 @@
...
+  display: flex;
+  align-items: center;
+  position: relative;
+  user-select: none;
+  border-radius: 2.75em;
+  font-size: 0.8em;
+  padding: 0 0.8em 0 2.5em;
+  cursor: pointer;
+  border: 2px solid;
+  height: calc(1.875rem + 4px);
```
WickyNilliams: Why `calc` here?

src/components/ToggleChip/index.js
```diff
@@ -0,0 +1,156 @@
...
+  align-items: center;
+  position: relative;
+  margin-right: 0.2em;
+  margin-bottom: 0.2em;
+`;
+
+export const Label = styled.label`
+  display: flex;
+  align-items: center;
+  position: relative;
```
WickyNilliams: I don't think this is necessary any more

src/components/ToggleChip/index.js
samiwel: The 4px is to account for the 2px border on both the bottom and top. I'll extract a variable to make it clearer.

src/components/ToggleChip/index.js
```diff
@@ -0,0 +1,156 @@
...
+const Truncated = styled.span`
+  display: inline-block;
+  max-width: ${props => props.width || 16}em;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+`;
+
+export const Input = styled.input`
+  width: 1.2em;
```
hamishtaplin: Increase width/height to `1.5em`. This sorts out the weird shape.

src/components/ToggleChip/index.js
hamishtaplin: I would increase this to `30em`—no real need to truncate this small at the moment.

src/components/ToggleChip/index.js
```diff
@@ -0,0 +1,156 @@
...
+  transition: background-color ${durationInMillis}ms ease-in-out
+    ${durationInMillis / 2}ms;
+  ${props => (props.checked ? checkboxStyle.checked : checkboxStyle.unchecked)};
+
+  &:focus {
+    opacity: 1;
+    outline: none;
+  }
+
+  /* stylelint-disable */
```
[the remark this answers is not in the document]
samiwel: @hamishtaplin Two rules get fired here. Something about the interpolation of another styled component in the selector.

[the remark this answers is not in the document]
samiwel: image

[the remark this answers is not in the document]
samiwel: The `selector-type-case` would usually be sensible rule to have, but in the case when we reference another component in the selector, where we have a capitalised naming convention going on, we have to ignore this rule.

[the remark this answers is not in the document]
samiwel: Yeah, I figured we'd look at disabling this rule for the project when we look at https://trello.com/c/yOW4xdZh/383-improve-stylelint

src/components/ToggleChip/ToggleChip.story.js
```diff
@@ -0,0 +1,59 @@
...
+    return (
+      <ToggleChip
+        id={uniqueId()}
+        onChange={this.handleChange}
+        checked={this.state.value}
+        {...this.props}
+      />
+    );
+  }
+}
```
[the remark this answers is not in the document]
samiwel: I need to read more about it, sounds useful.

PR #310, opened 2018-02-28
Fix guidance not being persisted to API

WickyNilliams:
Fixes bug where question guidance would not be persisted to API. Adds test to prevent regression.

### How to review 
Change looks good, tests pass


WickyNilliams:
[Question guidance not being saved](https://trello.com/c/bShgRvD7/387-question-guidance-not-being-saved)

PR #315, opened 2018-03-08
Enable pressing ENTER to add additional options.

samiwel:
enter-adds-options

It was identified that being able to press the ENTER key when adding Checkbox/Radio options would help to improve the user experience.

This PR enables changes Checkbox and Radio answers so that pressing the ENTER key adds a new Option and brings the new option into focus.

### How to review 
All existing tests and checks should pass.
Have a play around to make sure it behaves as expected.
Pressing the ENTER key in other fields should not add new answers.

### DIFF
--- a/src/components/Answers/MultipleChoiceAnswer/Option.js
+++ b/src/components/Answers/MultipleChoiceAnswer/Option.js
@@ -11,6 +11,9 @@ import Tooltip from "components/Tooltip";
 import { CHECKBOX, RADIO } from "constants/answer-types";
 import { get } from "lodash";
 import optionFragment from "graphql/fragments/option.graphql";
+import getIdForObject from "utils/getIdForObject";
+
+const ENTER_KEY = 13;
 
 export const DeleteContainer = styled.div`
   padding: 0.2em;
@@ -104,12 +107,19 @@ export class StatelessOption extends Component {
     onChange: PropTypes.func.isRequired,
     onUpdate: PropTypes.func.isRequired,
     onDelete: PropTypes.func.isRequired,
+    onEnterKey: PropTypes.func,
     hasDeleteButton: PropTypes.bool.isRequired,
     type: PropTypes.oneOf([RADIO, CHECKBOX]).isRequired
   };
 
   handleDeleteClick = e => {
     this.props.onDelete(this.props.option.id);
+  };
+
+  handleKeyDown = e => {
+    if (e.keyCode === ENTER_KEY) {
+      this.props.onEnterKey(e);
+    }
   };
 
   renderDeleteButton() {
@@ -131,7 +141,7 @@ export class StatelessOption extends Component {
     const { hasDeleteButton, option, onChange, onUpdate, type } = this.props;
 
     return (
-      <StyledOption key={option.id}>
+      <StyledOption id={getIdForObject(option)} key={option.id}>
         <LabelField>
           <DummyInput type={type} />
           <SeamlessLabel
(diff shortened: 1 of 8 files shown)


src/components/Answers/MultipleChoiceAnswer/Option.js
```diff
@@ -128,10 +129,17 @@ export class StatelessOption extends Component {
...
+      option,
+      onChange,
+      onUpdate,
+      onKeyDown,
+      type
+    } = this.props;
 
     return (
-      <StyledOption key={option.id}>
+      <StyledOption id={`Option${option.id}`} key={option.id}>
```
WickyNilliams: use `getIdForObject(option)`

src/components/Answers/MultipleChoiceAnswer/index.js
```diff
@@ -77,10 +79,15 @@ class MultipleChoiceAnswer extends Component {
 
   handleAddOptionClick = e => {
     e.preventDefault();
-
     return this.props.onAddOption(this.props.answer.id).then(focusOnEntity);
   };
 
+  handleKeyDown = e => {
+    if (e.keyCode === ENTER_KEY) {
+      this.handleAddOptionClick(e);
```
WickyNilliams: Can we rename `handleAddOptionClick` to just `handleAddOption` since it's no longer click specific?

src/components/Answers/MultipleChoiceAnswer/index.js
```diff
@@ -77,10 +79,15 @@ class MultipleChoiceAnswer extends Component {
 
   handleAddOptionClick = e => {
     e.preventDefault();
-
     return this.props.onAddOption(this.props.answer.id).then(focusOnEntity);
   };
 
+  handleKeyDown = e => {
```
WickyNilliams: I'm wondering if this logic should actually go into `Option`, and the prop for `Option` be `onEnterKey` (or something like that), so then it would read like:

```js
<Option onEnterKey={this.handleAddOption} />
```

which feels like it describes the requirement nicely

src/components/WrappingInput/index.js
```diff
@@ -74,6 +75,9 @@ class WrappingTextArea extends React.Component {
     if (e.keyCode === ENTER_KEY) {
       e.preventDefault();
     }
+    if (this.props.onKeyDown) {
```
WickyNilliams: How about providing a defaultProp of `noop` for this? Saves needing the if statement

samiwel: Yeah, i was wondering if lodash had anything like invokeIfExists

src/components/Answers/MultipleChoiceAnswer/index.js
samiwel: I did have something like this, but Option passes its props down onto wrapping input and onEnterKey is not a valid prop for TextArea which is used under the hood, so you get prop warnings at runtime.

src/components/WrappingInput/index.js
WickyNilliams: `_.invoke(this.props, "keyDown", e)`  should do it, i think

src/components/Answers/MultipleChoiceAnswer/index.js
WickyNilliams:
```jsclass Option

Comment (WickyNilliams) [reply src/components/Answers/MultipleChoiceAnswer/index.js]: why wouldn't this work (slimmed down version of Option for illustrative purposes)?

```js
class Option extends React.Component {
  handleKeyDown = e => {
    if (e.keyCode === ENTER_KEY) {
      this.props.onEnterKey(e);
    }
  };

  render() {
    return (
      <WrappingInput
        onKeyDown={this.handleKeyDown}
      />
    );
  }
}
```

No unknown props that way?

PR #316, opened 2018-03-09
392 share survey

SamGodwin2:
Added a share questionnaire button, The link it provides will direct to the go-launch-survey and will work multiple times.

### How to review 
All test should pass and if the services are all running correctly the link should direct to the preview page.


src/components/shareQuestionnaire/index.js
```diff
@@ -0,0 +1,115 @@
...
+  text-align: center;
+  margin-bottom: 2.5rem;
+`;
+
+export const CopyButton = styled.button`
+  border-color: #3e7b9c;
+  color: #3e7b9c;
+  height: 3.6em;
+  width: 73px;
+  border-top-right-radius: 10%;
```
hamishtaplin: using a percentage for the `border-radius` here is why you're getting a weird shape.

screenshot 2018-03-09 11 57 59

Use ems or rems instead. Also better to use `border-radius` shorthand, eg. `border-radius: 0 0.5em 0.5em 0`

src/components/shareQuestionnaire/index.js
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  static propTypes = {
+    isOpen: PropTypes.bool,
+    onClose: PropTypes.func,
+    previewLink: PropTypes.string
+  };
+
+  handleClick = () => {
+    var textField = document.createElement("textarea");
```
hamishtaplin: Use `const` here.

hamishtaplin: Why not just utilise the `input` that's already there instead of creating a dummy one, which is a bit of a hack.

src/components/shareQuestionnaire/index.js
```diff
@@ -0,0 +1,115 @@
...
+    textField.innerText = this.props.previewLink;
+    document.body.appendChild(textField);
+    textField.select();
+    document.execCommand("copy");
+    textField.remove();
+    this.setState({
```
hamishtaplin: Do we need to catch error cases here? There's nothing to ensure this has been successful.

Not sure what can go wrong with this, although security policy of IE is one thing that comes to mind—clipboard access is considered insecure.

src/components/shareQuestionnaire/index.js
```diff
@@ -0,0 +1,115 @@
...
+  render() {
+    return (
+      <StyledModal isOpen={this.props.isOpen} onClose={this.handleClose}>
+        <Message>
+          <CenteredHeading>Share Questionnaire</CenteredHeading>
+        </Message>
+        <InlineField id="done">
+          <Label>URL to share</Label>
+          <SucessLabel done={this.state.done}>
+            <img src={tick} alt="Done" />Copied to clipboard
```
hamishtaplin: Better to use a background image for this, `img` tag should be reserved for content whereas this decoration. Also, alignment needs fixing:

screenshot 2018-03-09 12 21 35

The icon should be centre-aligned with the text and the whole thing aligned left against the label, as per designs.

src/components/shareQuestionnaire/index.js
```diff
@@ -0,0 +1,115 @@
...
+    this.props.onClose();
+  };
+
+  render() {
+    return (
+      <StyledModal isOpen={this.props.isOpen} onClose={this.handleClose}>
+        <Message>
+          <CenteredHeading>Share Questionnaire</CenteredHeading>
+        </Message>
+        <InlineField id="done">
```
hamishtaplin: The `Label` and `Input` both need to be inside a single `Field` for it to wire up the `for`/`id` attributes which associate the `label` with the `input`.  Currently there's two labels for the id `done` but that doesn't actually exist.

I would also move the "copied to clipboard" bit inside the first label (inside a `span`)—you shouldn't really have 2 labels for an input.

src/components/shareQuestionnaire/index.js
```diff
@@ -0,0 +1,115 @@
...
+          <Label>URL to share</Label>
+          <SucessLabel done={this.state.done}>
+            <img src={tick} alt="Done" />Copied to clipboard
+          </SucessLabel>
+        </InlineField>
+        <InlineField id="link">
+          <StraightInput
+            disabled
```
hamishtaplin: I guess `disabled` is to stop people editing the URL before they copy. This also prevents any interaction (including focus) and would effectively hide the input from a screenreader. Better to disable editing the text by just making the `onChange` handler a 'no-op' function. Pretty sure this would work.

src/components/shareQuestionnaire/index.js
```diff
@@ -0,0 +1,115 @@
...
+          <CenteredHeading>Share Questionnaire</CenteredHeading>
+        </Message>
+        <InlineField id="done">
+          <Label>URL to share</Label>
+          <SucessLabel done={this.state.done}>
+            <img src={tick} alt="Done" />Copied to clipboard
+          </SucessLabel>
+        </InlineField>
+        <InlineField id="link">
+          <StraightInput
```
hamishtaplin: As per above, this input is un-labelled as it would be passed the `id="link"` from the parent Field.

src/components/shareQuestionnaire/index.js
```diff
@@ -0,0 +1,115 @@
...
+  };
+
+  handleClose = () => {
+    this.setState({ done: false });
+    this.props.onClose();
+  };
+
+  render() {
+    return (
+      <StyledModal isOpen={this.props.isOpen} onClose={this.handleClose}>
```
hamishtaplin: IMO, It's nicer to unpack all the props at the start of the render function than repeat the `this.props.` everywhere:

`const {  isOpen,  previewLink } = this.props`

This also makes things easier if you want to refactor to a stateless component as you can just copy the props object into the paramaters.

I often do this for `state` too.

src/components/shareQuestionnaire/index.js
hamishtaplin: Also, check the button text size and weight against the designs.

src/components/Header/index.js
```diff
@@ -81,6 +91,22 @@ export class UnconnectedHeader extends React.Component {
...
+                </IconButton>
+              )}
+              {questionnaire && (
+                <ShareQuestionnaire
+                  isOpen={this.state.isModalOpen}
+                  onClose={this.handleModalClose}
+                  previewLink={this.getPreviewUrl(questionnaire.id)}
+                />
+              )}
               {questionnaire && (
```
hamishtaplin: Why not just put the whole thing in one `questionnaire &&`? You can use a [fragment](https://reactjs.org/docs/fragments.html) to avoid having to wrap the components in a DOM element.

hamishtaplin: Also, design has the two buttons the other way around.

src/components/shareQuestionnaire/index.js
```diff
@@ -0,0 +1,115 @@
...
+
+const StraightInput = styled(Input)`
+  border-radius: 0;
+  border-right: none;
+`;
+
+const labelStyles = {
+  done: css`
+    color: #3e7b9c;
+    font-weight: 100;
```
hamishtaplin: What is this font-weight for? We use 400, 700 & 900.

SamGodwin2: Thats just me trying to eye it based on the images. Might need to pick up some new glasses.

src/components/Header/index.js
```diff
@@ -82,12 +107,22 @@ export class UnconnectedHeader extends React.Component {
...
+                    title="Preview"
+                    target="_blank"
+                  />
+                  <ShareButton
+                    icon={shareIcon}
+                    iconOnly
+                    onClick={this.handleShare}
+                    highlightOnHover={false}
+                  >
+                    Share
```
samiwel: Probably worth checking with Sarah what the copy should be for this. "Share" doesn't give any indication of what clicking the button actually does. To me "Copy sharable preview link to clipboard" is a bit more descriptive, but perhaps a bit wordy.

src/components/Header/index.js
```diff
@@ -46,9 +57,13 @@ export class UnconnectedHeader extends React.Component {
   static propTypes = {
     questionnaire: CustomPropTypes.questionnaire,
     user: CustomPropTypes.user,
-    signOutUser: PropTypes.func.isRequired
+    signOutUser: PropTypes.func.isRequired,
+    raiseToast: PropTypes.func
```
samiwel: Probably should make raiseToast a required prop, since we're never checking before using it.

SamGodwin2:
[Share survey](https://trello.com/c/U5Jex1CP/392-share-survey)

src/components/Header/index.js
```diff
@@ -82,12 +107,23 @@ export class UnconnectedHeader extends React.Component {
...
+                    target="_blank"
+                  />
+                  <ShareButton
+                    title="Share"
+                    icon={shareIcon}
+                    iconOnly
+                    onClick={this.handleShare}
+                    highlightOnHover={false}
+                  >
+                    Create link for sharing
```
WickyNilliams: Is this what was settled on as copy?

src/containers/ToastContainer/index.js
```diff
@@ -63,17 +63,19 @@ const ToastArea = ({ toasts, dismissToast, undoToast, ...otherProps }) => {
...
+                        undoToast(
+                          id,
+                          otherProps[toast.undoAction],
+                          toast.context
+                        );
+                      }}
+                    >
+                      Undo
+                    </UndoButton>
+                  )}
```
WickyNilliams: Can you add a test for this - that `UndoButton` is not rendered if no `undoAction`

charlotte-hirst:
Functionality wise I like it - in terms of the copy just check with Sarah.

src/containers/ToastContainer/ToastContainer.test.js
```diff
@@ -0,0 +1,44 @@
+import React from "react";
+import { shallow } from "enzyme";
+import { ToastArea, UndoButton } from "./index";
+
+describe("Toastcontainer", () => {
+  let dismissToast = jest.fn();
+
+  let undoToast = jest.fn();
```
samiwel: These are never re-assigned, so could be declared as `const` and moved inside the `beforeEach`.

WickyNilliams: in any case, you should never be reusing mocks between tests - they will be "polluted" with results from previous tests

src/components/Header/Header.test.js
```diff
@@ -54,6 +69,18 @@ describe("components/Header", () => {
...
+    it("should render a share button", () => {
+      const shareButton = wrapper.find({ title: "Share" });
+
+      expect(shareButton).toHaveLength(1);
+      expect(shareButton).toMatchSnapshot();
+    });
+
+    it("should let a user copy", () => {
+      wrapper.find({ title: "Share" }).simulate("click");
+      expect(document.execCommand).toHaveBeenCalled();
```
samiwel: Probably worth making sure `.toHaveBeenCalledWith("copy")`?

samiwel:
@SamGodwin2 Looks good, can you rebase against master then squash down the commits and I'll merge it in.

PR #319, opened 2018-03-13
Consolidated the two separate Icon Button components

SamGodwin2:
There were two separate Icon button components being used in separate parts of the application this PR has consolidated both the IconButtons into a single component. 

### How to review 
The Toolbar should operate and look the same as it did before this PR and all tests should continue to pass.


hamishtaplin:
I know it's not knew but it would be good to have a storybook for this?

src/components/RichTextEditor/Toolbar.test.js
```diff
@@ -125,12 +125,12 @@ describe("components/RichTextEditor/Toolbar", () => {
 
   describe("Button", () => {
     it("matches snapshot", () => {
-      const wrapper = mount(<Button icon="foo.svg" title="foo" />);
+      const wrapper = mount(<Button icon={() => <svg />} title="foo" />); // eslint-disable-line
```
hamishtaplin: Could we be specific about which rule is being disabled here?

src/components/RichTextEditor/PipingMenu.js
```diff
@@ -18,6 +18,28 @@ const validAnswerTypes = {
...
+  height: 3.5em;
+
+  &:hover {
+    opacity: 1;
+  }
+
+  &[disabled] {
+    opacity: 0.5;
+  }
+`;
```
WickyNilliams: Why can't we use the regular IconButton here?

src/components/IconDecorated/IconButton/index.js
```diff
@@ -1,40 +0,0 @@
...
-
-IconButton.propTypes = {
-  icon: PropTypes.string.isRequired,
-  title: PropTypes.string.isRequired,
-  disabled: PropTypes.bool,
-  handleClick: PropTypes.func,
-  className: PropTypes.string
-};
-
-export default IconButton;
```
WickyNilliams: Now all this has been deleted, do we still need the `IconDecorated` directory? Could you simplify `IconLink` as well? And move it up to `components` dir?

src/components/RichTextEditor/PipingMenu.js
SamGodwin2: Tried that but the withComponent function seems to override the the component its called on. If I use the IconButton here the RMLMenuButton seems to override it and make the icon button disappear.

SamGodwin2: screen shot 2018-03-14 at 13 12 27

Makes it looks like that.

samiwel:
Good clean up. I think this is definitely a step in the right direction. But I wonder if there's more that can be done to consolidate the two components.

`IconLink` and `IconButton` now have slightly inconsistent APIs now. `IconLink` expects a filename to load as a `background-image` and `IconButton` expects a Function (that yields a component). Although I guess it's not a deal breaker, and we're probably not using `IconLink` in that many places, I think we should strive to keep the APIs feeling consistent where possible.

src/components/IconButton/index.js
```diff
@@ -50,7 +50,8 @@ class IconButton extends React.Component {
     icon: PropTypes.func.isRequired,
     iconOnly: PropTypes.bool,
     highlightOnHover: PropTypes.bool,
-    children: PropTypes.node
+    children: PropTypes.node,
+    title: PropTypes.string
```
WickyNilliams: You shouldn't need to list `title` as a prop, since all other props are passed to the button component anyway

src/components/IconButton/index.js
```diff
@@ -59,7 +60,7 @@ class IconButton extends React.Component {
   };
 
   render() {
-    const { icon: Icon, children, iconOnly, ...otherProps } = this.props;
+    const { icon: Icon, children, iconOnly, title, ...otherProps } = this.props;
```
WickyNilliams: So no need to `title` here

src/components/IconButton/index.js
```diff
@@ -50,7 +50,8 @@ class IconButton extends React.Component {
     icon: PropTypes.func.isRequired,
     iconOnly: PropTypes.bool,
```
WickyNilliams: I can't comment on the specific line, but `iconOnly` is passed to `StyledButton` in render - which seems to be superfluous, `StyledButton` doesn't utilise that prop at all. So you can remove that

src/components/RichTextEditor/Toolbar.js
```diff
@@ -144,7 +146,9 @@ class ToolBar extends React.Component {
             onToggle(button);
           }
         }}
-      />
+      >
+        {button.title}
+      </Button>
```
WickyNilliams: Again, can't comment on the actual line, but there appears to be an issue with `Separator` - it's not visible in the app. Not sure how/why it's broken, since nothing has changed there!

PR #322, opened 2018-03-16
eq-381 Routing Rule Component

hamishtaplin:

[Trello card](https://trello.com/c/WeZsp2SM/381-routing-rule-component-5)

- adds main RoutingRule component encapsulating other Routing components into a CanvasSection with toolbar, etc.
- RoutingRuleset component encapsulates one or more RoutingRules
- moved `Toolbar` and `Buttons` into a module so they can be composed elsewhere. NOTE: I would normally recommend we refactor this a bit more but there's some major design changes in the pipeline so sensible to leave it for now.

### How to review 

- check storybook examples against AC set out in the Trello card


src/components/routing/RoutingCondition.js
```diff
@@ -108,7 +113,7 @@ RoutingCondition.propTypes = {
   selectedPage: CustomPropTypes.page.isRequired,
   id: PropTypes.string.isRequired,
   onPageChange: PropTypes.func.isRequired,
-  onRemoveClick: PropTypes.func.isRequired,
+  onRemoveClick: PropTypes.func,
```
WickyNilliams: rename -> `onRemove`?

src/components/routing/RoutingRule.js
```diff
@@ -0,0 +1,86 @@
...
+const RoutingRule = ({
+  children,
+  page,
+  onDeleteRule,
+  onThenChange,
+  onAddRule,
+  title,
+  sections
+}) => (
+  <React.Fragment>
```
WickyNilliams: This fragment is redundant?

src/components/routing/RoutingRuleEmptyMsg.js
```diff
@@ -0,0 +1,57 @@
...
+  flex-direction: column;
+  padding: 1em 0;
+`;
+
+const Title = styled.h2`
+  font-size: 1.8em;
+  font-weight: 600;
+  margin-bottom: 0.5em;
+  &::before {
+    content: url(${iconRouting});
```
WickyNilliams: do screen readers read out the url if you do this vs:

```
content: "",
background-image: `url(${iconRouting})`;
```

?

src/components/routing/RoutingStatement.js
```diff
@@ -9,24 +9,27 @@ const CenteringColumn = styled(Column)`
...
   <Fragment>
     {children}
-    <Grid align="center">
-      <CenteringColumn gutters={false} cols={1.5}>
-        <TextButton onClick={onAddCondition} data-test="btn-add">
-          AND
-        </TextButton>
-      </CenteringColumn>
-    </Grid>
+    {onAddCondition && (
```
WickyNilliams: When would you not pass this?

hamishtaplin: When you can't add a condition, eg. the examples in the storybook for 'no answers' or 'unsupported answer type'.

src/components/routing/RoutingRuleEmptyMsg.js
hamishtaplin: I don't think they read the url. It didn't even read a character out when we tested using it for use with inputs with a type (eg. currency)

PR #323, opened 2018-03-19
Fixes #314

### LINKED ISSUE #314: Weird focus style on Answer delete button.
### Expected behaviour
Focus style should align with other focus styles across the Author application.

### Actual behaviour
Weird shaped focus box is observed.

Focus on the delete answer button.

### Technical information
Observed on fresh checkout of `master`.

#### Browser
Chrome 64.0.3282.186 (Official Build) (64-bit)

#### Operating System
Mac OSX

### Screenshot
image

SamGodwin2:
Fixed the weird overlapping focus style on the delete answer button.
37143315-b820e42e-22b2-11e8-83c0-6519ccbf405d

### How to review 
Focus should look correct and all tests should pass.
screen shot 2018-03-19 at 10 35 48


src/components/AnswerEditor/index.js
```diff
@@ -26,9 +26,9 @@ import TextAreaAnswer from "components/Answers/TextAreaAnswer";
 export const AnswerDeleteButton = styled(DeleteButton)`
   position: absolute;
   right: 0.33em;
-  top: 0.33em;
+  top: 0.2em;
   width: 1em;
-  height: 1em;
+  height: 1.3em;
   line-height: 1;
```
samiwel: Not sure about the change to `top` value here. I think removing `height` and `line-height` attributes will fix the weird focus style.

Probably better to remove these as I think it's generally better not to use fixed height. @hamishtaplin can probably confirm this.

SamGodwin2: Yep that's a better solution, ta.

hamishtaplin: Using `height` is fine for things like this as it doesn't really have any child elements.

I'd prefer the button to be square though. Just adding `overflow: hidden` instead of these changes fixes the problem.

PR #324, opened 2018-03-19
Fixes section title not being truncated in navigation

### LINKED ISSUE #306: Section title not being truncated in the navigation
### Expected behaviour

All text in the navigation should truncate if too long to fit in the available space.

### Actual behaviour

Section title just overflows.

Just enter a section title that is long enough to reproduce this bug

### Screenshot

screenshot 2018-02-27 14 05 08

SamGodwin2:
This PR fixes #306 and where the section title wasn't being correctly truncated. As seen here:
36733131-884915f6-1bc7-11e8-9af3-b00903da9f51

### How to review 
The section title should now truncate correctly and tests should continue to pass.
screen shot 2018-03-19 at 11 01 04

### DIFF
--- a/src/components/NavigationSidebar/SectionTitle.js
+++ b/src/components/NavigationSidebar/SectionTitle.js
@@ -16,6 +16,8 @@ const textInverted = "#E1E1E1";
 const Link = styled(NavLink)`
   text-decoration: none;
   color: ${textInverted};
+  overflow: hidden;
+  text-overflow: ellipsis;
 
   &:link,
   &:visited {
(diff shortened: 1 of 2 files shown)


WickyNilliams:
I wonder if we should add a `title` attribute, so that you can hover the mouse to get the full title? @hamishtaplin

SamGodwin2:
Makes sense I've set it up as a tooltip here, not sure if this quite looks right though...
screen shot 2018-03-20 at 11 08 54

PR #327, opened 2018-03-22
390 - Move page modal

WickyNilliams:
* Adds modal `MovePageModal` which enables user to pick a section and position to move a page to.
* Not integrated with the rest of the app yet
* Refactor `Modal` component to be more flexible with styling, and remove the need for global styles

### How to review 
Inspect code, test in storybook


src/components/Modal/Modal.test.js
```diff
@@ -38,19 +38,19 @@ describe("components/Modal", () => {
 
     it("should close if overlay is clicked", () => {
       wrapper = createWrapper(props, mount, children);
-      wrapper.find(".ReactModal__Overlay").simulate("click");
+      wrapper.find(".Overlay").simulate("click");
```
samiwel: Is there any way of not relying on the css class name here, or is that the nature of the ReactModal library?

src/components/MovePageModal/story.js
```diff
@@ -0,0 +1,107 @@
...
+          section={questionnaire.sections[0]}
+          page={questionnaire.sections[0].pages[0]}
+          onMovePage={action("movePage")}
+        />
+      </div>
+    );
+  }
+}
+
+storiesOf("MovePageModal", module)
```
samiwel: This works well, but I think it would be nice to see the actions that get called too within storybook. Can you amend with relevant action('names') just to get a sense of when, e.g. onSubmit, onClose etc. are called?

src/components/Modal/Modal.test.js
WickyNilliams: Unfortunately not. We control the className though, so it won't randomly break

src/components/MovePageModal/story.js
WickyNilliams: Not sure what you mean? `onClose` is pretty obvious - it's called when it closes. `onSubmit` is not exposed. And there's already one for `onMovePage`, which is the crucial action

samiwel: Ah I didn't spot you were already using the actions addon to display the movePage action. Ignore my previous comment!

samiwel:
[Sortable item list pop up components (8)](https://trello.com/c/u62qHGjn/390-sortable-item-list-pop-up-components-8)

PR #329, opened 2018-03-27
Added frozen-lockfile flag to travis install step

SamGodwin2:
What is the context of this PR?
Travis seems to generate its own lockfile instead of using the one in the github repo. Which means tests can pass on travis but not be up to date when run locally.

How to review
Travis should fail if the lock-file is out of date with the package.json file and should pass if the inverse is true.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ addons:
   chrome: stable
 
 install:
- - yarn install
+ - yarn install --frozen-lockfile
  - yarn upgrade eq-author-graphql-schema
 
 before_script:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -14,7 +14,7 @@ addons:
   chrome: stable
 
 install:
- - yarn install
+ - yarn install --frozen-lockfile
  - yarn upgrade eq-author-graphql-schema
```
WickyNilliams: this line can be removed now, right @samiwel? iirc this was from when we were using git hashes for versions

samiwel: Yes we can remove this now 👍

WickyNilliams:
Needs rebase

PR #330, opened 2018-03-27
Add split button component.

samiwel:
This PR introduces a SplitButton component.
The split button is just a wrapper which applies styles to a button and a Popout component. It is required as part of the acceptance criteria for the add "other" option for checkbox and radio Answers.

### How to review 
The split button component can be viewed in Storybook.
A code review is required and all existing tests and checks should pass.

### DIFF
--- a/src/components/Popout/index.js
+++ b/src/components/Popout/index.js
@@ -15,10 +15,24 @@ const Container = styled.div`
 
 const Layer = styled.div`
   position: absolute;
-  bottom: 0;
-  left: 0;
+  ${props => `${props.horizontalAlignment}: ${props.offsetX}`};
+  ${props => `${props.verticalAlignment}: ${props.offsetY}`};
   z-index: 10;
 `;
+
+Layer.defaultProps = {
+  horizontalAlignment: "left",
+  verticalAlignment: "bottom",
+  offsetX: "0",
+  offsetY: "0"
+};
+
+Layer.propTypes = {
+  horizontalAlignment: PropTypes.oneOf(["left", "right"]),
+  verticalAlignment: PropTypes.oneOf(["top", "bottom"]),
+  offsetX: PropTypes.string,
+  offsetY: PropTypes.string
+};
 
 const DefaultTransition = props => (
   <CSSTransition {...props} timeout={0} classNames="" />
@@ -32,7 +46,11 @@ class Popout extends React.Component {
     onToggleOpen: PropTypes.func.isRequired,
     onEntered: PropTypes.func,
     onExited: PropTypes.func,
-    transition: PropTypes.any // eslint-disable-line react/forbid-prop-types
+    transition: PropTypes.any, // eslint-disable-line react/forbid-prop-types
+    horizontalAlignment: Layer.propTypes.horizontalAlignment,
+    verticalAlignment: Layer.propTypes.verticalAlignment,
+    offsetX: Layer.propTypes.offsetX,
+    offsetY: Layer.propTypes.offsetY
   };
 
   static defaultProps = {
@@ -121,7 +139,13 @@ class Popout extends React.Component {
           ref: trigger => (this.trigger = trigger)
         })}
 
-        <TransitionGroup component={Layer}>
+        <TransitionGroup
+          component={Layer}
+          horizontalAlignment={this.props.horizontalAlignment}
+          verticalAlignment={this.props.verticalAlignment}
+          offsetX={this.props.offsetX}
+          offsetY={this.props.offsetY}
+        >
           {this.props.open ? this.renderContent() : null}
         </TransitionGroup>
       </Container>
--- a/src/components/SplitButton/MenuButton.js
+++ b/src/components/SplitButton/MenuButton.js
@@ -1,0 +1,49 @@
+import React from "react";
+import styled from "styled-components";
+import Chevron from "./chevron.svg?inline";
+import { colors } from "constants/theme";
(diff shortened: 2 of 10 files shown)


samiwel:
[Add/edit 'other' answer (8)](https://trello.com/c/3i8vZj7b/70-add-edit-other-answer-8)

src/components/Popout/index.js
```diff
@@ -15,11 +15,21 @@ const Container = styled.div`
 
 const Layer = styled.div`
   position: absolute;
-  bottom: 0;
-  left: 0;
+  left: ${props => props.offsetX}em;
+  bottom: ${props => props.offsetY}em;
```
SamGodwin2: Might be cleaner to pass in values for left or right and top or bottom and use percentages for the offsets,

hamishtaplin: I don't think it should rely on magic numbers for positioning. Can either calculate the offset or rely on the natural flow of the document by making the dropdown container a sibling of the `SplitButton` itself and then absolutely positioning it underneath to remove it from the document flow.

src/components/SplitButton/index.js
```diff
@@ -0,0 +1,61 @@
...
+    border-right: 0;
+    border-top-right-radius: 0;
+    border-bottom-right-radius: 0;
+
+    &:focus {
+      ${focusStyle};
+    }
+  }
+
+  .SplitMenuButton {
```
hamishtaplin: Check the styling against the designs:

screenshot 2018-03-27 12 44 50


screenshot 2018-03-27 12 46 06

src/components/SplitButton/index.js
```diff
@@ -0,0 +1,61 @@
...
+  transition: background-color 100ms ease-out;
+`;
+
+const focusStyle = css`
+  color: #ffffff;
+  background: #3e7b9c;
+  outline: 0;
+`;
+
+const SplitButton = styled(ButtonGroup)`
```
hamishtaplin: This doesn't really seem to need to use `ButtonGroup`? The whole thing is more of a `DropDown` imo.

src/components/SplitButton/MenuButton.js
```diff
@@ -0,0 +1,40 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { UncontrolledPopout } from "components/Popout";
+import Button from "components/Button";
+import VisuallyHidden from "components/VisuallyHidden";
+import Chevron from "./chevron.svg?inline";
+
+class Trigger extends React.Component {
```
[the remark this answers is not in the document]
samiwel: This can't be a stateless functional component since the Popout manages a ref and you can't store a ref for a stateless component. You get errors in the console.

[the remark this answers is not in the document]
hamishtaplin: ah fair enough 👍

samiwel:
@hamishtaplin @SamGodwin2 please can you review the changes I've made since your original comments?

src/components/SplitButton/MenuButton.js
```diff
@@ -0,0 +1,69 @@
...
+        {this.props.children}
+      </UncontrolledPopout>
+    );
+  }
+}
+
+MenuButton.propTypes = {
+  children: PropTypes.node.isRequired,
+  horizontalAlignment: PropTypes.oneOf(["left", "right"]),
+  verticalAlignment: PropTypes.oneOf(["top", "bottom"])
```
SamGodwin2: Not sure but would it be better to just code in the necessary alignments here or put in some default values for when we call the popout rather than pass them down from above each time, can't imagine we'd ever want a orientation other than right: 0 and left: null as any other combination looks really weird with the design?

src/components/SplitButton/SplitButton.story.js
```diff
@@ -0,0 +1,34 @@
...
+import { action } from "@storybook/addon-actions";
+
+const Wrapper = styled.div`
+  padding: 1em;
+  width: 25em;
+`;
+
+storiesOf("SplitButton", module).add("Default", () => (
+  <Wrapper>
+    <SplitButton>
```
WickyNilliams: I think in this component we've gone too far down the composition road. There's a lot of work necessary whenever you want to use a split button. And the actual `SplitButton` itself has no logic, barely any styles, and doesn't contain a button - which to me suggests that responsibility hasn't been divided correctly.

It's worth re-thinking how this component works. Perhaps an API something like this:

```js
<SplitButton primaryAction={whatever} primaryText="foobar">
  {/* whatever you put in here is rendered when open */}
</SplitButton>
```

If we ever need something more flexible we can refactor later.

I also wonder whether you should be able to pass the open/closed state to the component so that a parent can close it when, for example, something is selected from the popout.

src/components/SplitButton/MenuButton.js
```diff
@@ -0,0 +1,68 @@
...
+      stroke: ${colors.white};
+    }
+  }
+`;
+
+class Trigger extends React.Component {
+  render() {
+    return (
+      <SplitMenuButton {...this.props}>
+        <VisuallyHidden>Menu</VisuallyHidden>
```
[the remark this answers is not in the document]
samiwel: Have changed to "Show additional options"

samiwel:
[Create SplitButton component](https://trello.com/c/iQSkYk0X/410-create-splitbutton-component)

WickyNilliams:
Looks great, just needs squashing

PR #335, opened 2018-04-05
Upgrade to react 16.3.1

SamGodwin2:
React is upgrading to a new minor version, 16.3. One of the features of this update is the depreciation of the componentWillMount, componentWillReceiveProps and componentWillUpdate. These methods will no longer be supported in a few versions down the line and so this PR replaces these methods with alternative ones. 

### How to review 
All tests should pass and the deprecated methods should not be present anywhere in the code base.


SamGodwin2:
[Upgrade to React 16.3 (5)](https://trello.com/c/hnT8Gr2J/400-upgrade-to-react-163-5)

src/components/Modal/index.js
```diff
@@ -110,10 +112,6 @@ class Modal extends React.Component {
     hasCloseButton: true
   };
 
-  componentWillMount() {
-    ReactModal.setAppElement("body");
```
hamishtaplin: Can this just be moved into `componentDidMount` instead or is that too late in the lifecycle?

src/components/withEntityEditor/index.js
```diff
@@ -23,12 +24,16 @@ const withEntityEditor = (entityPropName, fragment) => WrappedComponent => {
       };
     }
 
-    componentWillReceiveProps(nextProps) {
-      if (nextProps[entityPropName].id !== this.props[entityPropName].id) {
-        this.setState({
+    static getDerivedStateFromProps(nextProps, prevState) {
+      if (
+        !isNull(prevState) &&
```
WickyNilliams: in what circumstances will `prevState` be null?

package.json
```diff
@@ -87,7 +87,7 @@
     "react-dev-utils": "^4.1.0",
     "react-router-test-context": "^0.1.0",
     "react-svg-loader": "^2.1.0",
-    "react-test-renderer": "^16.0.0",
+    "react-test-renderer": "16.3.1",
```
WickyNilliams: probably want the `^` hat on all these changed dependencies

src/components/Modal/index.js
WickyNilliams: @hamishtaplin not sure it's worth even putting in a lifecycle - it's a static method, so it affects all Modal components

src/components/Popout/index.js
```diff
@@ -50,14 +50,14 @@ class Popout extends React.Component {
...
+  componentDidMount() {
     if (this.props.open) {
       this.bindRootCloseHandlers();
     }
   }
 
-  componentWillReceiveProps(nextProps) {
-    if (nextProps.open) {
+  componentDidUpdate(prevProps, prevState) {
+    if (!prevProps.open) {
```
WickyNilliams: This isn't right. Some other prop could have changed but open might still be false. 

Instead use `if(this.props.open)`

src/components/withEntityEditor/index.js
```diff
@@ -13,22 +16,19 @@ const withEntityEditor = (entityPropName, fragment) => WrappedComponent => {
...
-    componentWillReceiveProps(nextProps) {
-      if (nextProps[entityPropName].id !== this.props[entityPropName].id) {
-        this.setState({
-          [entityPropName]: nextProps[entityPropName]
-        });
+    static getDerivedStateFromProps(nextProps, prevState) {
+      if (prevState.id !== getEntityId(nextProps)) {
+        return {
+          [entityPropName]: nextProps[entityPropName],
+          id: getEntityId(nextProps)
```
WickyNilliams: What's this about?

samiwel: @WickyNilliams I think the above code updates the entity in state if the Id's are different. The change looks like it is to use `getDerrivedStateFromProps` which apparently is a [safer alternative](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops) to what we were previously using (`componentWillReceiveProps`).

@SamGodwin2 I think you can remove the need to set the initial state to a null/empty value above by simply doing a check before you try and reference the previous state Id.

WickyNilliams: Sorry, I specifically meant adding `id` to state, where it wasn't before (it's superfluous, since the entity in state will have the `id`).

BTW you need to set some initial state first. React complains otherwise. @SamGodwin2 and I went through this last week.

samiwel:
Did you forget to check in the changes to remove `id` prop from `StatelessBasicAnswer`?

PR #336, opened 2018-04-05
390 move page modal improvements

WickyNilliams:
This PR modifies MovePageModal so that selections are reset once section/position modal is closed. For example: open section modal, select a section, click "cancel" button. The selection should return back to what it was before opening the modal. 

This PR also includes a slight accessibility improvement: using an actual `label` for the modal trigger (and associating them with each other, via `for`/`id` attributes)

### How to review 
Changes look good


WickyNilliams:
[Sortable item list pop up components (8)](https://trello.com/c/u62qHGjn/390-sortable-item-list-pop-up-components-8)

PR #337, opened 2018-04-05
Used focus-within to clean up canvas styles

SamGodwin2:
Previously a redux store was being used to track focus styles across the different canvas sections, by using focus-within however the functionality is able to remain the same without needing to use a redux store. 

### How to review 
Tests should pass and all components should look and respond in the same way as they do before this PR.


src/components/EditorSurface/CanvasSection.js
```diff
@@ -1,54 +1,26 @@
...
-export const BasicSection = styled.div`
+const BasicSection = styled.div`
   padding: 2em 2.5em;
   background-color: white;
   position: relative;
   box-shadow: ${shadow};
   margin-bottom: 1em !important;
-`;
-
-const FocusableSection = styled(BasicSection)`
```
hamishtaplin: I think we still want this distinction between `BasicSection` and a `FocusableSection` as not all sections need to be focusable.

WickyNilliams: Reinstate but rename `FocusableSection` -> `CanvasSection`

src/components/QuestionPageEditor/index.js
```diff
@@ -26,14 +24,14 @@ import withDeleteOption from "containers/enhancers/withDeleteOption";
 import * as ToastActionCreators from "redux/toast/actions";
 import QuestionPageToolbar from "./QuestionPageToolbar";
 
-const AddAnswerSection = BasicSection.extend`
+const AddAnswerSection = CanvasSection.extend`
```
hamishtaplin: This should be a `BasicSection`—we don't want to highlight it when children have focus (hence the need to keep `BasicSection` around).

src/components/EditorSurface/CanvasSection.js
```diff
@@ -21,34 +11,18 @@ export const BasicSection = styled.div`
...
-  transition: outline-color 100ms ease-in;
-  margin-bottom: 1px;
-  outline: 1px solid transparent;
-  ${props => props.isFocused && focusedStyle};
-
+const CanvasSection = styled(BasicSection)`
   &:hover {
+    outline: 1px solid transparent;
     outline-color: ${rgba(colors.blue, 0.5)};
```
samiwel: Can the `-color` not be collapsed into the line above?

PR #338, opened 2018-04-05
Set default Cypress baseUrl and clean up CI script.

samiwel:
We aren't setting the default baseUrl for running Cypress tests, which places extra configuration burden on developers.

This PR sets this as a default and updates the travis CI script to invoke a yarn command for running cypress in CI.

### How to review 
Changes look sensible and build should pass.


package.json
```diff
@@ -22,7 +22,7 @@
     "storybook-deploy": "yarn storybook-build && gh-pages -d .out",
     "precommit": "lint-staged",
     "coverage": "yarn test -- --coverage",
-    "cypress:open": "cypress open"
+    "cypress:open": "CYPRESS_baseUrl=http://localhost:3000 cypress open"
```
SamGodwin2: Does this give the option to input a baseurl other then localhost:3000, just incase we need to point it at an alternate url?

hamishtaplin: Shouldn't this get the `CYPRESS_baseUrl` from the environment vars?

WickyNilliams:
```sh
CYPRESS_base_Url=${CYPRESS_base_Url:-http://localhost:3000/} cypress open
```

the above will make it overridable if necessary (say you want to run tests against preprod)

PR #339, opened 2018-04-12
eq-415 New Header/Button Styles

hamishtaplin:

[Trello](https://trello.com/c/9ycq3cOD/415-new-header)

- Implements design of new header
- Rewrote `Button` component with new styles
- Introduced `LinkButton` component for external, preview link
- Introduced `RouteButton` component for buttons that navigate internally
- Removed `IconButton` component, moving icon+text pattern into `IconText`
- Primary blue colour updated (more refactoring around colours required)

### How to review 

- Check header against designs
- Check buttons all work as expected


hamishtaplin:
[New Header](https://trello.com/c/9ycq3cOD/415-new-header)

src/components/EditorSurface/Toolbar.js
```diff
@@ -5,10 +5,10 @@ export const Toolbar = styled.div`
...
   border-bottom: 1px solid ${colors.borders};
   margin: 0 0 1em;
-  padding: 0.25em 1em;
+  padding: 0;
   align-items: center;
 `;
 
 export const Buttons = styled.div`
-  margin: 0 -0.5em 0 auto;
+  margin: 0 0 0 auto;
```
WickyNilliams: Out of curiosity: I've never seen `auto` used on a single side, what does it do?

hamishtaplin: pushes it to the other side

src/components/RichTextEditor/ToolbarButton.js
```diff
@@ -0,0 +1,52 @@
...
+    opacity: 0.2;
+    pointer-events: none;
+  }
+
+  ${props => props.active && activeState};
+`;
+
+ToolbarButton.propTypes = {
+  active: PropTypes.bool
+};
```
WickyNilliams: Why can't we use our fancy new `IconButton` for this?

src/constants/theme.js
```diff
@@ -1,18 +1,20 @@
...
   red: "#D0021B",
   text: "#4A4A4A",
   orange: "#F5A623",
   borders: "#EAEAEA",
   white: "#FFFFFF",
-  highlight: "#008ad2"
+  highlight: "#FDBD56"
 };
 
+colors.primary = colors.blue;
```
WickyNilliams: This doesn't seem to be used anywhere?

src/components/RichTextEditor/ToolbarButton.js
hamishtaplin: Too many styles to override, wasn't worth it.

src/constants/theme.js
hamishtaplin: Ah, I meant to...

src/components/RichTextEditor/ToolbarButton.js
hamishtaplin: I think this (floating toolbar) may go away soon, not sure it's going to work when we change how the fields are edited.

src/components/Button/propTypes.js
```diff
@@ -0,0 +1,10 @@
+import PropTypes from "prop-types";
+export const propTypes = {
+  primary: PropTypes.bool,
+  secondary: PropTypes.bool,
```
WickyNilliams: Not sure bools are great for this - what if you set a button to be `primary` and `secondary`? Same with light and naked? four bools mean 16 possible combos - how many are genuinely valid?

samiwel:
Kinda feels a bit weird that the preview and share buttons look more rectangular with some padding on hover but the logout button has really rounded borders and no padding on the left hand side.

image
image
image

hamishtaplin:
@samiwel would look even weirder if the hover style was the same as the other buttons due to the shape/size of the avatar

src/components/Button/propTypes.js
samiwel: Booleans can be useful to apply the styles in flexible ways, but Nick's right. I guess there's some combinations that just wouldn't make sense. e.g. you'd probably never want to allow a button that is both primary and secondary. Wonder if this is the time to make this change?

src/components/Button/index.test.js
```diff
@@ -1,23 +1,23 @@
...
+    wrapper = mount(<Button small>Button</Button>);
     expect(wrapper).toMatchSnapshot();
-    wrapper = shallow(<Button clear>Button</Button>);
+    wrapper = mount(<Button naked>Button</Button>);
     expect(wrapper).toMatchSnapshot();
   });
 
   it("renders as disabled", () => {
-    wrapper = shallow(<Button disabled>Button</Button>);
+    wrapper = mount(<Button disabled>Button</Button>);
```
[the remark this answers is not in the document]
hamishtaplin: If we `shallow` then the snapshots just test that the prop was passed. Unsure about this, do we need them at all in that case?

src/components/Button/__snapshots__/index.test.js.snap
```diff
@@ -1,41 +1,356 @@
...
+  >
+    <button
+      className="c0"
+      disabled={true}
+      type="button"
+    >
+      Button
+    </button>
+  </Button__StyledButton>
+</Button>
```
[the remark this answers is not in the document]
hamishtaplin: See above

src/components/Button/propTypes.js
hamishtaplin: You're right but what's the alternative?

Some aren't valid combos but I guess you could enforce with prop-types?

samiwel: Yeah propTypes such as `oneOf['primary', 'secondary']` and a `oneOf['light', 'dark']` would be one approach. The problem is what do you name those props, `importance` and `theme`?

hamishtaplin: I was thinking more along the lines of keeping the existing props and writing custom prop-types that warn if you pass props that conflict.

src/components/Button/index.test.js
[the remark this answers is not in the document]
WickyNilliams: Looking at the `Button` implementation - it is functionally identical to exporting the styled-component itself. Can't we just export that? In which case i'd agree, tests are pointless

src/components/Button/propTypes.js
WickyNilliams: I think custom prop types would get complex and is unnecessary extra work. Better to design things so that the only possible states are valid states (as sam described above)

hamishtaplin: Is this really a problem? This doesn't really change how it works, just adds some more styles.

It'll be obvious if you mess it up.

src/components/Button/LinkButton.js
```diff
@@ -0,0 +1,32 @@
+import React from "react";
+import PropTypes from "prop-types";
+import styled from "styled-components";
+import { styles } from "components/Button/styles";
+import { propTypes } from "components/Button/propTypes";
+
+const StyledLink = styled.a`
+  ${styles};
+`;
```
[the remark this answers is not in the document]
hamishtaplin: Hmm, didn't know about that, will take a look 👍

src/components/Button/__snapshots__/index.test.js.snap
```diff
@@ -1,41 +1,355 @@
...
+}
+
+<button
+  className="c0"
   disabled={true}
+  type="button"
 >
   Button
-</Button__StyledButton>
+</button>
```
samiwel: This is quite a verbose snapshot. Any reason why we need to snapshot all the CSS styles?

hamishtaplin: @WickyNilliams commented on this too. Shallow rendering just snapshots the props were passed. Can probably just get rid if we don't like snapshotting the CSS?

samiwel: It's not a big deal, but I think it's enough to snapshot the `variant` prop as  there's a prop type defined for this so it won't render if invalid prop is passed in. Having CSS in the snapshot just means that any small tweaks to the style / class names break the snapshot tests.

PR #341, opened 2018-04-16
Saving Indicator

SamGodwin2:
Implemented a saving icon to show the user when data is being saved to the db.

### How to review 
Test should pass, the saving icon should be unobtrusive, should only be visible when data is actually being saved and should only be visible for a reasonable length of time.


src/components/AutoSaver/AutoSaver.test.js
```diff
@@ -0,0 +1,59 @@
...
+
+  const advanceByTime = amount => {
+    currentTime += amount;
+    jest.runTimersToTime(currentTime);
+  };
+
+  it("should render when starting saving", () => {
+    const wrapper = shallow(<AutoSaver isSaving={false} />);
+    wrapper.setProps({ isSaving: true });
+    expect(wrapper.state().timeGoing).toEqual(true);
```
WickyNilliams: You shouldn't assert on state as it's an implementation detail. There's always something else you can assert on, since state is always used in `render`, something in the output will use it. Test that instead

src/components/AutoSaver/index.js
```diff
@@ -0,0 +1,102 @@
...
+`;
+
+const AnimatedSection = styled.div`
+  color: ${colors.text};
+  display: flex;
+  position: absolute;
+  right: 1em;
+  top: 1em;
+  align-items: center;
+  opacity: 0;
```
WickyNilliams: We should probably look at using a combination of `visibility:hidden` and `opacity` for this, to take the element out of the document. 

We could also maybe use `aria-hidden` html attribute. `aria-hidden` is good as we can then test for that instead state as you have done above

src/components/AutoSaver/index.js
```diff
@@ -0,0 +1,102 @@
...
+  };
+
+  static defaultProps = {
+    minDisplayTime: 1000
+  };
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      timeGoing: false
```
WickyNilliams: This doesn't feel like a good name. Is the flag to indicate that the timer is running?

src/components/AutoSaver/index.js
```diff
@@ -0,0 +1,102 @@
...
+
+  handleClose = () => {
+    this.setState({
+      timeGoing: false
+    });
+  };
+
+  componentDidUpdate(prevProps, prevState) {
+    if (!prevProps.isSaving && this.props.isSaving) {
+      this.timer.start();
```
WickyNilliams: You need to stop the timer in `componentWillUnmount`

src/components/AutoSaver/AutoSaver.story.js
```diff
@@ -0,0 +1,21 @@
+import React from "react";
```
WickyNilliams: Rename all references `AutoSaver` -> `SavingIndicator`, including files/folders/variable names

src/components/withEntityEditor/index.test.js
```diff
@@ -61,6 +70,7 @@ describe("withEntityEditor", () => {
   });
 
   it("should pass filtered entity to callback onUpdate", () => {
+    wrapper.setState({ isDirty: true });
```
WickyNilliams: Don't use setState, it's an internal implementation detail. Instead `simulate("update")` with some values to make it dirty.

WickyNilliams: All the other changes to withEntityEditor need tests too:

* dirty tracking
* calling the correct action props before requests, and on complete/error
* etc

src/containers/NavigationSidebarContainer/index.js
```diff
@@ -9,9 +9,16 @@ import { raiseToast } from "redux/toast/actions";
 
 import NavigationSidebar from "components/NavigationSidebar";
 import { connect } from "react-redux";
+import { startSaving } from "redux/saving/actions";
+
+export const mapStateToProps = state => {
+  return {
+    meta: state.saving.meta
+  };
+};
```
WickyNilliams: What's this?

src/containers/NavigationSidebarContainer/index.js
```diff
@@ -9,9 +9,16 @@ import { raiseToast } from "redux/toast/actions";
...
 
 export default flowRight(
-  connect(null, { raiseToast }),
+  connect(mapStateToProps, { raiseToast, startSaving }),
```
WickyNilliams: Is this used?

src/containers/QuestionnaireDesignPage/index.js
```diff
@@ -13,10 +15,15 @@ import withCreateSection from "containers/enhancers/withCreateSection";
 import withDeleteSection from "containers/enhancers/withDeleteSection";
 import { getUrlParams } from "utils/UrlUtils";
 
-export const mapStateToProps = (state, { match }) => getUrlParams(match.params);
+export const mapStateToProps = (state, { match }) => {
+  let urlPrams = getUrlParams(match.params);
+  return {
+    ...urlPrams
+  };
```
WickyNilliams: What's this change for?

src/containers/QuestionnaireDesignPage/index.js
```diff
@@ -13,10 +15,15 @@ import withCreateSection from "containers/enhancers/withCreateSection";
...
+};
 
 export default flowRight(
-  connect(mapStateToProps, { raiseToast }),
+  connect(mapStateToProps, { raiseToast, startSaving }),
```
WickyNilliams: Is this needed?

src/redux/saving/reducer.js
```diff
@@ -0,0 +1,25 @@
+import { assign, get } from "lodash";
+import { SAVING, STOP_SAVING } from "redux/saving/actions";
+
+const initialState = {
+  numOfProcesses: 0
```
WickyNilliams: I think processes isn't a great name. `numOfPendingRequests` or `pendingRequestCount` would be better

src/redux/saving/reducer.js
```diff
@@ -0,0 +1,25 @@
...
+};
+
+export default (state = initialState, action) => {
+  switch (action.type) {
+    case SAVING: {
+      return assign({}, state, {
+        numOfProcesses: state.numOfProcesses + 1
+      });
```
WickyNilliams: You can use object spread to simplify things:

```js
return {
  ...state,
  numOfProcesses: state.numOfProcesses + 1
}
```

src/redux/saving/reducer.js
```diff
@@ -0,0 +1,25 @@
...
+    }
+    case STOP_SAVING: {
+      return assign({}, state, {
+        numOfProcesses: state.numOfProcesses - 1
```
WickyNilliams: There should probably be some logic here to prevent this value ever going below 0 (in case of some bug elsewhere):

`Math.max(0, state.numOfProcesses - 1)`

src/redux/saving/actions.js
```diff
@@ -0,0 +1,14 @@
+export const SAVING = "SAVING";
+export const STOP_SAVING = "STOP_SAVING";
+
+export const startSaving = error => {
```
WickyNilliams: Can we call these `startRequest`/`endRequest` (and update the action type appropriately)

src/components/AnswerEditor/answereditor.story.js
```diff
@@ -47,7 +49,13 @@ class AnswerEditorWrapper extends React.Component {
...
 }
 
-const CenterDecorator = storyFn => <CenterXY>{storyFn()}</CenterXY>;
+const store = configureStore();
+
+const CenterDecorator = storyFn => (
+  <Provider store={store}>
+    <CenterXY>{storyFn()}</CenterXY>
+  </Provider>
+);
```
WickyNilliams: We're using Provider and store in a lot of these stories, I wonder if it's worth creating a helper util for creating a Provider decorator. Something like this should work

```js
const createProviderDecorator = initialState => {
  return story => (
    <Provider store={configureStore(null, null, initialState)}>
      {story()}
    </Provider>
  );
};
```

You could then compose it with other decorators like this:

```js
storiesOf("Answer Editor", module)
  .addDecorator(createProviderDecorator())
  .addDecorator(CenterDecorator)
```

src/components/withEntityEditor/index.test.js
```diff
@@ -92,6 +113,45 @@ describe("withEntityEditor", () => {
...
+  it("should call startRequest and stopRequest on failure", () => {
+    const newValue = "foo1";
+    handleUpdate = jest.fn(() =>
+      SynchronousPromise.reject(new Error("message"))
+    );
+    const failingWrapper = render();
+    failingWrapper.setProps({
+      startRequest: handleStartRequest,
+      endRequest: handleEndRequest
+    });
```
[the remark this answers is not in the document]
SamGodwin2: Because if they get passed in as regular props the connect magic seems to overwrite them with its own props so they're nolonger jest.fn functions.

src/components/withEntityEditor/index.test.js
```diff
@@ -87,13 +108,60 @@ describe("withEntityEditor", () => {
...
+    expect(handleUpdate).toHaveBeenCalled();
+  });
+
+  it("should call startRequest on Update and stopRequest Completion", () => {
+    const newValue = "foo1";
+
+    wrapper.setProps({
+      startRequest: handleStartRequest,
+      endRequest: handleEndRequest
+    });
```
samiwel: Any reason why these props cannot be set up in beforeEach and rendered as part of the render function?

src/components/withEntityEditor/index.test.js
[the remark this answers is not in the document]
samiwel: Just read this, ignore my previous comment!

.storybook/config.js
```diff
@@ -2,9 +2,19 @@ import React from "react";
 import { configure, addDecorator } from "@storybook/react";
 import { setOptions } from "@storybook/addon-options";
 import App from "components/App";
+import { Provider } from "react-redux";
+import configureStore from "redux/configureStore";
 
 addDecorator(story => <App>{story()}</App>);
 
+const initialState = {};
```
WickyNilliams: may as well inline this below

src/utils/createMockStore.js
```diff
@@ -0,0 +1,5 @@
+export default () => ({
+  subscribe: jest.fn(),
+  dispatch: jest.fn(),
+  getState: jest.fn()
+});
```
WickyNilliams: this file should be placed in `tests/utils`

src/components/QuestionnaireMeta/story.js
```diff
@@ -3,15 +3,23 @@ import { storiesOf } from "@storybook/react";
...
 
+const store = configureStore();
+
 storiesOf("QuestionnaireMeta", module)
-  .addDecorator(story => <Background>{story()}</Background>)
+  .addDecorator(story => (
+    <Provider store={store}>
+      <Background>{story()}</Background>
+    </Provider>
+  ))
```
WickyNilliams: This can be removed now?

src/components/QuestionnaireSettingsModal/story.js
```diff
@@ -2,17 +2,23 @@ import React from "react";
...
-      id: "1",
-      title: "Lorem ipsum dolor sit, amet consectetur adipisicing elit",
-      navigation: true
-    }}
-    isOpen
-    onSubmit={action("submit")}
-    onClose={action("close")}
-    confirmText="Create"
-  />
+  <Provider store={store}>
```
WickyNilliams: This can be removed now

src/redux/saving/actions.js
```diff
@@ -0,0 +1,14 @@
+export const START_REQUEST = "START_REQUEST";
+export const END_REQUEST = "END_REQUEST";
+
+export const startRequest = error => {
+  return {
+    type: START_REQUEST
+  };
+};
+
+export const endRequest = error => {
```
hamishtaplin: No `error` needed.

src/components/SavingIndicator/index.js
```diff
@@ -0,0 +1,109 @@
...
+  componentWillUnmount() {
+    this.timer.stop();
+  }
+
+  renderIndicator() {
+    return (
+      <FadeTransition>
+        <AnimatedSection data-test="saving-indicator">
+          <Icon src={savingIcon} />
+          Saving...
```
hamishtaplin: Use a proper ellipsis `&hellip;`

src/components/SavingIndicator/index.js
```diff
@@ -0,0 +1,109 @@
...
+
+  componentWillUnmount() {
+    this.timer.stop();
+  }
+
+  renderIndicator() {
+    return (
+      <FadeTransition>
+        <AnimatedSection data-test="saving-indicator">
+          <Icon src={savingIcon} />
```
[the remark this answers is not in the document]
hamishtaplin: the icon itself shouldn't fill the artboard, it just provides consistent sizing/whitespace across all the instances of `IconText`

PR #350, opened 2018-04-26
Fixed table column widths

### LINKED ISSUE #347: Created By field wraps awkwardly for all but the shortest of names
### Expected behaviour

Names are truncated if especially long

### Actual behaviour

Names wrap awkwardly

Sign in as a user with a long name, created quesitonnaire, view on listing page

### Technical information

N/A

#### Browser

N/A

#### Operating System

N/A

### Screenshot

image

WickyNilliams:
The Questionnaires listing previoulsy had flexible columns, and we didn't truncate text in the Created By column. This meant that usernames could awkwardly wrap, as seen below

### How to review 
* Code looks good
* Tests pass
* Table handles long title and long user names
* Check truncation works correctly throughout app (since there is now a single component for this purpose)

**Before**

image

**After**

image

**TODO**

- [x] work out why truncation is broken for MovePageModal

Fixes #347

### DIFF
--- a/src/components/Header/index.js
+++ b/src/components/Header/index.js
@@ -22,6 +22,7 @@ import shareIcon from "./icon-share.svg?inline";
 import previewIcon from "./icon-preview.svg?inline";
 
 import IconText from "components/IconText";
+import Truncated from "../Truncated";
 
 const StyledHeader = styled.header`
   display: flex;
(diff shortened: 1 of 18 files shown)


WickyNilliams:
I need to work out why truncation is no longer working in the MovePageModal before this can be merged (though the problem seems to be in master too, so maybe not worth holding this up over it)

src/components/Header/index.js
```diff
@@ -41,16 +42,14 @@ const QuestionnaireTitle = styled.div`
   white-space: pre;
 `;
 
-const Title = styled.h1`
+const TruncatedTitle = Truncated.withComponent("h1");
+const Title = styled(TruncatedTitle)`
```
samiwel: Couldn't this just become:

```javascript
const Title = Truncated.withComponent("h1")`
  font-size: 1em;
  font-weight: 600;	  font-weight: 600;
  margin: 0 2em;	  margin: 0 2em;
  width: 100%;	  width: 100%;
  text-align: center;	  text-align: center;
  line-height: 1;	  line-height: 1;
`
```

PR #353, opened 2018-05-04
run e2e tests against real API

WickyNilliams:
In order to have more confidence in our build/deploy pipeline, we should run e2e tests agains the real API. This PR changes the e2e script to start docker containers for the API and the DB, then build the front-end to point to these.

### How to review 
Green status from travis confirms it works.

### Todo

- [x] clean up all remnants of mock API


cypress/integration/author_spec.js
```diff
@@ -321,7 +321,7 @@ describe("eq-author", () => {
 
     times(2, i => {
       addQuestionPage();
-      cy.wait(100); // this is needed for real API for some reason
+      cy.wait(200); // this is needed for real API for some reason
```
samiwel: Rather than have an explicit wait here, can we pass in an extended timeout to the `typeIntoDraftEditor` function?

WickyNilliams: I don't think the issue is with `typeIntoDraftEditor`. So it feels wrong to add it a timeout in there, given it works fine everywhere else. I will see if i can use a better selector to help cypress wait on the right element to appear

WickyNilliams:
[Delete mock API (5)](https://trello.com/c/y46GXNp6/430-delete-mock-api-5)

Issue #358, opened 2018-05-14
UI becomes inconsistent when doing multiple move page operations


WickyNilliams:
### Steps to reproduce the behaviour

1. create questionnaire
2. add 3 pages to the existing section - name them "A", "B", "C" so they're easily tracked
3. move page "C" to position 1 (zero-indexed).
4. move page "B" to position 0 (zero-indexed).

### Expected behaviour

Pages shown to have order "B", "A", "C"

### Actual behaviour

Pages shown to have order "C", "A", "B"

### Technical information

This is only an issue with caching on the front-end. If you refresh the browser, then you will see pages in the correct order.

#### Browser

n/a

#### Operating System

n/a

WickyNilliams:
![](https://github.trello.services/images/mini-trello-icon.png) [Inconsistencies when performing multiple move page operations](https://trello.com/c/FmdRvoeg/444-inconsistencies-when-performing-multiple-move-page-operations)

PR #359, opened 2018-05-14
fix inconsistencies with multiple page moves

WickyNilliams:
See #358 

### How to review 
Changes look sensible, confirm that bug is no longer reproducible

### DIFF
--- a/src/containers/enhancers/withMovePage.js
+++ b/src/containers/enhancers/withMovePage.js
@@ -2,14 +2,17 @@ import { graphql } from "react-apollo";
 import movePageMutation from "graphql/movePage.graphql";
 import fragment from "graphql/fragments/movePage.graphql";
 import { getLink } from "utils/UrlUtils";
+import { remove } from "lodash";
 
 export const createUpdater = ({ from, to }) => (proxy, result) => {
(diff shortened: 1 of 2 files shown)


WickyNilliams:
[Inconsistencies when performing multiple move page operations](https://trello.com/c/FmdRvoeg/444-inconsistencies-when-performing-multiple-move-page-operations)

src/containers/enhancers/withMovePage.js
```diff
@@ -2,26 +2,30 @@ import { graphql } from "react-apollo";
...
 export const createUpdater = ({ from, to }) => (proxy, result) => {
   result = result.data.movePage;
 
-  // remove page from previous section
   const fromSectionId = `Section${from.sectionId}`;
   const fromSection = proxy.readFragment({ id: fromSectionId, fragment });
-  const [page] = fromSection.pages.splice(from.position, 1);
+
+  // remove page from previous section and update position values
+  const [page] = remove(fromSection.pages, { id: from.id });
```
samiwel: Just wonder if renaming `page` to `movedPage` would make it clearer that it is not the same as the page referred to in the forEach closure below?

PR #360, opened 2018-05-16
ensure page `position` values are updated on create and delete

WickyNilliams:
Realised that we weren't updating the `position` values for pages, whenever inserting or deleting. This may have caused subtle inconsistencies in the UI.

### How to review 
Tests pass, code looks good


WickyNilliams:
[Page position values fall out of sync after insertion and deletion of pages](https://trello.com/c/UMBiXksx/445-page-position-values-fall-out-of-sync-after-insertion-and-deletion-of-pages)

PR #361, opened 2018-05-16
hide scrollbars on autoresize textarea

WickyNilliams:
On Windows, a scrollbar was shown on the autoresize textarea, causing it not to grow with the content. This PR fixes that.

### How to review 
Change looks good, tests pass, confirmed that bug is fixed on Windows and hasn't introduced issues on OSX


WickyNilliams:
[Windows scroll error](https://trello.com/c/1buH3F4Y/446-windows-scroll-error)

PR #369, opened 2018-06-05
455 saving defect

SamGodwin2:
Saving icon causing issues as it was being was being displayed when the user was offline. This PR handles the case that if a request is made while offline the indicator won't be shown. 

### How to review 
If the internet is down and changes are made to the questionnaire then the saving icon should not be shown. Also changes make sense and test pass.


src/components/Header/index.js
```diff
@@ -111,6 +115,22 @@ export class UnconnectedHeader extends React.Component {
...
+    window.addEventListener("offline", this.updateOnlineStatus);
+  };
+
+  componentWillUnmount = () => {
+    window.removeEventListener("online", this.updateOnlineStatus);
+    window.removeEventListener("offline", this.updateOnlineStatus);
+  };
+
+  updateOnlineStatus = event => {
+    navigator.onLine
```
samiwel: Shouldn't have to check `navigator.onLine` property here. The network status will be passed in by the event.

src/redux/saving/reducer.js
```diff
@@ -1,12 +1,30 @@
...
+  START_REQUEST,
+  END_REQUEST,
+  LOST_CONNECTION,
+  GAIN_CONNECTION
+} from "redux/saving/actions";
 
 const initialState = {
-  pendingRequestCount: 0
+  pendingRequestCount: 0,
+  online: navigator.onLine
```
samiwel: Is there any way of setting the initial state without having to reference the global `navigator` object here?

src/components/App/index.js
```diff
@@ -47,12 +50,38 @@ injectGlobal`
...
+    window.addEventListener("offline", this.updateOnlineStatus);
+  };
 
-App.propTypes = {
-  children: PropTypes.element.isRequired
-};
+  componentWillUnmount = () => {
+    window.removeEventListener("online", this.updateOnlineStatus);
+    window.removeEventListener("offline", this.updateOnlineStatus);
+  };
```
WickyNilliams: why not this?

```js
window.removeEventListener("online", this.props.gainConnection);
window.removeEventListener("offline", this.props.lostConnection);
```

You can rid of `updateOnlineStatus` and `connectivityFunctions` then.

src/components/Header/Header.test.js
```diff
@@ -21,13 +23,19 @@ describe("components/Header", () => {
...
+    gainConnection = jest.fn();
   });
 
   const render = (props = {}) =>
     shallow(
       <UnconnectedHeader
         raiseToast={raiseToast}
         signOutUser={handleSignOut}
+        lostConnection={lostConnection}
+        gainConnection={gainConnection}
```
WickyNilliams: Changes to this file are no longer relevant

src/components/Header/index.js
```diff
@@ -174,6 +174,7 @@ const mapStateToProps = state => ({
   user: getUser(state)
 });
 
-export default connect(mapStateToProps, { signOutUser, raiseToast })(
-  UnconnectedHeader
-);
+export default connect(mapStateToProps, {
+  signOutUser,
+  raiseToast
+})(UnconnectedHeader);
```
WickyNilliams: may as well undo these changes too

src/components/Header/story.js
```diff
@@ -22,11 +22,15 @@ storiesOf("Header", module)
...
+      gainConnection={action("connected")}
+      lostConnection={action("disconected")}
     />
   ))
   .add("Without Utility Buttons", () => (
     <Header
       signOutUser={action("sign out user")}
       raiseToast={action("copied")}
+      gainConnection={action("connected")}
+      lostConnection={action("disconected")}
```
WickyNilliams: And again

src/redux/saving/reducer.js
WickyNilliams: Set this to `true` initially? Code wouldn't have loaded and run if you were offline, so no need to check status?

PR #371, opened 2018-06-19
Added a style filter to the RTE's

SamGodwin2:
There was a bug where the Rich text editors weren't stripping out unsupported styles meaning that you could get styles like a list bullet stuck in the editor. This PR adds a filter package to filter out unwanted styles.

### How to review 
Try pasting a bullet list into the description or title boxes and make sure that the unwanted styles are not present.

### DIFF
--- a/src/components/RichTextEditor/index.js
+++ b/src/components/RichTextEditor/index.js
@@ -19,6 +19,9 @@ import PipedValueDecorator, {
   insertPipedValue
 } from "./entities/PipedValue";
 
+import { filterEditorState } from "draftjs-filters";
+import { mapControlsToFilterConfig } from "./utils/mapControlsToFilterConfig";
+
 import { flow, uniq, map, keyBy, mapValues } from "lodash/fp";
 import { sharedStyles } from "../Forms/css";
 import { Field, Label } from "../Forms";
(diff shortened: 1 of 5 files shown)


src/components/RichTextEditor/index.js
```diff
@@ -233,7 +238,14 @@ class RichTextEditor extends React.Component {
   };
 
   handleChange = (editorState, callback) => {
-    return this.setState({ editorState }, callback);
+    let filteredState = editorState;
+    if (this.props.controls) {
+      filteredState = filterEditorState(
+        mapControlsToFilterConfig(this.props.controls),
```
WickyNilliams: I wonder if this should be generated once, and stored on the instance. e.g. in the constructor:

WickyNilliams: I wonder if this should be generated once and stored on the instance (I don't think there's a need to place into `state`, as it's doesn't change and doesn't affect render) e.g

```js
this.filterConfig = mapControlsToFilterConfig(this.props.controls);
```

Saves doing the same work with every single keypress.

src/components/RichTextEditor/utils/mapControlsToFilterConfig.js
```diff
@@ -0,0 +1,25 @@
+import { mapKeys, includes } from "lodash";
+
+export function mapControlsToFilterConfig(controls) {
+  const mapper = {
+    bold: { format: "BOLD", type: "styles" },
+    emphasis: { format: "ITALIC", type: "styles" },
+    list: { format: "unordered-list-item", type: "blocks" },
+    heading: { format: "header-two", type: "blocks" }
+  };
```
WickyNilliams: how about moving this out of the function, since it's static/unchanging?

src/components/RichTextEditor/utils/mapControlsToFilterConfig.js
```diff
@@ -0,0 +1,25 @@
...
+  const filterConfiguration = {
+    blocks: [],
+    styles: [],
+    entities: [],
+    maxNesting: 0,
+    whitespacedCharacters: []
+  };
+
+  mapKeys(controls, (value, key) => {
+    if (!includes(["piping"], key)) {
```
WickyNilliams: how about `if(mapper[key])`? it doesn't need to know about piping then, and is restricted to types defined in the `mapper`

PR #374, opened 2018-06-26
Alert user when offline

SamGodwin2:
The app need some way to let the user know when they have dropped their internet connection so they can troubleshoot and that their work may not be being saved. 

### How to review 
Run the app then drop your connection, the alert should be very noticeable and remain on screen until you re-connect. Also tests should pass.

### DIFF
--- a/src/components/ExpansionTransition/index.js
+++ b/src/components/ExpansionTransition/index.js
@@ -1,0 +1,41 @@
+import { CSSTransition } from "react-transition-group";
+import PropTypes from "prop-types";
+import styled from "styled-components";
+
+const ExpansionTransition = styled(CSSTransition).attrs({
+  classNames: "expansion"
+})`
+  transition-property: height;
+  transition-duration: ${props => props.timeout}ms;
+  transition-timing-function: ease-in-out;
+
+  &.expansion-enter {
+    height: 0;
+  }
+
+  &.expansion-enter.expansion-enter-active {
+    height: ${props => props.finalHeight};
+  }
+
+  &.expansion-exit {
+    height: ${props => props.finalHeight};
+  }
+
+  &.expansion-exit.expansion-exit-active {
+    height: 0;
+  }
+`;
+
+ExpansionTransition.defaultProps = {
+  in: false,
+  timeout: 250,
+  finalHeight: 1
+};
+
+ExpansionTransition.propTypes = {
+  timeout: PropTypes.number,
+  children: PropTypes.element,
+  finalHeight: PropTypes.string
+};
+
+export default ExpansionTransition;
--- a/src/components/OfflineBanner/index.js
+++ b/src/components/OfflineBanner/index.js
@@ -1,0 +1,69 @@
+import React from "react";
+import PropTypes from "prop-types";
+import styled, { keyframes } from "styled-components";
+import { colors } from "constants/theme";
+import { connect } from "react-redux";
+import { isOnline } from "redux/saving/reducer";
+import WarningIcon from "./icon-warning.svg?inline";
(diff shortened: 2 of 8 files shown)


src/components/ExpansionTransition/index.js
```diff
@@ -0,0 +1,41 @@
+import { CSSTransition } from "react-transition-group";
+import PropTypes from "prop-types";
+import styled from "styled-components";
+
+const ExpansionTransition = styled(CSSTransition).attrs({
+  classNames: "expansionTransition"
```
hamishtaplin: I don't think we put the `transition` bit in the classes elsewheres—seems a bit redundant?

src/components/ExpansionTransition/index.js
```diff
@@ -0,0 +1,41 @@
...
+  transition-property: height;
+  transition-duration: ${props => props.timeout}ms;
+  transition-timing-function: ease-in-out;
+
+  &.expansionTransition-enter {
+    height: 0;
+  }
+
+  &.expansionTransition-enter.expansionTransition-enter-active {
+    height: ${props => props.finalHeight}em;
```
hamishtaplin: Make this unitless (put the `ems` in the prop)—just in case we want to use `rems` or somthing.

src/components/OfflineBanner/index.js
```diff
@@ -0,0 +1,77 @@
...
+  animation-timing-function: ease-in;
+`;
+
+export const UnconnectedOfflineBanner = props => {
+  return (
+    <TransitionGroup>
+      {!props.isOnline && (
+        <StyledExpansionTransition finalHeight={2.5}>
+          <Banner>
+            <WarningMessage icon={warningIcon}>
```
hamishtaplin: convention is to capitalise the first letter of the icon as it's a component, essentially

src/components/OfflineBanner/index.js
```diff
@@ -0,0 +1,77 @@
...
+  }
+`;
+
+const WarningMessage = styled(IconText)`
+  align-items: center;
+  display: flex;
+  color: ${colors.white};
+  animation: ${fade} 750ms;
+  animation-fill-mode: forwards;
+  animation-timing-function: ease-in;
```
hamishtaplin: can this all be a one-liner?

src/components/OfflineBanner/index.js
```diff
@@ -0,0 +1,77 @@
...
+  display: flex;
+`;
+
+const StyledExpansionTransition = styled(ExpansionTransition)`
+  &.expansionTransition-exit {
+    > :first-child {
+      visibility: hidden;
+    }
+  }
+  &.expansionTransition-exit.expansionTransition-exit-active {
```
hamishtaplin: This can be merged into the above selector:
```js
const StyledExpansionTransition = styled(ExpansionTransition)`
  &.expansionTransition-exit,
  &.expansionTransition-exit.expansionTransition-exit-active {
    > :first-child {
      visibility: hidden;
    }
  }
`;
```

src/components/OfflineBanner/index.js
```diff
@@ -0,0 +1,77 @@
...
+      )}
+    </TransitionGroup>
+  );
+};
+
+UnconnectedOfflineBanner.propTypes = {
+  isOnline: PropTypes.bool.isRequired
+};
+
+const mapStateToProps = state => {
```
hamishtaplin: use implicit return: 
```js
const mapStateToProps = state => ({
  isOnline: isOnline(state)
});
```

SamGodwin2:
[Alert user when offline](https://trello.com/c/wsyNC4bI/456-alert-user-when-offline)

PR #375, opened 2018-06-29
459 update toggle and numeric field styles

adammahmood:
- https://trello.com/c/FcPqON0I/459-update-toggle-switch-and-numeric-field-to-match-new-designs
- Updated toggle component to match new designs
- Removed increment/decrement buttons from number components

### How to review 
- Run tests
- See visual changes in Storybook

### DIFF
--- a/src/components/Forms/Number/index.js
+++ b/src/components/Forms/Number/index.js
@@ -3,9 +3,6 @@ import Input from "components/Forms/Input";
 import styled from "styled-components";
 import PropTypes from "prop-types";
 import { clamp, isNaN } from "lodash";
-import arrowIcon from "./arrow.svg";
-
-const spinnerIconSize = "20px";
 
 const StyledDiv = styled.div`
   display: flex;
@@ -24,29 +21,6 @@ const StyledInput = styled(Input)`
   &[type="number"] {
     appearance: textfield;
   }
-`;
-
-const SpinnerButtonWrapper = styled.div`
-  display: flex;
-  flex-direction: column;
-  justify-content: space-between;
-`;
-
-export const SpinnerButton = styled.button`
-  border: none;
-  width: ${spinnerIconSize};
-  height: ${spinnerIconSize};
-  background: url(${arrowIcon}) no-repeat;
-  background-size: contain;
-  cursor: pointer;
-  opacity: 0.6;
-
-  &:hover {
-    opacity: 1;
-    transition: opacity 0.2s ease-in-out;
-  }
-
-  transform: scaleY(${props => (props.inverted ? -1 : 1)});
 `;
 
 class Number extends React.Component {
@@ -87,23 +61,6 @@ class Number extends React.Component {
           aria-live="assertive"
           role="alert"
         />
-        {this.props.showSpinner && (
-          <SpinnerButtonWrapper>
-            <SpinnerButton
-              type="button"
-              title={`Add ${this.props.step}`}
-              onClick={this.handleUp}
-              aria-controls={this.props.id}
-              inverted
-            />
-            <SpinnerButton
-              type="button"
-              title={`Subtract ${this.props.step}`}
-              onClick={this.handleDown}
-              aria-controls={this.props.id}
-            />
-          </SpinnerButtonWrapper>
-        )}
       </StyledDiv>
     );
   }
@@ -111,7 +68,6 @@ class Number extends React.Component {
 
 Number.defaultProps = {
   min: 0,
-  showSpinner: true,
   step: 1
 };
 
@@ -119,7 +75,6 @@ Number.propTypes = {
   id: PropTypes.string,
   name: PropTypes.string,
   onChange: PropTypes.func,
-  showSpinner: PropTypes.bool,
   value: PropTypes.string,
   step: PropTypes.number,
   min: PropTypes.number,
--- a/src/components/ToggleSwitch/index.js
+++ b/src/components/ToggleSwitch/index.js
@@ -6,9 +6,9 @@ import { uniqueId } from "lodash";
 import { Input } from "components/Forms";
 import { colors } from "constants/theme";
 
-const knobColors = {
-  off: colors.white,
-  on: colors.white
+const hitTarget = {
+  height: 3,
+  width: 3
 };
 
 const backgroundColors = {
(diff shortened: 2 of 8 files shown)


WickyNilliams:
[Update Toggle Switch and Numeric field to match new designs](https://trello.com/c/FcPqON0I/459-update-toggle-switch-and-numeric-field-to-match-new-designs)

src/components/Forms/Number/index.js
```diff
@@ -87,39 +61,20 @@ class Number extends React.Component {
...
-              inverted
-            />
-            <SpinnerButton
-              type="button"
-              title={`Subtract ${this.props.step}`}
-              onClick={this.handleDown}
-              aria-controls={this.props.id}
-            />
-          </SpinnerButtonWrapper>
-        )}
```
WickyNilliams: Why remove the buttons for this? The trello card shows them as still existing

adammahmood:
Updated to match attached designs.
number_field

switch_styles

src/components/ToggleSwitch/index.js
```diff
@@ -7,28 +7,31 @@ import { Input } from "components/Forms";
 import { colors } from "constants/theme";
 
 const knobColors = {
-  off: "#E8E8E8",
-  on: colors.blue
+  off: colors.white,
+  on: colors.white
```
WickyNilliams: Can get rid of this now, since the colour is the same regardless of whether checked or not

src/components/ToggleSwitch/index.js
```diff
@@ -55,34 +59,18 @@ const ToggleSwitchKnob = styled.div`
...
-    transform: translateX(
-      ${props => -((props.size * glowScale - props.size) / 2)}em
-    );
-    border-radius: 50%;
-    opacity: 0;
-    transition: opacity 100ms ease-in-out;
-  }
+  border: ${props => (props.checked ? border.on : border.off)};
   /* stylelint-disable declaration-block-semicolon-newline-after */
   ${HiddenInput}:focus + ${ToggleSwitchBackground} &::before {
```
hamishtaplin: This doesn't seem to do anything anymore and can be removed.

src/components/ToggleSwitch/index.js
```diff
@@ -55,34 +59,18 @@ const ToggleSwitchKnob = styled.div`
   display: inline-block;
   height: ${props => props.size}em;
   width: ${props => props.size}em;
-  background: ${props => (props.checked ? knobColors.on : knobColors.off)};
+  background: ${colors.white};
+  top: calc(50% - ${props => props.size}em / 2);
+  left: calc(33% - ${props => props.size}em / 2);
   content: "";
```
hamishtaplin: `content` not needed now this is no longer a pseudo-element.

src/components/ToggleSwitch/index.js
```diff
@@ -38,10 +36,16 @@ export const ToggleSwitchBackground = styled.div`
...
-  border-radius: 1em;
+  border-radius: 2em;
   content: "";
-  transition: background 100ms ease-in-out;
   position: relative;
+  border: ${props => (props.checked ? border.on : border.off)};
+  transition: background 100ms ease-out, border-color 100ms ease-in;
+  outline: 1px solid transparent;
+  &:hover,
+  &:hover > * {
```
hamishtaplin: If `ToggleSwitchKnob` just inherits its border styles then no need for greedy `&:hover > * ` selector.

src/components/ToggleSwitch/index.js
```diff
@@ -55,34 +59,18 @@ const ToggleSwitchKnob = styled.div`
...
-    top: ${props => -((props.size * glowScale - props.size) / 2)}em;
-    will-change: transform;
-    transform: translateX(
-      ${props => -((props.size * glowScale - props.size) / 2)}em
-    );
-    border-radius: 50%;
-    opacity: 0;
-    transition: opacity 100ms ease-in-out;
-  }
+  border: ${props => (props.checked ? border.on : border.off)};
```
hamishtaplin: Use `border: inherit` to remove need for selector mentioned above.

src/components/ToggleSwitch/index.js
```diff
@@ -6,29 +6,27 @@ import { uniqueId } from "lodash";
...
 
-const glowScale = 1.5;
-
 export const HiddenInput = styled(Input)`
   opacity: 0;
   position: absolute;
 
-  &:focus {
-    border: none;
+  &:focus + div {
```
hamishtaplin: Better to be more specific about the element being selected here: 

```&:focus + ${ToggleSwitchBackground} ```

You'll have to move `ToggleSwitchBackground` above `HiddenInput` though.

**Comment by hamishtaplin** [comment]
@adammahmood @WickyNilliams @SamGodwin2 @samiwel 

Should we remove state from our stories as we go? Seems unnecessary to maintain that code every time we change a component. Recently, we've preferred just documenting each state separately and using `knobs` where appropriate.

File: src/components/ToggleSwitch/index.js [hunk 8]
```diff
@@ -95,7 +83,7 @@ ToggleSwitchKnob.propTypes = {
 };
 
 ToggleSwitchKnob.defaultProps = {
-  size: 1.4,
+  size: 1,
```
hamishtaplin: I think we can remove all concept of `size` for this now and rely on scaling with font-size if we need a bigger switch (we don't currently).

src/components/ToggleSwitch/index.js
```diff
@@ -6,29 +6,27 @@ import { uniqueId } from "lodash";
...
 
 const SharedDefaultProps = {
-  width: 2.6,
+  width: 1.5,
   height: 1
 };
 
-const glowScale = 1.5;
-
 export const HiddenInput = styled(Input)`
```
hamishtaplin: There isn't a relevant line for this comment but so I'll leave it here:

The hit box for clicking on this component needs to be at least `40px`, it's currently much too small.

adammahmood: `HiddenInput` has been updated to be 2 x size of the toggle switch.

WickyNilliams:
@hamishtaplin i think it's a good idea to do that, yeah

src/components/ToggleSwitch/index.js
hamishtaplin: I'd go to `3em` square.

Also, `cursor: pointer` for the whole thing so it's clearer when the click can happen.

PR #380, opened 2018-07-10
474 alert users when api is unreachable

K9Morphed:
Work to display error banner when API is unreachable, including differentiating between a network error and an API error.

### How to review 
Test should pass

If the API is down, the banner should display when an API call is made.

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,6 +114,7 @@
     "apollo-client": "^2.2.8",
     "apollo-client-preset": "^1.0.8",
     "apollo-link": "1.1.0",
+    "apollo-link-error": "^1.1.0",
     "apollo-link-http": "^1.5.3",
     "connected-react-router": "^4.2.1",
     "draft-convert": "^2.0.0",
--- a/src/apollo/createApolloErrorLink.js
+++ b/src/apollo/createApolloErrorLink.js
@@ -1,0 +1,9 @@
+import { onError } from "apollo-link-error";
+import { apiDownError } from "../redux/saving/actions";
+
+export default getStore =>
+  onError(({ networkError }) => {
+    if (networkError) {
+      getStore().dispatch(apiDownError());
+    }
+  });
--- a/src/components/OfflineBanner/index.js
+++ b/src/components/OfflineBanner/index.js
@@ -3,7 +3,7 @@ import PropTypes from "prop-types";
 import styled, { keyframes } from "styled-components";
 import { colors } from "constants/theme";
 import { connect } from "react-redux";
-import { isOnline } from "redux/saving/reducer";
+import { isOffline, hasApiError, hasError } from "redux/saving/reducer";
 import WarningIcon from "./icon-warning.svg?inline";
 import { TransitionGroup } from "react-transition-group";
 import ExpansionTransition from "components/ExpansionTransition";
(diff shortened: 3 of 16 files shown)


samiwel:
[Alert users when API is unreachable.](https://trello.com/c/SxzbzicZ/474-alert-users-when-api-is-unreachable)

src/apollo/createApolloErrorLink.js
```diff
@@ -0,0 +1,10 @@
+import { onError } from "apollo-link-error";
+import store from "../../src/index";
```
WickyNilliams: This should be passed in as an arg.

src/apollo/createApolloErrorLink.js
```diff
@@ -0,0 +1,10 @@
...
+
+export default onError(({ networkError }) => {
+  if (networkError) {
+    store.dispatch({
+      type: "API_DOWN_ERROR"
+    });
```
WickyNilliams: This action should be extracted into a function

src/index.js
```diff
@@ -31,17 +33,21 @@ const cache = createApolloCache({
...
+
+const client = createApolloClient(link, cache);
 
 const history = createHistory({
   basename: process.env.REACT_APP_BASE_NAME
 });
 
 const store = configureStore(history, client);
 
+export default store;
```
WickyNilliams: This should not be exported

src/index.js
```diff
@@ -31,17 +33,21 @@ const cache = createApolloCache({
   }
 });
 
-const client = createApolloClient(
-  createHttpLink(process.env.REACT_APP_API_URL),
-  cache
-);
+const link = ApolloLink.from([
+  createErrorLink,
```
WickyNilliams: `createErrorLink(store)` would allow you to avoid exporting the store in this module and importing it elsewhere. It also makes testing easier

src/redux/saving/reducer.js
```diff
@@ -37,10 +39,19 @@ export default (state = initialState, action) => {
...
     default:
       return state;
   }
 };
 
 export const isSaving = state => get(state, "saving.pendingRequestCount") > 0;
 export const isOnline = state => get(state, "saving.online");
+export const hasApiError = state => get(state, "saving.apiDownError");
+export const hasError = state =>
+  get(state, "saving.apiDownError") || !get(state, "saving.online");
```
WickyNilliams: You could call the two functions above, instead of duplicating their functionality

src/apollo/createApolloErrorLink.js
WickyNilliams: Oh, it seems you already have a function `apiDownError` in `saving/actions.js`. Use that instead:

```js
store.dispatch(apiDownError());
```

src/components/OfflineBanner/index.js
```diff
@@ -41,15 +41,21 @@ const WarningMessage = styled(IconText)`
...
+    "We're unable to save your progress right now. Try refreshing the page or try again shortly",
+  isOffline:
+    "You're currently offline and any changes you make won't be saved. Check your connection and try again."
+};
+
 export const UnconnectedOfflineBanner = props => {
   return (
     <TransitionGroup>
-      {!props.isOnline && (
+      {(!props.isOnline || props.apiError) && (
```
WickyNilliams: if you added `hasError` to `mapStateFromProps`, you could just say `props.hasError` here?

package.json
```diff
@@ -111,6 +111,7 @@
     "apollo-client": "^2.2.8",
     "apollo-client-preset": "^1.0.8",
     "apollo-link": "1.1.0",
+    "apollo-link-error": "^1.1.0",
     "apollo-link-http": "^1.5.3",
```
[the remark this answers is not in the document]
WickyNilliams: I looked into it a little more, and it seems it only supports the simplest use cases, and won't support things like fragment matchers, which we require.

src/index.js
```diff
@@ -31,17 +33,20 @@ const cache = createApolloCache({
...
-const client = createApolloClient(
-  createHttpLink(process.env.REACT_APP_API_URL),
-  cache
-);
-
 const history = createHistory({
   basename: process.env.REACT_APP_BASE_NAME
 });
 
+// eslint-disable-next-line no-use-before-define
```
WickyNilliams: Hmm... this isn't a good idea. This linting rule is there for a reason. If we ever started using native `const` (i.e. turned off the babel compilation of `const` -> `var`), this would blow up. 

I'm not sure what to do here as there's a circular dependency: 

```
+--> Link ------> Store ---+
|                          |
|                          |
+--------- Client <--------+
```

Perhaps you could pass in a function to `createErrorLink` which returns the store, which allows you to sort of defer the value:

```js

let store;
const getStore = () => store;

const link = ApolloLink.from([
  createErrorLink(getStore),
  createHttpLink(process.env.REACT_APP_API_URL)
]);

const client = createApolloClient(link, cache);
store = configureStore(history, client);

const renderApp = render(document.getElementById("root"), {
  store,
  client,
  history
});
```

samiwel:
LGTM. Let's squash down the commits before we merge.

PR #381, opened 2018-07-12
503 piping broken

WickyNilliams:
Discovered that piping was being removed when formatting is stripped. This PR whitelists piping as allowed formatting.

### How to review 
Tests pass, can pipe values

### DIFF
--- a/src/components/RichTextEditor/index.js
+++ b/src/components/RichTextEditor/index.js
@@ -19,8 +19,7 @@ import PipedValueDecorator, {
   insertPipedValue
 } from "./entities/PipedValue";
 
-import { filterEditorState } from "draftjs-filters";
-import { mapControlsToFilterConfig } from "./utils/mapControlsToFilterConfig";
+import createFormatStripper from "./utils/createFormatStripper";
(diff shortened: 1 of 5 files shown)


WickyNilliams:
[Bug: piping not working](https://trello.com/c/rPTCbJBA/503-bug-piping-not-working)

src/components/RichTextEditor/index.js
```diff
@@ -240,11 +239,8 @@ class RichTextEditor extends React.Component {
   };
 
   handleChange = (editorState, callback) => {
-    let filteredState = editorState;
-    if (this.props.controls) {
-      filteredState = filterEditorState(this.filterConfig, filteredState);
-    }
-    return this.setState({ editorState: filteredState }, callback);
+    editorState = this.stripFormatting(editorState);
```
samiwel: Does this need to mutate the input variable?

WickyNilliams: it reassigns, but doesn't mutate (all draft-js stuff is immutable)

PR #385, opened 2018-07-18
Prevented description from showing for some answers

SamGodwin2:
The Description for TextField, TextArea and Number answers weren't showing in Runner as the Runner templates do not currently support these fields. This Pr disables description inputs for those answers to prevent user confusion.

### How to review 
Tests pass and Description inputs should only be visible on the following answer types: Currency, Radio and Checkbox.

### DIFF
--- a/cypress/integration/author_spec.js
+++ b/cypress/integration/author_spec.js
@@ -9,7 +9,7 @@ import {
   addSection,
   addQuestionPage
 } from "../utils";
-import { times } from "lodash";
+import { times, includes } from "lodash";
 import { Routes } from "../../src/utils/UrlUtils";
 
 describe("eq-author", () => {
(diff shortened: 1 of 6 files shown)


cypress/integration/author_spec.js
```diff
@@ -256,9 +256,11 @@ describe("eq-author", () => {
     answerTypes.forEach(answerType => {
       addAnswerType(answerType);
       cy.get("[data-test='txt-answer-label']").type(answerType + " label");
-      cy
-        .get("[data-test='txt-answer-description']")
-        .type(answerType + " description");
+      if (includes(answerType, ["Currency"])) {
```
samiwel: Not sure that we need to use includes here

PR #386, opened 2018-07-19
Piping date ranges v2

SamGodwin2:
This PR includes the front-end changes required to support Piping on both the from and to date range answers. 

### How to review 
Tests pass and you are able to pipe a date range answer from a previous question.

### Dependant on:

https://github.com/ONSdigital/eq-author-graphql-schema/pull/40
https://github.com/ONSdigital/eq-publisher/pull/65
https://github.com/ONSdigital/eq-author-api/pull/78


SamGodwin2:
[Piping Date Ranges (5)](https://trello.com/c/pjFZp9OC/315-piping-date-ranges-5)

samiwel:
[Piping Date Ranges - Frontend changes (5)](https://trello.com/c/pjFZp9OC/315-piping-date-ranges-frontend-changes-5)

SamGodwin2:
@adammahmood Agreed the functional method does look cleaner however I think it would mean needing to bring in map and filter from lodash/fp? Trying to avoid using both lodash and lodash/fp in the same file as it has caused confusion in the past.

Could still be possible by currying the functions which results in something like this: 
`const getAnswer = answer => {
  if (answer.__typename === "CompositeAnswer") {
    return answer.childAnswers;
  }
  return answer;
};

const isValidType = ({ type }) => validAnswerTypes[type];

const curriedMap = curry(map);

const curriedFilter = curry(filter);

const filterAnswers = flow(
  curriedMap(curry.placeholder, getAnswer),
  flatten,
  curriedFilter(curry.placeholder, isValidType)
);`

What do you think still worth it or am I missing something somewhere?

PR #387, opened 2018-07-23
Fix broken firebase login

SamGodwin2:
The firebase ui behind the log in screen was experiencing some issues so this PR aims to fix that.

### How to review 
Check on netlify that you can log out and then log back in again.


samiwel:
[Sign-in bug](https://trello.com/c/Lbo1RIXX/517-sign-in-bug)

PR #389, opened 2018-08-01
Improve functional tests

samiwel:
This PR aims to make a number of improvements to the Author functional tests.
There will be some more work to be done off the back of this PR, but hopefully this is a step in the right direction.

Some of the changes made, include:

- Where possible remove reliance on built up state.
- Make it possible to run tests in isolation.
- Allow programmatic login from cypress tests.
- Begin mocking GraphQL responses.
- Split spec files out into different screens.

### How to review 
Should be run any of the tests in isolation and they should pass.


samiwel:
[Investigate how to improve functional tests](https://trello.com/c/EB74AU7B/498-investigate-how-to-improve-functional-tests)

cypress/integration/auth/login_spec.js
```diff
@@ -0,0 +1,49 @@
...
+
+  it("should prompt user for password if account exists", () => {
+    cy.get(firebase).within(() => {
+      cy
+        .get(email)
+        .clear()
+        .type("cypressTest@ons.com{enter}");
+      cy.get(password).should("be.visible");
+    });
+  });
```
SamGodwin2: Do we want a test here to follow all the way though and actually login through the firebase ui rather than just though the store?

cypress/integration/authenticated/builder_spec.js
```diff
@@ -142,29 +139,31 @@ describe("eq-author", () => {
...
-          currentPath: Routes.SECTION,
-          equality: {
-            questionnaireId: true,
-            sectionId: false
-          }
-        });
-      });
+      .click();
+
+    cy.hash().should("not.eq", resultingHash);
```
SamGodwin2: Would it be worth running some regex on the hash to check that the hash is defiantly a section design page? Just thinking that this test only really checks that we've gone to a new page and doesn't necessarily check its the right one.

SamGodwin2:
Just a couple minor questions.

cypress/integration/auth/login_spec.js
samiwel: I'd say probably not as part of the functional tests. The real firebase auth is tested as part of the smoke test. But your comment got me thinking, I think I need to introduce another env var to make sure that store is only exposed to Cypress during functional tests and not the smoke test.

e.g. 

```js
if (window.Cypress && REACT_APP_IS_FUNCTIONAL_TEST === "true") {
  // make the store available
}
```

cypress/integration/authenticated/dashboard_spec.js
```diff
@@ -0,0 +1,113 @@
...
+    cy.get(testId("create-questionnaire")).click();
+    cy
+      .get(testId("questionnaire-settings-modal", "testid"))
+      .should("be.visible");
+  });
+
+  describe("empty state", () => {
+    beforeEach(() => {
+      cy.visitStubbed("/", {
+        GetQuestionnaireList: {
```
hamishtaplin: Awesome we can do this now 👍

cypress/integration/authenticated/dashboard_spec.js
```diff
@@ -0,0 +1,113 @@
...
+              undeleteQuestionnaire: { id: "1", __typename: "Questionnaire" }
+            }
+          }
+        });
+
+        cy.login();
+      });
+
+      it("should delete the questionnaire", () => {
+        cy.get("tr button[type='button']").click();
```
hamishtaplin: Can we use some more robust selectors for this?

cypress/integration/authenticated/dashboard_spec.js
```diff
@@ -0,0 +1,113 @@
...
+        cy.contains("You have no questionnaires").should("be.visible");
+      });
+
+      it("should display toast when questionnaire is deleted", () => {
+        cy.get("tr button[type='button']").click();
+        cy.contains("Undo").should("be.visible");
```
hamishtaplin: Should this be scoped to a more specific place? Adding anything else with the word "Undo" in would make this test pass. Pretty sure we already had a situation like this elsewhere.

cypress/integration/authenticated/dashboard_spec.js
```diff
@@ -0,0 +1,113 @@
...
+      });
+
+      it("should restore questionnaire when undo is clicked", () => {
+        cy.get("tr button[type='button']").click();
+        cy.contains("Undo").click();
+        cy.contains("Test Questionnaire").should("be.visible");
```
hamishtaplin: Again this might be to broad.

cypress/integration/authenticated/settings_spec.js
```diff
@@ -0,0 +1,136 @@
+import { testId, setQuestionnaireSettings } from "../../utils";
+
+const stubbedGraphQLResponses = {
```
hamishtaplin: This all looks pretty generic. Can we extract these into some files for re-use? eg. could export each `createQuestionnaire`, `GetQuestionPage`, `GetQuestionnaire`, etc.

cypress/support/commands.js
```diff
@@ -1,8 +1,34 @@
 import { setQuestionnaireSettings, testId } from "../utils";
 
-Cypress.Commands.add("login", () => {
-  cy.visit("/");
-  cy.contains("Sign in as Guest").click();
+Cypress.Commands.add("login", options => {
+  const payload = Object.assign(
```
hamishtaplin: Use spread operator?

samiwel: I tried using the spread operator here but it didn't work. I think it might have something to do with being run by cypress vs being run as a normal node script. AFAIK the commands file just gets loaded as a script tag in the browser when cypress begins.

PR #391, opened 2018-08-03
Typeahead component

hamishtaplin:

[Trello card](https://trello.com/c/POnn9CQN/483-survey-data-typeahead-component)

Adds `Typeahead` component that takes a list of objects with values and handles the filtering of those values according to user input as well as the dropdown menu displaying options for selection.

screenshot 2018-08-06 13 57 33

This component uses [Downshift](https://github.com/paypal/downshift) and exposes two components: 

- `Typeahead` exposes a Downshift component and simply handles mapping `onChange` and `itemToString` props, the latter handles the mapping of `value` of items into strings.
- `TypeaheadMenu` contains logic for filtering menu items by user input and all the styling associated with the menu and items.

A11y concerns are handled by Downshift itself.

### How to review 

- Check component works in storybook
- Tests pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,6 +117,7 @@
     "apollo-link-error": "^1.1.0",
     "apollo-link-http": "^1.5.3",
     "connected-react-router": "^4.2.1",
+    "downshift": "^2.0.20",
     "draft-convert": "^2.0.0",
     "draft-js": "^0.10.5",
     "draft-js-raw-content-state": "^1.2.5",
--- a/src/components/Forms/Input/index.js
+++ b/src/components/Forms/Input/index.js
@@ -41,7 +41,12 @@ const StyledInput = styled.input`
   ${props => props.type === "radio" && radioButton};
 `;
 
-const Input = ({ type, defaultValue, id, ...otherProps }) => (
+export const UncontrolledInput = ({
+  type,
+  defaultValue,
+  id,
+  ...otherProps
+}) => (
   <StyledInput
     type={type}
     defaultValue={defaultValue}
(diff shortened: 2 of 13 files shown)


package.json
```diff
@@ -117,6 +117,7 @@
     "apollo-link-error": "^1.1.0",
     "apollo-link-http": "^1.5.3",
     "connected-react-router": "^4.2.1",
+    "downshift": "^2.0.20",
```
adammahmood: This looks like a choice.

src/components/Forms/Input/index.js
```diff
@@ -41,7 +41,12 @@ const StyledInput = styled.input`
   ${props => props.type === "radio" && radioButton};
 `;
 
-const Input = ({ type, defaultValue, id, ...otherProps }) => (
+export const UncontrolledInput = ({
```
adammahmood: Why does this have to be renamed?

package.json
hamishtaplin: What do you mean?

src/components/Forms/Input/index.js
hamishtaplin: Because I needed to export the component without the `withChangeHandler` enhancement. We do this in various places but not always consistent about the name.

package.json
adammahmood: Downshift looks like a good package. (Edited comment above).

hamishtaplin: Ah right 👍 

Yeah I think we can use for a bunch of things, we could refactor any existing menus to use it even if we don't have the typehead functionality

PR #395, opened 2018-08-09
Add DataTable component

adammahmood:
This is part of a wider piece of work to capture 'Survey Data'.  The UI will require a Data Table which this PR implements. 

### How to review 
- Run Tests
- View Storybook

### DIFF
--- a/src/components/DataTable/story.js
+++ b/src/components/DataTable/story.js
@@ -1,0 +1,133 @@
+import React from "react";
+import styled from "styled-components";
+import { storiesOf } from "@storybook/react";
+import { action } from "@storybook/addon-actions";
+import { map, size, times } from "lodash";
+
+import IconText from "components/IconText";
+import Typeahead from "components/Typeahead";
+import {
+  Table,
+  TableHead,
+  TableBody,
+  TableFoot,
+  TableRow,
+  TableColumn,
+  TableHeadColumn
+} from "components/DataTable/Elements";
+import {
+  AddRowButton,
+  TableInput,
+  TableInputDate,
+  TableSelect,
+  TableTypeaheadInput,
+  TableTypeaheadMenu,
+  DeleteRowButton
+} from "components/DataTable/Controls";
+
+import Icon from "./icon-plus.svg?inline";
+
+const Wrapper = styled.div`
+  position: absolute;
+  padding: 2em;
+`;
+
+const input = (
+  <TableInput id="table-input" type="text" onChange={action("change")} />
+);
+
+const date = (
+  <TableInputDate id="table-date" type="date" onChange={action("change")} />
+);
+
+const select = selected => (
+  <TableSelect value={selected} onChange={action("change")} id="options">
+    <option value="1">Option 1</option>
+    <option value="2">Option 2</option>
+    <option value="3">Option 3</option>
+  </TableSelect>
+);
+const items = [{ value: "apple" }, { value: "pear" }, { value: "orange" }];
+
+const typeahead = (
+  <Typeahead onChange={action("change")}>
+    {({ getInputProps, getLabelProps, isOpen, openMenu, ...otherProps }) => (
+      <div>
+        <TableTypeaheadInput {...getInputProps({ onFocus: openMenu })} />
+        {isOpen && <TableTypeaheadMenu items={items} {...otherProps} />}
+      </div>
+    )}
+  </Typeahead>
+);
+
+const header = [
+  {
+    name: "Heading 1",
+    width: 15
+  },
+  {
+    name: "Heading 2",
+    width: 20
+  },
+  {
+    name: "Heading 3",
+    width: 20
+  },
+  {
+    name: "Heading 4",
+    width: 40
+  }
+];
+
+const row = () => [typeahead, input, select(1), date];
+
+const content = times(4, row);
+
+const Decorator = storyFn => <Wrapper>{storyFn()}</Wrapper>;
+
+storiesOf("DataTable", module)
+  .addDecorator(Decorator)
+  .add("With dynamic content", () => (
+    <Table>
+      <TableHead>
+        <TableRow>
+          {map(header, ({ name, width }, index) => (
+            <TableHeadColumn key={`dt-th-${index}`} width={`${width}%`}>
+              {name}
+            </TableHeadColumn>
+          ))}
+          <TableHeadColumn width="36px" />
+        </TableRow>
+      </TableHead>
+      <TableBody>
+        <React.Fragment>
+          {map(content, (row, rowIndex) => {
+            return (
+              <TableRow key={`dt-row-${rowIndex}`}>
+                {map(row, (col, colIndex) => (
+                  <TableColumn key={`dt-col-${colIndex}`}>{col}</TableColumn>
+                ))}
+                <TableColumn>
+                  <DeleteRowButton
+                    size="medium"
+                    onClick={action("deleteRow")}
+                  />
+                </TableColumn>
+              </TableRow>
+            );
+          })}
+        </React.Fragment>
+      </TableBody>
+      <TableFoot>
+        <TableRow>
+          <TableColumn colSpan={size(header) + 1}>
+            <AddRowButton onClick={action("addRow")} variant="tertiary" small>
+              <IconText icon={Icon} dark>
+                Add row
+              </IconText>
+            </AddRowButton>
+          </TableColumn>
+        </TableRow>
+      </TableFoot>
+    </Table>
+  ));
--- a/src/components/AnswerEditor/index.js
+++ b/src/components/AnswerEditor/index.js
@@ -36,7 +36,7 @@ const Answer = styled.div`
 `;
 
 const AnswerType = styled.div`
-  background: #e4e8eb;
+  background: ${colors.lightMediumGrey};
   border-bottom: 1px solid ${colors.bordersLight};
   text-align: center;
   padding: 0.5em 1em;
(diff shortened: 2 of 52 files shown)


samiwel:
[Survey Data - Datatable](https://trello.com/c/3vedxb1M/482-survey-data-datatable)

src/components/DataTable/index.js
```diff
@@ -0,0 +1,154 @@
...
+    const { header } = this.props;
+
+    return (
+      <tr>
+        {map(header, ({ name, width }, index) => (
+          <th key={`dt-th-${index}`} width={`${width}%`}>
+            {name}
+          </th>
+        ))}
+        <th width="36px" />
```
samiwel: Should we be hard coding px values into front end components?

src/components/DataTable/story.js
```diff
@@ -0,0 +1,90 @@
...
+    name: "Heading 4",
+    width: 40
+  }
+];
+
+storiesOf("DataTable", module)
+  .add("No content", () => (
+    <DataTable
+      header={header}
+      content={[]}
```
samiwel: What's your opinion about passing in arrays as props vs making it composable? e.g.
I guess on the one hand the DataTable takes care of rendering the arrays, but there's more flexibility in the latter approach.

```
<DataTable>
  <THead>
    <Tr>
      <Th>Key</Th>
    </Tr>
  </Thead>
  <Tbody>
    {map(metadata, row => (
       <Tr>
			...
       </Tr>
     )}
  </Tbody>
</DataTable>
```

samiwel: @hamishtaplin may have some thoughts?

src/components/DataTable/index.js
```diff
@@ -0,0 +1,154 @@
...
+  table-layout: fixed;
+  text-align: left;
+  border: var(--border-style-dark);
+`;
+
+const Thead = styled.thead`
+  background-color: var(--header-background-color);
+  border-bottom: var(--border-style-dark);
+  color: ${colors.textLight};
+  tr > th {
```
hamishtaplin: Try to avoid using element selectors if possible, they're fragile and extending the styles gets you into a specificity arms-race. Styled-components makes this much easier (eg. create a `TableHeading` component with the padding styles`).

Also the styles of the `th` are not a concern of the `Thead` component.

src/components/DataTable/index.js
```diff
@@ -0,0 +1,154 @@
...
+    padding: 0.75em 1em;
+  }
+  tr > th + th {
```
hamishtaplin: Same here with `th` etc. This can be solved in a `TableHeading` component and using either a specific prop for no-border or `:not(:last-of-type)`

src/components/DataTable/index.js
```diff
@@ -0,0 +1,154 @@
...
+    border-left: var(--border-style-light);
+  }
+`;
+
+const Tbody = styled.tbody`
+  tr {
```
hamishtaplin: `TableRow` component

src/components/DataTable/index.js
```diff
@@ -0,0 +1,154 @@
...
+  --border-style-dark: 1px solid ${colors.borders};
+  width: 100%;
+  font-size: 1em;
+  border-collapse: collapse;
+  table-layout: fixed;
+  text-align: left;
+  border: var(--border-style-dark);
+`;
+
+const Thead = styled.thead`
```
hamishtaplin: I think `TableHead` is a better name as you can then use that convention for other components, eg `TableRow`. Could argue that we don't need to scope the names with `Table` at all though as it's a bit redundant.

src/components/DataTable/index.js
```diff
@@ -0,0 +1,154 @@
...
+    border-left: var(--border-style-light);
+  }
+`;
+
+const Tbody = styled.tbody`
+  tr {
+    border-bottom: var(--border-style-light);
+  }
+  tr > td + td {
+    border-left: var(--border-style-light);
```
hamishtaplin: Same as above.

src/components/DataTable/index.js
hamishtaplin: Probably not. The problem with tables though (as seen on the questionnaires table) is that they are so dependent on their content for the layout of the whole thing then they tend to jump around when you add/remove stuff so can be a bit weird if left to their own devices.

I think opening this up for composition would help though instead of baking these values into this component.

src/components/DataTable/story.js
hamishtaplin: Agree with @samiwel here, this should export components for later composition. This makes it more flexible when requirements change or we need to re-use things.

src/components/DataTable/DataTableColumn.js
```diff
@@ -0,0 +1,21 @@
...
+
+import { colors } from "constants/theme";
+
+const Column = styled.td`
+  :not(:last-of-type) {
+    border-right: 1px solid ${colors.bordersLight};
+  }
+`;
+
+const DataTableColumn = ({ children, ...otherProps }) => (
```
samiwel: Not sure we gain a lot by defining a separate `DataTableColumn` component? I think styled components will pass down props and render children anyway, so you can probably just export Column? Same for the other intermediate components for DataTableHead, DataTableFooter etc..

src/components/DataTable/DataTableFoot.js
```diff
@@ -0,0 +1,19 @@
...
+import PropTypes from "prop-types";
+import styled from "styled-components";
+
+import { colors } from "constants/theme";
+
+const TableFoot = styled.tfoot`
+  border-top: 1px solid ${colors.borders};
+`;
+
+const DataTableFoot = ({ children, ...otherProps }) => (
```
samiwel: Don't think we need this.

src/components/DataTable/DataTableHead.js
```diff
@@ -0,0 +1,21 @@
...
+
+import { colors } from "constants/theme";
+
+const TableHead = styled.thead`
+  background-color: ${colors.lightMediumGrey};
+  border-bottom: 1px solid ${colors.borders};
+  color: ${colors.textLight};
+`;
+
+const DataTableHead = ({ children, ...otherProps }) => (
```
samiwel: Don't think we need this.

src/components/DataTable/DataTableHeadColumn.js
```diff
@@ -0,0 +1,22 @@
...
+import { colors } from "constants/theme";
+
+const Column = styled.th`
+  padding: 0.75em 1em;
+  :not(:last-of-type) {
+    border-right: 1px solid ${colors.bordersLight};
+  }
+`;
+
+const DataTableHeadColumn = ({ children, ...otherProps }) => (
```
samiwel: Don't think we need this.

src/components/DataTable/DataTableRow.js
```diff
@@ -0,0 +1,21 @@
...
+
+import { colors } from "constants/theme";
+
+const Row = styled.tr`
+  :not(:last-of-type) {
+    border-bottom: 1px solid ${colors.bordersLight};
+  }
+`;
+
+const DataTableRow = ({ children, ...otherProps }) => (
```
samiwel: Don't think we need this.

src/components/DataTable/index.js
```diff
@@ -0,0 +1,24 @@
...
+const Table = styled.table`
+  width: 100%;
+  font-size: 1em;
+  border-collapse: collapse;
+  table-layout: fixed;
+  text-align: left;
+  border: 1px solid ${colors.borders};
+`;
+
+const DataTable = ({ children, ...otherProps }) => (
```
samiwel: Don't think we need this.

src/components/DataTable/story.js
```diff
@@ -0,0 +1,144 @@
...
+          >
+            <IconText icon={Icon} dark>
+              Add row
+            </IconText>
+          </AddRowButton>
+        </DataTableColumn>
+      </DataTableRow>
+    </DataTableFoot>
+  </DataTable>
+));
```
samiwel: I think this is better now that it can be composed in different ways 👍

src/components/DataTable/story.js
```diff
@@ -0,0 +1,144 @@
...
+import DeleteButton from "components/DeleteButton";
+import Button from "components/Button";
+import Typeahead from "components/Typeahead";
+import TypeaheadMenu from "components/Typeahead/TypeaheadMenu";
+
+import Icon from "./icon-plus.svg?inline";
+
+const TableInput = styled(Input)`
+  border-radius: 0;
+  border-color: transparent;
```
hamishtaplin: Need to override the focus style so it's inside the `Input`

```
&:focus {
    box-shadow: inset 0 0 0 3px ${colors.tertiary};
}
```

src/components/DataTable/story.js
```diff
@@ -0,0 +1,144 @@
...
+import IconText from "components/IconText";
+import { Input, Select } from "components/Forms";
+import DeleteButton from "components/DeleteButton";
+import Button from "components/Button";
+import Typeahead from "components/Typeahead";
+import TypeaheadMenu from "components/Typeahead/TypeaheadMenu";
+
+import Icon from "./icon-plus.svg?inline";
+
+const TableInput = styled(Input)`
```
hamishtaplin: I think it's better to move all these styled components into file/s so they can be re-used.

src/components/DataTable/story.js
```diff
@@ -0,0 +1,144 @@
...
+            <DataTableRow key={`dt-row-${rowIndex}`}>
+              {map(row, (col, colIndex) => (
+                <DataTableColumn key={`dt-col-${colIndex}`}>
+                  {col}
+                </DataTableColumn>
+              ))}
+              <DataTableColumn>
+                <DeleteRowButton
+                  data-test="delete-row-btn"
+                  size={"medium"}
```
hamishtaplin: linter should complain on this prop being a string inside curly braces

PR #396, opened 2018-08-13
Replace logic that strips out HTML tags with API calls for plaintext values.

samiwel:
This PR addresses some technical debt by removing logic that strips out HTML tags from section and page titles in the front end and replace it with a query for a field on the GraphQL API that retrieves plaintext representations (using the `format: Plaintext` field argument).

The aim going forward is for the API to be the single source of truth for the data and to remove and to keep the logic for stripping out HTML tags in one place. 

### How to review 
Plaintext titles are displayed in the navigation, move page/section modals and the piping menu. A visual inspection should be carried out to ensure that titles are displayed correctly.

All tests and checks should pass and the changes should look sensible.


adammahmood:
LGTM.  However, looks like Cypress tests are failing.

src/components/RichTextEditor/Piping.graphql
```diff
@@ -4,10 +4,12 @@ query GetQuestionnaire_Piping($id: ID!) {
     sections {
       id
       title
+      plaintextTitle: title(format: Plaintext)
```
SamGodwin2: Do we still need to request title here or can we remove it?

src/graphql/getQuestionnaire.graphql
```diff
@@ -6,6 +6,7 @@ query GetQuestionnaire($id: ID!) {
     sections {
       id
       title
+      plaintextTitle: title(format: Plaintext)
```
SamGodwin2: Same again here

src/components/RichTextEditor/Piping.graphql
samiwel: Good shout

PR #399, opened 2018-08-16
Add date format property

adammahmood:
Implements functionality required to specify the date format of a date answer. 

### How to review 
- Run Tests
- Ensure feature works as per https://trello.com/c/bsxNpqSy/461-date-format-picker-ui-api

### Dependencies
API: https://github.com/ONSdigital/eq-author-api/pull/105


src/components/withEntityEditor/index.js
```diff
@@ -25,7 +22,7 @@ const withEntityEditor = (entityPropName, fragment) => WrappedComponent => {
     state = {};
 
     static getDerivedStateFromProps(nextProps, prevState) {
-      if (entityIsUnset(prevState) || entityHasChanged(nextProps, prevState)) {
+      if (entityIsUnset(prevState) || !isEqual(nextProps, prevState)) {
```
adammahmood: @SamGodwin2 @samiwel @hamishtaplin - Does anyone know why the `entityHasChanged` method was only comparing IDs rather than the object contents from `nextProps` and `prevState`? 

I've had to update this method so it does a deep comparison on the `nextProps` and `prevState` so the Answer components will update with the most recent values from the Apollo Cache if a nested property has been updated. 

This was necessary to render updates from the Properties Panel in the Answer Editor.

src/components/withEntityEditor/index.test.js
```diff
@@ -102,7 +102,7 @@ describe("withEntityEditor", () => {
     expect(wrapper.state("entity")).toEqual(newEntity);
   });
 
-  it("should not update state when new entity with same id passed via props", () => {
+  it("should update state when new entity with same id passed via props", () => {
```
adammahmood: Please comment on src/components/withEntityEditor/index.js  to understand why this test has changed.

src/components/withEntityEditor/index.js
adammahmood: The only reason I can think this only checked on ID's was to prevent unnecessary renders.

adammahmood:
I've created a separate task to update the width of the navigation sidebar and properties panel as it is beyond the scope of this task.

https://trello.com/c/bsxNpqSy/461-increase-width-of-navigationsidebar-and-propertiespanel

adammahmood:
[Date format picker - UI / API](https://trello.com/c/yEk9c98U/562-date-format-picker-ui-api)

adammahmood:
[Date format picker - UI / API](https://trello.com/c/yEk9c98U/562-date-format-picker-ui-api)

src/components/withEntityEditor/index.js
adammahmood: This article explains `getDerivedStateFromProps` further: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html

src/components/AnswerProperties/Labels/InlineLabel.js
```diff
@@ -0,0 +1,30 @@
...
+const InlineField = styled(Field)`
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+  margin-bottom: 0;
+`;
+
+const InlineLabel = ({ id, label, children }) => (
+  <InlineField key={id}>
```
hamishtaplin: This doesn't feel like it should be part of this component as we separate `Field` and its children in other implementations. I think the abstraction here is the `InlineField` rather than `InlineLabel`? Particularly as the `Label` itself doesn't encapsulate any new styles.

src/components/AnswerProperties/Labels/MultiLineLabel.js
```diff
@@ -0,0 +1,31 @@
...
+const MultiLineField = styled(Field)`
+  padding: 0.2em 0;
+  margin-bottom: 0;
+`;
+
+const MultiLineControl = styled.div`
+  margin-top: 0.4em;
+`;
+
+const MultiLineLabel = ({ id, label, children }) => (
```
hamishtaplin: Same comment as the `InlineLabel`, I think the `MultiLineField ` is the key component here.

src/components/AnswerProperties/Properties/Decimal.js
```diff
@@ -0,0 +1,22 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+import { Number } from "components/Forms";
+
+const Decimal = ({ id, value, handleChange }) => (
```
hamishtaplin: Is this worth extracting to a component? It doesn't really encapsulate anything beyond the `max` value.

src/components/AnswerProperties/Properties/Required.js
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+import ToggleSwitch from "components/ToggleSwitch";
+
+const Required = ({ id, value, handleChange }) => (
```
hamishtaplin: Same as previous comment, not sure it's worth creating this component.

src/components/withEntityEditor/index.js
hamishtaplin: @adammahmood Not sure, probably just an oversight.

src/components/AnswerProperties/Properties/Decimal.js
adammahmood: All of the properties have been extracted to their own components so they can be reused in this method: 

https://github.com/ONSdigital/eq-author/blob/3229fb9b4290e150e91caa85d7756572badf02c7/src/components/AnswerProperties/index.js#L39

src/components/AnswerProperties/Labels/InlineLabel.js
adammahmood: Yes. I think you're correct.  These are "Fields" rather than "Labels".  I'll rename. I think it's useful to keep the Label within this component to reduce repetitiveness in:

https://github.com/ONSdigital/eq-author/blob/3229fb9b4290e150e91caa85d7756572badf02c7/src/components/AnswerProperties/index.js#L39

src/components/AnswerProperties/Properties/Decimal.js
hamishtaplin: Fair enough 👍

PR #400, opened 2018-08-17
Implement routing to a new section

SamGodwin2:
This Pr re-adds the logic behind adding options to the selector of pages that are outside the current section.

### How to review 
You can route between sections, tests pass.

### Dependencies:
https://github.com/ONSdigital/eq-publisher/pull/85


SamGodwin2:
[Route to a new section - Front End changes](https://trello.com/c/8CEIfTiY/560-route-to-a-new-section-front-end-changes)

src/components/routing/RoutingRuleDestinationSelector.js
```diff
@@ -47,14 +47,33 @@ const toOption = defaultTitle => entity => ({
...
 
-    return [pagesGroup].filter(group => group.options.length);
+    const sectionGroup = {
+      label: "Sections in this questionnaire",
+      id: "sections",
+      options: sections.map(toOption("Section Title"))
+    };
+
+    const logicalGroup = {
+      label: "Other destinations",
```
hamishtaplin: Change to "End of Questionnaire", I think "Other destinations" implies there could be several destinations here but there's only going to be one.

Issue #397, opened 2018-08-14
Error message in console when creating a new Questionnaire


hamishtaplin:
### Expected behaviour

Creating a Questionnaire should happen without error

### Actual behaviour

When creating a Questionnaire an error is logged to the browser console:

<img width="834" alt="screenshot 2018-08-14 13 49 17" src="https://user-images.githubusercontent.com/930398/44092648-0273551e-9fc9-11e8-8a33-44576565bf76.png">

### Steps to reproduce the behaviour

- Login
- Create a Questionnaire
- Error is logged when app forwards you to the new Questionnaire

#### Browser

Chrome

#### Operating System

macOS

adammahmood:
https://github.com/ONSdigital/eq-author/pull/409

adammahmood:
Please close this issue once merged: https://github.com/ONSdigital/eq-author/issues/397

PR #409, opened 2018-08-23
Fix setState being called on unmounted component. Issue #397

adammahmood:
- Fixes an error caused by setting the state on unmounted components which indicates a memory leak
- Fixes https://github.com/ONSdigital/eq-author/issues/397
- We suspect this was contributing to our Cypress tests becoming slow, unpredictable and flakey

### How to review 
- Run Tests
- Run Cypress tests, open browser console and ensure there are no errors
- Use steps described in https://github.com/ONSdigital/eq-author/issues/397

### DIFF
--- a/src/components/RichTextEditor/index.js
+++ b/src/components/RichTextEditor/index.js
@@ -166,6 +166,10 @@ class RichTextEditor extends React.Component {
     }
   }
 
+  componentWillUnmount() {
+    this.unmounted = true;
+  }
+
   updatePipedValues() {
     const { editorState } = this.state;
     const { fetchAnswers } = this.props;
@@ -263,7 +267,7 @@ class RichTextEditor extends React.Component {
     });
 
     this.timeoutID = setTimeout(() => {
-      if (this.state.focused) {
+      if (!this.unmounted && this.state.focused) {
         this.setState({ focused: false });
       }
     }, 0);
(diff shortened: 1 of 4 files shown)


adammahmood:
[Fix error caused when creating a new questionnaire](https://trello.com/c/Nl66xnAK/572-fix-error-caused-when-creating-a-new-questionnaire)

src/components/RichTextEditor/index.js
```diff
@@ -264,7 +268,9 @@ class RichTextEditor extends React.Component {
 
     this.timeoutID = setTimeout(() => {
       if (this.state.focused) {
-        this.setState({ focused: false });
+        if (!this.unmounted) {
```
samiwel: Can avoid nested `if` statements by `&&` the operands. Also wonder if it's worth short-circuiting the check by evaluating the `unmounted` value first?

e.g.

```javascript
if (!this.unmounted && this.state.focused) {
 ...
}
```

src/components/RichTextEditor/index.js
```diff
@@ -273,7 +279,9 @@ class RichTextEditor extends React.Component {
     clearTimeout(this.timeoutID);
 
     if (!this.state.focused) {
-      this.setState({ focused: true });
+      if (!this.unmounted) {
```
samiwel: Same as above.

samiwel:
Looks good just some minor comments.

src/components/RichTextEditor/index.js
adammahmood: I think short-circuiting would be the better option.  I wanted to make it as clear as possible that the state should not be set if the component is unmounted.

adammahmood: Although, doesn't look as neat.  Will use the suggested approach.

PR #417, opened 2018-08-31
Spike - stop importing schema

tgandrews:
This is a change that attempts to remove the need for a published npm package containing the schema by having the UI query a running API for all the schema data it needs for linting and fragment matching.

### How to review 
This is should not be reviewed properly as it is a POC. The app should remain unchanged apart from the dependency on `eq-author-graphql-schema`.


adammahmood:
@tgandrews Shall we close this PR?

tgandrews:
Will come back in a different form

PR #419, opened 2018-09-07
468 - Pill tabs - Part 1 - Extract base tabs component

tgandrews:
This refactors `ModalWithNav` component to create a `BaseTabs` component that handles all of the tab logic, and accessibility. State management would be handled by the parent component.

As tabs can have many different designs BaseTabs tries to be flexible and uses props that take components and functions which it will call during render.

### How to review 
1. Ensure that ModalWithNav (which is used for the answer validation modal) still renders correctly and performs as well now.
2. Review the API for BaseTabs to ensure that it is flexible enough to build other tab mechanisms on top of.


samiwel:
[Validation - Horizontal Pill Tabs (2)](https://trello.com/c/fYSmJ9GN/468-validation-horizontal-pill-tabs-2)

PR #420, opened 2018-09-07
468 - Pill tabs - Part 1 - Extract base tabs component

tgandrews:
### What is the context of this PR?
This refactors `ModalWithNav` component to create a `BaseTabs` component that handles all of the tab logic, and accessibility. State management would be handled by the parent component.

As tabs can have many different designs BaseTabs tries to be flexible and uses props that take components and functions which it will call during render.

### How to review 
1. Ensure that ModalWithNav (which is used for the answer validation modal) still renders correctly and performs as well now.
2. Review the API for BaseTabs to ensure that it is flexible enough to build other tab mechanisms on top of.

## Note
This was originally opened at #419 but due to the branch name the build could not be completed

PR #421, opened 2018-09-07
468 - Pill tabs - Part 2 Component

tgandrews:
This implements the pill tabs component as per the design. This relies on #420 and so it has the target branch set to it to make reviewing possible.

### How to review 
Load up the storybook and look at the pill tabs component.

### DIFF
--- a/src/components/PillTabs/index.js
+++ b/src/components/PillTabs/index.js
@@ -1,0 +1,74 @@
+import React from "react";
+import PropTypes from "prop-types";
+import styled from "styled-components";
+
+import BaseTabs from "components/BaseTabs";
+
+import { colors } from "constants/theme";
+
+const ListWrapper = styled.div`
+  background-color: ${colors.lightMediumGrey};
+  border-radius: 3em;
+  display: flex;
+  margin: 0.25em 0 1em;
+`;
+const ListItem = styled.button`
+  background: none;
+  border: none;
+  border-radius: 3em;
+  color: ${colors.textLight};
+  cursor: pointer;
+  flex: 1 1 auto;
+  padding: 0.5em 0;
+
+  text-align: center;
+
+  font: inherit;
+
+  &:focus {
+    outline: none;
+    box-shadow: 0 0 0 3px ${colors.tertiary};
+  }
+  &:hover {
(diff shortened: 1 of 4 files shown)


tgandrews:
[Validation - Horizontal Pill Tabs (2)](https://trello.com/c/fYSmJ9GN/468-validation-horizontal-pill-tabs-2)

tgandrews:
Tests not passing as not pointing at master branch.

src/components/PillTabs/index.js
```diff
@@ -0,0 +1,71 @@
+import React from "react";
+import PropTypes from "prop-types";
+import styled from "styled-components";
+
+import BaseTabs from "components/BaseTabs";
+
+import { colors } from "constants/theme";
+
+const ListWrapper = styled.div`
+  background-color: #dde2e6;
```
hamishtaplin: designs have this as `#E4E8EB` which is `lightMediumGrey` from our palette.

src/components/PillTabs/index.js
```diff
@@ -0,0 +1,71 @@
...
+
+  &:focus {
+    outline: none;
+    box-shadow: 0 0 0 3px ${colors.tertiary};
+  }
+  &:hover {
+    background: ${colors.grey};
+    color: ${colors.white};
+  }
+
```
hamishtaplin: There should be an additional focus + hover style that applies our secondary blue to the pill.

src/components/PillTabs/story.js
```diff
@@ -0,0 +1,88 @@
...
+    )
+  }
+];
+
+storiesOf("PillTabs", module)
+  .addDecorator(story => <CenterX>{story()}</CenterX>)
+  .add("Default", () => (
+    <PillTabs onChange={action("change")} options={OPTIONS} />
+  ))
+  .add("State management", () => {
```
hamishtaplin: We tend to avoid having state in stories nowadays as it creates more code to maintain.

tgandrews: I think its useful in this instance as you can see a version where it is working and what you'd need to implement. Thoughts?

PR #422, opened 2018-09-11
Add and delete metadata

adammahmood:
Implements functionality required to add/modify/delete Metadata.

### How to review 
- Run tests
- Ensure feature works as https://trello.com/c/giG4PIKu/548-add-metadata-12 and https://trello.com/c/ZnCEpBZZ/557-delete-metadata-2


adammahmood:
[Metadata front end changes](https://trello.com/c/j1E4jm1p/485-metadata-front-end-changes)

cypress/integration/authenticated/metadata_spec.js
```diff
@@ -0,0 +1,87 @@
+import { testId, addQuestionnaire, selectOptionByLabel } from "../../utils";
+
+describe("Metadata", () => {
```
tgandrews: Should this be stubbing out the network calls?

src/components/MetadataModal/index.test.js
```diff
@@ -0,0 +1,183 @@
...
+      onClose: jest.fn(),
+      isOpen: true,
+      questionnaireId: "1"
+    };
+  });
+
+  describe("data fetching", () => {
+    const render = (mocks, props = {}) =>
+      mount(
+        <TestProvider reduxProps={{ store }} apolloProps={{ mocks }}>
```
tgandrews: Could we test the unwrapped component and not worry about all this setup?

src/components/MetadataTable/Controls/KeySelect.js
```diff
@@ -0,0 +1,105 @@
...
+    const { name, onChange } = this.props;
+    let value;
+
+    if (changes.hasOwnProperty("selectedItem")) {
+      value = changes.selectedItem.value;
+    } else if (changes.hasOwnProperty("inputValue")) {
+      value = changes.inputValue;
+    }
+
+    if (isUndefined(value)) {
```
tgandrews: Could we do this as an else to the block above? Or does it sometimes come as undefined?

src/components/MetadataTable/Controls/KeySelect.test.js
```diff
@@ -0,0 +1,45 @@
+import React from "react";
+import { mount } from "enzyme";
+
+import { KeySelect } from "components/MetadataTable/Controls";
+import {
+  TableTypeaheadInput,
+  TableTypeaheadMenu
+} from "components/DataTable/Controls";
+
+const render = (props = {}) => mount(<KeySelect {...props} />);
```
tgandrews: Do we need mount? Or would shallow be enough?

src/components/MetadataTable/Row.js
```diff
@@ -0,0 +1,121 @@
...
+import {
+  DeleteRowButton,
+  TableInput,
+  TableInputDate
+} from "components/DataTable/Controls";
+import { TableColumn, TableRow } from "components/DataTable/Elements";
+import { KeySelect, Select } from "components/MetadataTable/Controls";
+import { TEXT, DATE, LANGUAGE, REGION } from "constants/metadata-types";
+import { EN, CY } from "constants/languages";
+import { GB_ENG, GB_GBN, GB_NIR, GB_SCT, GB_WLS } from "constants/regions";
```
tgandrews: Could these be exported as an array - do need them separately? Same for metadata types and languages

src/components/MetadataTable/index.js
```diff
@@ -0,0 +1,79 @@
...
+  TableRow,
+  TableColumn,
+  TableHeadColumn
+} from "components/DataTable/Elements";
+import { AddRowButton } from "components/DataTable/Controls";
+import IconText from "components/IconText";
+import Icon from "components/DataTable/icon-plus.svg?inline";
+import Row from "components/MetadataTable/Row";
+
+class MetadataTable extends Component {
```
tgandrews: Does this need to be a class?

src/components/MetadataTable/story.js
```diff
@@ -0,0 +1,73 @@
...
+import MetadataTable from "components/MetadataTable";
+
+const Wrapper = styled.div`
+  position: absolute;
+  padding: 2em;
+`;
+
+const Decorator = storyFn => <Wrapper>{storyFn()}</Wrapper>;
+
+const metadata = [
```
tgandrews: You could share this structure between the story and test

src/components/Typeahead/index.js
```diff
@@ -1,17 +1,8 @@
...
-import PropTypes from "prop-types";
 import Downshift from "downshift";
 
-const Typeahead = ({ onChange, ...otherProps }) => (
-  <Downshift
-    itemToString={item => (item ? item.value : "")}
-    onChange={selection => onChange(selection.value)}
-    {...otherProps}
-  />
+const Typeahead = ({ ...otherProps }) => (
```
tgandrews: I think this could just be `(props)` and then on line 5 `{...props}`

src/components/Typeahead/test.js
```diff
@@ -29,9 +29,9 @@ describe("Typeahead", () => {
     wrapper = shallow(<Typeahead onChange={handleChange} />);
 
     wrapper.simulate("change", apple);
-    expect(handleChange).toHaveBeenCalledWith(apple.value);
+    expect(handleChange).toHaveBeenCalledWith(apple);
```
tgandrews: Is this typeahead used anywhere else? I'm worried somewhere else may be expecting this structure

src/constants/regions.test.js
```diff
@@ -0,0 +1,10 @@
+import * as regions from "constants/regions";
+import { values } from "lodash";
+
+describe("constants/regions", () => {
```
tgandrews: No need for a test for a list of constants.

src/constants/languages.test.js
```diff
@@ -0,0 +1,10 @@
+import * as languages from "constants/languages";
+import { values } from "lodash";
+
+describe("constants/languages", () => {
```
tgandrews: No need to test constants. They'd be covered if they are imported somewhere else.

cypress/integration/authenticated/metadata_spec.js
adammahmood: Do we want to stub everything out or just utilise questionnaire stubs?

src/components/MetadataTable/Controls/KeySelect.js
adammahmood: The value is sometimes `undefined`

src/components/MetadataTable/Controls/KeySelect.test.js
adammahmood: Unfortunately, we need to mount to be able to interact with the component

src/components/MetadataTable/Row.js
adammahmood: If we export these as an array we lose the flexibility of importing them separately.  We don't currently use these anywhere else but thought it would be best to leave it as flexible as possible.

src/components/MetadataTable/index.js
adammahmood: Nope. Will make a dumb component.

src/components/MetadataTable/story.js
adammahmood: Tests don't necessarily have to be DRY.  It's much nicer to have the data in the tests themselves.  Is there anywhere else in eq-author we do this?

src/components/Typeahead/test.js
adammahmood: This is the first time we've used it. 

I had to change this structure to make it easier to control the components state and allow users to select items from the list and input custom values.  In hindsight, we should have written storybook entries demonstrating both use cases.

src/constants/languages.test.js
adammahmood: Best to keep the coverage there IMO.

src/constants/regions.test.js
adammahmood: Best to keep the coverage there IMO.

src/components/CloseButton/index.js
```diff
@@ -0,0 +1,14 @@
+import DeleteButton from "components/DeleteButton";
+import styled from "styled-components";
+
+const CloseButton = styled(DeleteButton).attrs({
+  "aria-label": "Close",
+  size: "medium"
+})`
+  position: absolute;
```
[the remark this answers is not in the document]
adammahmood: What do you think is the best solution?

1) Duplicate this component in Modal and FullScreenModal
2) Abstract styling to Modal and FullScreenModal (this leaves this component a bit pointless)
3) Rename to ModalCloseButton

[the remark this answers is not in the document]
hamishtaplin: yeah sorry I should have offered a solution

can we just keep in the Modal directory or even just export from it's original location?

cypress/integration/authenticated/metadata_spec.js
tgandrews: I think this needs to be discussed at retro. Maybe could leave for the moment.

src/components/MetadataTable/Controls/KeySelect.test.js
tgandrews: Looks like we could simulate a `stateChange` and then just shallow render?

src/components/MetadataTable/Row.js
tgandrews: We could collect them as `import * as Regions from "constants/regions"` as we aren't using them separately at the moment and it clutters the import with unimportant details.

src/components/MetadataTable/story.js
tgandrews: I don't mind it living in the tests and exported to the story. I just felt like its quite a lot of stuff and the majority isn't important. We do this for tabs at the moment.

adammahmood:
Dependant on: https://github.com/ONSdigital/eq-author-api/pull/126

adammahmood:
@samiwel I've had a look and not too sure where that is coming from.  Can you reproduce? Is it definitely set as a default value returned from API and not client side autocomplete?

src/components/MetadataModal/index.js
```diff
@@ -0,0 +1,117 @@
...
+  isOpen: PropTypes.bool.isRequired,
+  onClose: PropTypes.func.isRequired,
+  onAddMetadata: PropTypes.func.isRequired,
+  onDeleteMetadata: PropTypes.func.isRequired,
+  onUpdateMetadata: PropTypes.func.isRequired,
+  questionnaireId: PropTypes.string.isRequired
+};
+
+export default flowRight(
+  withApollo,
```
tgandrews: Is this needed?

src/components/MetadataTable/Controls/KeySelect.js
```diff
@@ -0,0 +1,130 @@
...
+
+  getChangeValue = changes => {
+    if (changes.hasOwnProperty("selectedItem")) {
+      return changes.selectedItem.value;
+    } else if (changes.hasOwnProperty("inputValue")) {
+      return changes.inputValue;
+    }
+  };
+
+  stateReducer = (state, changes) => {
```
adammahmood: It's rather annoying how complex this component has become.  It's mainly due to our requirements of being able to select items from the list, edit selected items and add custom values into the input.  If we use the typeahead elsewhere we should abstract some of this logic into a separate component and expose a simpler API.

PR #423, opened 2018-09-12
449 - Earliest date validation

tgandrews:
This change implements updating earliest date validation for `Date` answers.

### How to review 
1. Add a Date answer to a page.
2. Ensure you can enable it and update all the fields in the modal that is opened.

### Dependencies
- [x] Schema - https://github.com/ONSdigital/eq-author-graphql-schema/pull/62
- [x] API - https://github.com/ONSdigital/eq-author-api/pull/119

### DIFF
--- a/cypress/integration/authenticated/validation_spec.js
+++ b/cypress/integration/authenticated/validation_spec.js
@@ -1,4 +1,5 @@
-import { CURRENCY, NUMBER } from "../../../src/constants/answer-types";
+/* eslint-disable  camelcase */
+import { CURRENCY, DATE, NUMBER } from "../../../src/constants/answer-types";
 
 import {
   addAnswerType,
@@ -8,6 +9,20 @@ import {
   toggleCheckboxOn,
   toggleCheckboxOff
 } from "../../utils";
+
+import createQuestionnaire from "../../fixtures/createQuestionnaire";
+import GetQuestionPage from "../../fixtures/GetQuestionPage";
+import GetQuestionnaire from "../../fixtures/GetQuestionnaire";
+import GetQuestionnaire_Piping from "../../fixtures/GetQuestionnaire_Piping";
+import createDateAnswer from "../../fixtures/answers/createDateAnswer";
+import UpdateQuestionPage from "../../fixtures/duplicatePage/UpdateQuestionPage";
+import ToggleValidationRule from "../../fixtures/answers/ToggleValidationRule";
+import {
+  updateValidationRuleOffsetValue,
+  updateValidationRuleOffsetUnit,
+  updateValidationRuleRelativePosition,
+  updateValidationRuleCustomDate
+} from "../../fixtures/answers/date/updateValidationRule";
 
 describe("Answer Validation", () => {
   it("Can create a questionnaire", () => {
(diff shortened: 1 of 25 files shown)


tgandrews:
[Front end changes for earliest date](https://trello.com/c/R18Ifixp/600-front-end-changes-for-earliest-date)

samiwel:
I will review the code on Monday. But I had a few observations just from playing with the feature using the netlify preview link.

1. The native HTML date picker renders with mm/dd/yyyy in my Chrome window. Do you know if there's anyway of customising this e.g. date format? I think most users would be thinking in terms of dd/mm/yyyy, might be worth verifying with Sarah.

2. If you set a date and tab off mutation is fired off okay. But if you then clear the date and tab off it fires off the mutation as expected but I get an error response.

Wonder if we should be a bit smarter and not fire the mutation if the custom value is empty/null?

tgandrews:
@samiwel 
1. The UI shows dd/mm/yyyy for me - can you check your "Language & Region" settings?
2. Sounds like an API bug as its valid for it to be null. I'll have a look

src/components/Validation/EarliestDate.js
```diff
@@ -0,0 +1,205 @@
...
+
+export const readToWriteMapper = ({ id, customDate, enabled, ...rest }) => ({
+  id,
+  earliestDateInput: {
+    ...rest,
+    custom: customDate ? customDate : null
+  }
+});
+
+const remapToOnUpdate = (propName, mapper) => Component => props => (
```
adammahmood: We should test this method

src/components/Validation/EarliestDate.js
```diff
@@ -0,0 +1,205 @@
...
+`;
+const ConnectedPath = styled.div`
+  padding-left: 1em;
+  position: relative;
+  height: 100%;
+
+  &::after {
+    position: absolute;
+    content: "";
+    background: url(${({ pathUrl }) => pathUrl}) no-repeat center center;
```
adammahmood: We should test `background` is correctly populated with `pathUrl`

adammahmood:
LGTM apart from a couple of missing unit tests.

src/components/Validation/EarliestDate.js
tgandrews: As discussed not sure its worth it given how simple the function is and the component isn't exposed as part of the public api.

cypress/integration/authenticated/validation_spec.js
```diff
@@ -162,4 +177,90 @@ describe("Answer Validation", () => {
...
+            case 0:
+              return updateValidationRuleOffsetValue;
+            case 1:
+              return updateValidationRuleOffsetUnit;
+            case 2:
+              return updateValidationRuleRelativePosition;
+            case 3:
+              return updateValidationRuleCustomDate;
+            default:
+              throw new Error("Call not expected");
```
samiwel: Nice!

src/components/Validation/AnswerValidation.js
```diff
@@ -26,17 +25,37 @@ const Container = styled.div`
...
-    render: () => <MaxValueValidation />
+    render: () => <MaxValueValidation />,
+    types: [CURRENCY, NUMBER]
+  },
+  {
+    id: "earliestDate",
+    title: "Earliest Date",
+    render: () => <EarliestDateValidation />,
+    types: [DATE],
+    custom: "customDate"
```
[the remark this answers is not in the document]
tgandrews: You got it in your comment.

PR #424, opened 2018-09-12
Content picker component

SamGodwin2:
This PR add the new components behind the new style of content picker.

### How to review 
New content picker looks like the designs and the test should pass, (once they're written).

### DIFF
--- a/src/components/ContentPicker/ContentPicker.js
+++ b/src/components/ContentPicker/ContentPicker.js
@@ -1,0 +1,143 @@
+import React from "react";
+import styled from "styled-components";
+import PropTypes from "prop-types";
+import { isNil } from "lodash";
+
+import ButtonGroup from "components/ButtonGroup";
+import Button from "components/Button";
+
+import ContentPickerSingle from "./ContentPickerSingle";
+
+const ContentWrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+  flex-grow: 1;
+  width: 40em;
+  height: 16em;
+  margin: 1em 0;
+`;
+
+export default class ContentPicker extends React.Component {
+  static propTypes = {
+    onSubmit: PropTypes.func,
+    onClose: PropTypes.func,
+    data: PropTypes.arrayOf(
+      PropTypes.shape({
+        id: PropTypes.string.isRequired
+      })
+    ).isRequired,
+    config: PropTypes.arrayOf(
+      PropTypes.shape({
+        id: PropTypes.string.isRequired,
+        title: PropTypes.string.isRequired,
+        childKey: PropTypes.string
+      })
+    ).isRequired
+  };
+
+  state = {
+    selectedItems: []
+  };
+
+  getDataAtLevel(level) {
+    if (level === 0) {
+      return this.props.data;
+    }
+    const parentLevel = level - 1;
+    const parentItem = this.state.selectedItems[parentLevel];
+    if (!parentItem) {
+      return [];
+    }
+    const { config } = this.props;
+    return parentItem[config[parentLevel].childKey];
+  }
+
+  handleTitleClick = level => {
+    const { config } = this.props;
+    this.setState(state => {
+      let openLevel = level;
+      if (state.openLevel === level && state.selectedItems[level]) {
+        openLevel = Math.min(state.selectedItems.length, config.length - 1);
+      }
+      return {
+        openLevel
+      };
+    });
+  };
+
+  handleOptionClick = (level, option) => {
+    const { config } = this.props;
+    this.setState(state => ({
+      openLevel: Math.min(level + 1, config.length - 1),
+      selectedItems: [...state.selectedItems.slice(0, level), option]
+    }));
+  };
+
+  renderPickers() {
+    const { config } = this.props;
+    const { selectedItems, openLevel } = this.state;
+    return config.map(({ id, title: defaultTitle }, level) => {
+      const isDisabled = level > selectedItems.length;
+      const isHidden = level > openLevel;
+      const isOpen = level === openLevel;
+
+      const selectedItem = selectedItems[level];
+      const isSelected = selectedItem && !isOpen;
+      const isLastLevel = config.length - 1 === level;
+      const useCustomTitle =
+        !isLastLevel && selectedItem && selectedItem.plaintextTitle;
+
+      const title = useCustomTitle
+        ? `${defaultTitle}: ${selectedItem.plaintextTitle}`
+        : defaultTitle;
+
+      const data = this.getDataAtLevel(level);
+
+      return (
+        <ContentPickerSingle
+          key={id}
+          title={title}
+          data={data}
+          disabled={isDisabled}
+          hidden={isHidden}
+          open={isOpen}
+          selected={isSelected}
+          selectedOption={(selectedItem || {}).id}
+          onTitleClick={() => this.handleTitleClick(level)}
+          onOptionClick={option => this.handleOptionClick(level, option)}
+          data-test={`${id}-picker`}
+        />
+      );
+    });
+  }
+
+  render() {
+    const { config } = this.props;
+    const { selectedItems } = this.state;
+    const selectedItem = selectedItems[config.length - 1];
+    return (
+      <div>
+        <ContentWrapper>{this.renderPickers()}</ContentWrapper>
+        <ButtonGroup horizontal align="right">
+          <Button
+            variant="secondary"
+            type="button"
+            onClick={this.props.onClose}
+            data-test="cancel-button"
+          >
+            Cancel
+          </Button>
+          <Button
+            type="submit"
+            variant="primary"
+            disabled={isNil(selectedItem)}
+            onClick={() => this.props.onSubmit(selectedItem.id)}
+            data-test="submit-button"
+          >
+            Select
+          </Button>
+        </ButtonGroup>
+      </div>
+    );
+  }
+}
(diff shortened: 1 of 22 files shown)


SamGodwin2:
[Content picker - component](https://trello.com/c/YIF5SEZ6/521-content-picker-component)

src/components/ContentPicker/contentPicker.test.js
```diff
@@ -0,0 +1,163 @@
...
+      const wrapper = createWrapper({ data: generateMockPiping(1, 1, 1) });
+
+      wrapper.setState({
+        selectedSection: 1,
+        selectedQuestion: 1,
+        selectedAnswer: 1
+      });
+
+      wrapper.find("[data-test='submit-button']").simulate("click");
+      expect(onSubmit).toHaveBeenCalledWith(1);
```
tgandrews: Could we have the generateMockPiping return more complex ids such as `answer1` so that this assertion makes more sense this test could return the selected section id and would still pass.

src/components/ContentPicker/contentPickerPanel.js
```diff
@@ -0,0 +1,51 @@
...
+    role="tabpanel"
+    aria-hidden={!open}
+    aria-labelledby={labelledBy}
+    {...otherProps}
+  >
+    <StyledScrollPane>{children}</StyledScrollPane>
+  </div>
+);
+
+const StyledScrollPane = styled(ScrollPane)`
```
tgandrews: Definition before use. Maybe we could move it up

src/components/ContentPicker/contentPickerTitle.js
```diff
@@ -0,0 +1,105 @@
...
+  border-radius: 5px;
+  width: 100%;
+  height: 2.5em;
+  font-size: 1em;
+  text-align: left;
+  display: inline;
+
+  ${props => (props.disabled ? titleStyles.disabled : null)};
+  ${props => (props.selected ? titleStyles.selected : null)};
+  ${props => (props.hidden ? titleStyles.hidden : null)};
```
tgandrews: Could we not implement this in the parent so that the button isn't rendered when hidden?

src/components/ContentPicker/index.js
```diff
@@ -0,0 +1,242 @@
...
+import { find, get, isNil } from "lodash";
+import ButtonGroup from "components/ButtonGroup";
+import Button from "components/Button";
+
+const PickerWrapper = styled.div`
+  padding-bottom: 3px;
+
+  ${props =>
+    props.open
+      ? css`
```
tgandrews: IMO Little bit clunky in the ternary could we move out to be `openStyles`

src/components/ContentPicker/index.js
```diff
@@ -0,0 +1,242 @@
...
+};
+
+const getDataLevel = (data, selectedSectionId, selectedQuestionId) => {
+  const pagesInSection = get(
+    find(data.sections, ({ id }) => id === selectedSectionId),
+    "pages"
+  );
+  return !isNil(selectedQuestionId)
+    ? get(
+        find(pagesInSection, ({ id }) => id === selectedQuestionId),
```
tgandrews: I'm not sure I understand this - find a page that has the same `id` as the `selectedQuestionID`?

src/components/ContentPicker/index.js
```diff
@@ -0,0 +1,242 @@
...
+    >
+      {data.map(option => {
+        return (
+          <OptionButton
+            id={`option-${option.id}`}
+            key={option.id}
+            selected={selectedOption === option.id}
+            onClick={() => handleOptionClick(option)}
+          >
+            {option.plaintextTitle
```
tgandrews: I presume this funky-ness will be removed with the displayName change

src/components/ContentPicker/index.js
```diff
@@ -0,0 +1,242 @@
...
+      )
+    : pagesInSection;
+};
+
+export const ContentPickerSingle = ({
+  id,
+  handleTitleClick,
+  handleOptionClick,
+  title,
+  openAccordionId,
```
tgandrews: This feels like it should not be passed down - and maybe an isOpen prop should be passed in

src/components/ContentPicker/index.js
```diff
@@ -0,0 +1,242 @@
...
+    this.setState({
+      openAccordion: 1,
+      sectionTitle: defaultTitles.section,
+      selectedQuestion: null,
+      questionTitle: defaultTitles.question,
+      selectedAnswer: null
+    });
+  };
+
+  onHandleSectionOptionClick = option => {
```
tgandrews: A lot of these handlers look very similar and I think we can rework this to (maybe) make the rendering more complex but remove a lot of this duplication

src/tests/utils/generateMockPiping.js
```diff
@@ -0,0 +1,44 @@
...
+  let pageArray = [];
+  let answerArray = [];
+
+  let x, y, z;
+
+  for (x = 0; x < sections; x++) {
+    for (y = 0; y < pages; y++) {
+      for (z = 0; z < answers; z++) {
+        answerArray.push({
+          id: answerId,
```
tgandrews: You are tracking `x`, `y`, `z` and `answerId`, `pageId` and `sectionId`. And the only difference is that they are off by one from each other. I don't think we need both.

src/components/ContentPicker/OptionButton.js
```diff
@@ -0,0 +1,59 @@
...
+import styled, { css } from "styled-components";
+import { colors } from "constants/theme";
+
+const selectedStyle = css`
+  color: ${colors.white};
+  background: ${colors.primary};
+`;
+
+const OptionButton = styled(Button)`
+  width: 98%;
```
tgandrews: Bit of an odd width - could `calc` help here?

src/components/ContentPicker/index.js
SamGodwin2: Yeah was just a easy way to do it until the display name stuff comes in.

src/components/ContentPicker/index.js
SamGodwin2: Yeah it's a bit inconsistent, does questionsInSection make more sense?

src/components/ContentPicker/index.js
SamGodwin2: So the alternative here would be to do the props derivation in the non-functional component and pass all the individual props down. Would that be neater?

src/components/ContentPicker/OptionButton.js
SamGodwin2: Never seen that before, would defo help, ta.

src/tests/utils/generateMockPiping.js
SamGodwin2: They're not quite the same, x,y,z reset every loop cycle whereas the id values always keep climbing, if we only tracked one then I think the answer id's would always be answer 1- answer 5 for every page and the same for pages, if that makes sense??

tgandrews: Got it - makes sense.

src/components/ContentPicker/OptionButton.js
```diff
@@ -0,0 +1,59 @@
...
+    onClick: PropTypes.func.isRequired,
+    id: PropTypes.string.isRequired,
+    children: PropTypes.node.isRequired,
+    className: PropTypes.string,
+    selected: PropTypes.bool
+  };
+
+  setTitleRef = title => {
+    this.title = title;
+  };
```
samiwel: I don't think this is being used?

samiwel: If it's not being used, I think PickerOption can be removed altogether and just export OptionButton?

src/components/ContentPicker/contentPickerTitle.js
```diff
@@ -0,0 +1,101 @@
...
+    open: PropTypes.bool.isRequired,
+    children: PropTypes.node.isRequired,
+    className: PropTypes.string,
+    selected: PropTypes.bool,
+    disabled: PropTypes.bool
+  };
+
+  setTitleRef = title => {
+    this.title = title;
+  };
```
samiwel: Actually, we don't seem to be doing anything with this reference either. As far as I can tell we're passing in `onClick()` to handle selecting the title. So it can probably be removed and the ref prop in the render function below can also be removed, maybe?

src/components/ContentPicker/index.js
```diff
@@ -0,0 +1,267 @@
...
+
+const defaultTitles = {
+  section: "Section",
+  question: "Question",
+  answer: "Answer"
+};
+
+const getDataLevel = (data, selectedSectionId, selectedQuestionId) => {
+  const questionsInSection = get(
+    find(data.sections, ({ id }) => id === selectedSectionId),
```
samiwel: I think this could be simplified to:
```javascript
find(data.sections, { id: selectedSectionId})
```

src/components/ContentPicker/index.js
```diff
@@ -0,0 +1,267 @@
...
+    "pages"
+  );
+  return !isNil(selectedQuestionId)
+    ? get(
+        find(questionsInSection, ({ id }) => id === selectedQuestionId),
```
samiwel:
```javascript
find(questionsInSection, { id: selectedQuestionId})
```

src/components/ContentPicker/ContentPicker.js
```diff
@@ -0,0 +1,143 @@
...
+          <Button
+            variant="secondary"
+            type="button"
+            onClick={this.props.onClose}
+            data-test="cancel-button"
+          >
+            Cancel
+          </Button>
+          <Button
+            type="submit"
```
hamishtaplin: Does this need to be a submit button? There's no `form` element so semantically this isn't correct. Probably doesn't matter but could be removed.

PR #425, opened 2018-09-13
Refactor plaintextTitle to utilise displayName

adammahmood:

We have inconsistent display values scattered through the author application. In some places, we display titles, in other places we display labels and in other places we have inconsistent unnamed or untitled placeholders.

This technical improvement makes it the responsibility of the API to provide the most relevant displayable value.

More info can be found here: https://github.com/ONSdigital/eq-author/wiki/Display-names

### How to review

- Run tests

### Dependencies 

- [x] https://github.com/ONSdigital/eq-author-graphql-schema/pull/63
- [x] https://github.com/ONSdigital/eq-author-api/pull/122

### DIFF
--- a/cypress/integration/authenticated/routing_spec.js
+++ b/cypress/integration/authenticated/routing_spec.js
@@ -142,7 +142,7 @@ describe("Routing", () => {
 
     addSection();
 
-    cy.contains("Page Title").click();
+    cy.contains("Untitled Page").click();
 
     typeIntoDraftEditor(testId("txt-question-title", "testid"), "Question 4");
     buildMultipleChoiceAnswer(["G", "H", "I"]);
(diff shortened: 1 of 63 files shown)


cypress/integration/authenticated/routing_spec.js
```diff
@@ -292,8 +281,7 @@ describe("Routing", () => {
 
     cy.get(testId("routing-rule"));
 
-    cy
-      .get(testId("result-selector"))
+    cy.get(testId("result-selector"))
```
tgandrews: We should run prettier on the whole codebase as these sorts of changes are a little noisy in every PR.

src/components/QuestionnaireMenu/index.js
```diff
@@ -13,7 +13,7 @@ const PageMenu = ({ pages, sectionNumber, menuZIndex, ...otherProps }) => (
   <Dropdown maxWidth={"25em"}>
     <MenuList maxHeight={`${MAX_HEIGHT}em`}>
       {map(pages, (page, i) => {
-        const title = page.plaintextTitle;
+        const title = page.displayName;
```
tgandrews: This file also has defaults on lines 31 and 51 that I think can be removed

cypress/integration/authenticated/routing_spec.js
adammahmood: @tgandrews Yea I agree.  I will do it shortly when I'm inbetween tasks.

tgandrews:
PR looks good just one issue discussed on slack with the piping menu answer name not being updated.

adammahmood:
@tgandrews Fixed the answer not updating in the piping menu

samiwel:
[Introduce displayName](https://trello.com/c/zUaulBzP/607-introduce-displayname)

src/containers/enhancers/withDeleteSection.js
```diff
@@ -68,7 +68,9 @@ export const deleteUpdater = (questionnaireId, sectionId) => (
 };
 
 export const displayToast = (ownProps, questionnaire) => {
-  const { match: { params } } = ownProps;
+  const {
```
hamishtaplin: Any idea why this is being reformatted like this? Original formatting was much nicer

adammahmood: Think it's related to the Prettier upgrade.  I'll run on the entire codebase and raise a separate PR.

src/components/RichTextEditor/entities/PipedValue.js
```diff
@@ -78,7 +78,7 @@ export const insertPipedValue = (answer, contentState, selection) => {
   );
 
   const entityKey = newContent.getLastCreatedEntityKey();
-  const text = answer.label || "Piped Value";
+  const text = answer.displayName;
```
hamishtaplin: We should make this change everywhere so there's no display logic around display of labels/titles and everything defers to the response from the API.

src/components/RichTextEditor/entities/PipedValue.js
```diff
@@ -64,7 +64,7 @@ export const replacePipedValues = labels => (
 ) => {
   const text = labels.hasOwnProperty(entity.data.id)
     ? labels[entity.data.id]
-    : "Deleted Piped Value";
+    : `Untitled Answer`;
```
hamishtaplin: Is this right? I thought we said leave deleted answers as-is.

Issue #428, opened 2018-09-19
Sidebar not updating with latest piped values 


adammahmood:
### Steps to reproduce the behaviour

1) Create Answer with title "Answer One"
2) Create Answer in a new section and pipe "Answer One"
3) Delete "Answer One"
4) Navigate back to the new section 

Expected: 
- Question title piped value is updated with "[Untitled Answer]"
- Sidebar piped value is updated with "[Untitled Answer]"

Actual:
- Question title piped value is updated with "[Untitled Answer]"
- Sidebar piped value is **NOT** updated with "[Untitled Answer]" and retains the old piped value
- Hard reload does not resolve this issue
- Editing the title resolves this issue

samiwel:
@adammahmood is this something we're deferring. It sounds similar to what @tgandrews mentioned to me this morning about the pain he's going to be having piping metadata. Shall we discuss in the next tech session?

adammahmood:
We should definitely look at this soon-ish. 

Currently, the sidebar does not attempt to fetch the latest data for a piped answer, therefore these values end up stale. We need to re-architect how piping values are stored and stop relying on the UI to rebuild pipe values.

PR #430, opened 2018-09-20
592 - Integrate content picker with piping menu

tgandrews:
Integrates the ContentPicker with the new piping menu.

### How to review 
1. Ensure you can add piped values from previous answers.
2. Ensure the tests pass.

### Dependencies
- [x] - Content Picker component - #424


tgandrews:
[Hooking content picker into the existing piping ui.](https://trello.com/c/PQWO7GWS/615-hooking-content-picker-into-the-existing-piping-ui)

adammahmood:
Yet to review code but a few minor things I have noticed when testing the feature locally:

1) Section/Question/Answer is marked as "Unlabelled Option" even though I have a Title/Label.  Should this not be using `displayName` and utilise the title or fallback to our default?
2) When opening the content picker, you can click on the "Section" accordion to open a list of Sections, however, you can't close the list until you select an item.  You can only click on an item.  Is this intended?  
3) When deleting a piped answer, the updates aren't reflected on the input where the value is being piped. A hard reload fixes this. There's an [existing issue](https://github.com/ONSdigital/eq-author/issues/428) where the sidebar isn't updated, however, the input was still updating.

tgandrews:
1 and 3. I presume these are related. I will look into it.
2. Yes, this as designed.

adammahmood:
@tgandrews Looks like 1 and 3 are now resolved. 👍

src/components/RichTextEditor/PipingMenu.test.js
```diff
@@ -81,7 +81,7 @@ describe("PipingMenu", () => {
     match = createMatch(questionnaire.id, section.id, page.id);
     const wrapper = render();
 
-    expect(wrapper.prop("disabled")).toBe(true);
+    expect(wrapper.find("PipingMenu__MenuButton").prop("disabled")).toBe(true);
```
adammahmood: Is it better to export `MenuButton` and use `expect(wrapper.find(MenuButton).prop("disabled")).toBe(true);` ?

adammahmood:
A couple of things relating to the content picker (which I appreciate isn't part of this PR) and I can't comment on the lines:

1) `src/components/ContentPicker/ContentPickerSingle.js`  
`data: PropTypes.any // eslint-disable-line react/forbid-prop-types`.  Why do we allow any structure of `data` when we expect the object to be a certain shape when it's used:

```
  {data.map(option => {
          return (
            <PickerOption
              id={`option-${option.id}`}
              key={option.id}
              selected={selectedOption === option.id}
              onClick={() => onOptionClick(option)}
            >
              {option.displayName}
            </PickerOption>
          );
        })}
```

2) Same question for `src/components/ContentPickerModal`

cypress/integration/authenticated/piping_spec.js
```diff
@@ -0,0 +1,54 @@
...
+    cy.get(testId("nav-page-link"))
+      .last()
+      .click();
+
+    cy.get(testId("txt-question-title", "testid")).focus();
+    cy.get(testId("piping-button"))
+      .first()
+      .click();
+
+    cy.get("#Section-title").click();
```
adammahmood: Why is the component capitalising these ids?  We usually keep all ids lower case.

tgandrews: It's deriving them from input but can change.

tgandrews:
@adammahmood I have updated to remove the eslint ignores and specify what we actually require. I have also changed to remove the ids which were using title which was wrong as this was a text to display to the user.

samiwel:
image
Nice 👍

PR #439, opened 2018-09-25
493 piping metadata

tgandrews:
This implements the ability to pipe metadata to a question title.

### How to review 
1. Create some metadata
2. Click the piping menu in a question title
3. Selecting the metadata should show it in the title editor.
4. Updating the alias or deleting the metadata will update the displayed value in the editor.

### Dependencies
- [x] - API - https://github.com/ONSdigital/eq-author-api/pull/137


tgandrews:
[Pipe metadata into a question - front end changes](https://trello.com/c/fkHnQhid/536-pipe-metadata-into-a-question-front-end-changes)

src/components/ContentPicker/ContentPicker.test.js
```diff
@@ -55,6 +55,21 @@ describe("Content Picker", () => {
       expect(wrapper.find(ANSWER_PICKER_SELECTOR).prop("disabled")).toBe(true);
     });
 
+    it("should start open when there is only level", () => {
```
adammahmood: Typo.  I guess this should read: "should start open when there is only **one** level"

src/components/RichTextEditor/index.js
```diff
@@ -172,35 +183,91 @@ class RichTextEditor extends React.Component {
...
-    const createAnswerMap = flow(
-      keyBy("id"),
-      mapValues("displayName")
+    const metadataPipes = getMetadataPipes(pipes);
+    if (metadataPipes.length === 0) {
+      return;
+    }
+
+    const renamedMetadata = this.props.metadata.map(m => ({
+      displayName: m.alias,
```
adammahmood: Another instance of `displayName` being useful for Metadata

src/components/RichTextEditor/index.js
```diff
@@ -172,35 +183,91 @@ class RichTextEditor extends React.Component {
...
+
+    this.renamePipedValues(
+      () => renamedMetadata,
+      metadataPipes,
+      "Deleted Metadata"
     );
-    const fetchAnswersForEntities = flow(
+  }
+
+  updateAnswerPipedValues(pipes, editorState, contentState) {
```
adammahmood: `editorState` and `contentState` params are unnecessary.

src/components/RichTextEditor/index.js
```diff
@@ -172,35 +183,91 @@ class RichTextEditor extends React.Component {
...
+    if (!pipes.length) {
+      return;
+    }
+
+    this.updateAnswerPipedValues(pipes, editorState, contentState);
+    this.updateMetadataPipedValues(pipes, editorState, contentState);
+  }
 
-    if (!entities.length) {
+  updateMetadataPipedValues(pipes, editorState, contentState) {
```
adammahmood: `editorState` and `contentState` params are unnecessary.

adammahmood: Maybe we should add https://eslint.org/docs/rules/no-unused-vars to eslint.

adammahmood:
Also just been using the feature.  Works nicely apart from a few things:

1) Deleting a Metadata which is being used doesn't update to `[Deleted Metadata]` until you navigate away from the page and return to it. 
2) Updating a Metadata is not reflected in the question (even after a refresh). Updating a piped answer is immediately reflected.
3) If you enter an empty row in the Metadata table and then attempt to pipe this Metadata, it appears as a selectable item in the ContentPicker list and you can then select it (see screenshots).
4) The Answer and Metadata tabs on the ContentPicker aren't keyboard accessible.

screen shot 2018-09-25 at 12 01 23
screen shot 2018-09-25 at 12 01 31

src/components/ContentPicker/index.js
```diff
@@ -21,3 +22,28 @@ const config = [
...
+    title: "Metadata"
+  }
+];
+
+export const MetadataContentPicker = props => (
+  <ContentPicker
+    {...props}
+    data={props.data.map(md => ({
+      ...md,
+      displayName: md.alias
```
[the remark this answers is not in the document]
tgandrews: Yes, I agree but in a separate change on the API. I'll look at doing that this week and then we can remove this but I don't want to block this because of it.

tgandrews:
1. Yes, that works like answers. It requires the component to remount.
2. Which field are you updating? It *only* uses the alias. It works fine when you update alias.
3. Yes, this is displayName logic - I guess I should default it to some value for the moment. And make the change server side to remove this after.
4. You can tab to every field and use it that way - are you expecting something different?

src/components/ContentPicker/index.js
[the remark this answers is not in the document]
adammahmood: Much nicer.

src/components/ContentPickerModal/__snapshots__/index.test.js.snap
```diff
@@ -0,0 +1,41 @@
...
+          "render": [Function],
+          "title": "Metadata",
+        },
+      ]
+    }
+  />
+</ContentPickerModal__StyledModal>
+`;
+
+exports[`ContentPickerModal should show an error message when there are answers to pick from 1`] = `
```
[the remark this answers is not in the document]
adammahmood: Where's this test gone?

adammahmood:
Another issue I have noticed:

- Go to a question page. Add an empty Metadata row.  Click Done.  The question page displays "Something went wrong".  A refresh solves this. 

screen shot 2018-09-26 at 09 39 50

tgandrews:
@adammahmood please re-review. The issue is resolved and the API has been merged so it should be less painful! 🎉

adammahmood:
LGTM 👍

PR #441, opened 2018-09-25
Changed the link generated for preview surveys to not use go-launch 

SamGodwin2:
This PR adds the ability to handle the go-launch env var being null so that we can gracefully switch over from using GLAS to our own api endpoint.
 
### How to review 
Tests pass all existing behaviour is intact.

### DIFF
--- a/src/components/Header/index.js
+++ b/src/components/Header/index.js
@@ -97,14 +97,21 @@ export class UnconnectedHeader extends React.Component {
   };
 
   getPreviewUrl(questionnaireId) {
-    const timestamp = Date.now();
-    const publisherUrl = process.env.REACT_APP_PUBLISHER_URL;
-    const goLaunchASurveyQuickLaunchUrl =
-      process.env.REACT_APP_GO_LAUNCH_A_SURVEY_URL;
-    const urlEncodedParam = encodeURIComponent(
-      `${publisherUrl}/${questionnaireId}?r=${timestamp}`
-    );
-    return `${goLaunchASurveyQuickLaunchUrl}?url=${urlEncodedParam}`;
+    if (process.env.REACT_APP_GO_LAUNCH_A_SURVEY_URL) {
+      const timestamp = Date.now();
+      const publisherUrl = process.env.REACT_APP_PUBLISHER_URL;
+      const goLaunchASurveyQuickLaunchUrl =
+        process.env.REACT_APP_GO_LAUNCH_A_SURVEY_URL;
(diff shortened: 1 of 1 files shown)


src/components/Header/index.js
```diff
@@ -96,24 +96,15 @@ export class UnconnectedHeader extends React.Component {
...
-  }
-
   handleSignOut = () => {
     this.props.signOutUser();
   };
 
   handleShare = () => {
     const textField = document.createElement("textarea");
-    textField.innerText = this.getPreviewUrl(this.props.questionnaire.id);
+    textField.innerText = `${process.env.REACT_APP_API_URL}/launch/${
```
tgandrews: Can we add a test for this at all

.env.development
```diff
@@ -1,7 +1,6 @@
 REACT_APP_BASE_NAME=""
 REACT_APP_API_URL="http://localhost:4000/graphql"
 REACT_APP_PUBLISHER_URL="http://docker.for.mac.localhost:9000/publish"
-REACT_APP_GO_LAUNCH_A_SURVEY_URL="http://localhost:8000/quick-launch"
```
samiwel: Should we leave this in until we've flipped the switch?

src/components/Header/index.js
```diff
@@ -97,14 +97,21 @@ export class UnconnectedHeader extends React.Component {
...
+      const goLaunchASurveyQuickLaunchUrl =
+        process.env.REACT_APP_GO_LAUNCH_A_SURVEY_URL;
+      const urlEncodedParam = encodeURIComponent(
+        `${publisherUrl}/${questionnaireId}?r=${timestamp}`
+      );
+      return `${goLaunchASurveyQuickLaunchUrl}?url=${urlEncodedParam}`;
+    } else {
+      const baseURL = process.env.REACT_APP_API_URL
+        ? process.env.REACT_APP_API_URL.replace(/(\/graphql)/, "")
+        : "";
```
samiwel: When we switch over for real, we should probably remove `/graphql` from the env variable so we don't have to do this regex replacement.

.env.development
SamGodwin2: Yup took it out for testing and forgot to put it back in.

src/components/Header/index.js
samiwel: Or have two separate env variables e.g. `REACT_APP_API_URL` and `REACT_APP_LAUNCH_URL`

SamGodwin2: Yep agreed 👍

PR #442, opened 2018-09-25
Add missing unit tests

adammahmood:
- Adds missing unit tests highlighted by our test coverage

### How to review 
- Run tests


adammahmood:
[Improve test coverage on Author](https://trello.com/c/vyPqDT9o/586-improve-test-coverage-on-author)

src/components/AnswerEditor/answereditor.test.js
```diff
@@ -152,4 +164,20 @@ describe("Answer Editor", () => {
...
+      ...mockAnswer,
+      type: DATE,
+      properties: { format: "mm/yyyy" }
+    };
+    const instance = createWrapper({
+      answer,
+      ...mockMutations
+    }).instance();
+
+    expect(instance.formatIncludes("dd")(answer)).toEqual(false);
```
tgandrews: Can we derive this another way as this is testing private API of the component not the expected output.

src/components/Answers/Date/index.test.js
```diff
@@ -19,11 +19,10 @@ describe("Date", () => {
     handleUpdate = jest.fn();
     store = createMockStore();
```
tgandrews: We can remove this as well.

src/components/AnswerEditor/answereditor.test.js
adammahmood: Not that I can think of.  Any ideas?

src/components/App/index.test.js
```diff
@@ -0,0 +1,28 @@
...
+      <div>Content</div>
+    </UnconnectedApp>
+  );
+
+describe("components/App", () => {
+  let props, wrapper;
+
+  beforeEach(() => {
+    props = {
+      gainConnection: jest.fn(),
```
tgandrews: We could stub the window to prove these are called.

src/components/MoveSectionModal/index.test.js
```diff
@@ -1,25 +1,48 @@
...
-    );
+  let questionnaire, section, props, wrapper;
+
+  beforeEach(() => {
+    questionnaire = buildQuestionnaire();
+    section = questionnaire.sections[0];
+    props = {
+      questionnaire,
+      section,
+      onClose: jest.fn(),
```
tgandrews: We could add a test to ensure this is called.

src/components/AnswerEditor/answereditor.test.js
tgandrews: They result of the functions are shown - so if the answer is type "Date" then you can see them in the output.

PR #445, opened 2018-10-01
Make author forwards compatible

SamGodwin2:
After the changes in the one of the previous PR's the api link was made backwards compatible but was unfortunately not also made forwards compatible as well. This PR aims to change that and allow the terraform changes to go in smoothly.

### How to review 
Run Author and ensure that it works both with and without the `/graphql` endpoint on the api.


src/index.js
```diff
@@ -43,7 +43,9 @@ const history = createHistory({
 
 const link = ApolloLink.from([
   createErrorLink(getStore),
-  createHttpLink(process.env.REACT_APP_API_URL)
+  createHttpLink(
+    `${process.env.REACT_APP_API_URL.replace(/(\/graphql)/, "")}/graphql`
```
samiwel: Let's introduce a separate launch URL variable rather than do this.

PR #446, opened 2018-10-01
Switched min/max values to use ints instead

SamGodwin2:
After graphql was updated to to the latest version a bug was introduced where the api no-longer coerced string int values into actual in values and instead it threw an error. This PR changes that so it now imports int values instead.

### How to review 
Tests pass and min and max values can be changed.

### DIFF
--- a/src/components/Validation/MaxValue.js
+++ b/src/components/Validation/MaxValue.js
@@ -5,7 +5,7 @@ import styled from "styled-components";
 import { withApollo } from "react-apollo";
 import { propType } from "graphql-anywhere";
 
-import { flowRight, inRange } from "lodash";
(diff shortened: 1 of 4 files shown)


src/components/Validation/MaxValue.js
```diff
@@ -56,11 +56,13 @@ export const MaxValue = ({
...
     }
 
+    const intValue = parseInt(value, 10);
+
     const updateValidationRuleInput = {
       id: maxValue.id,
       maxValueInput: {
         inclusive: maxValue.inclusive,
-        custom: value === "" ? null : value
+        custom: isNaN(intValue) ? null : intValue
```
tgandrews: Can you add a unit test to show this is always come out as ints now and not strings.

PR #448, opened 2018-10-02
Removed all mentions of previous launching mechanisms

SamGodwin2:
This PR aims to remove all mention of the survey-launcher and publisher from the front-end and allow it to rely on the new launching mechanism provided by the api.

### How to review 
All tests should pass and there should no-longer be any reference to Publisher or go-launcher.


src/components/Header/index.js
```diff
@@ -96,28 +96,15 @@ export class UnconnectedHeader extends React.Component {
...
-  }
-
   handleSignOut = () => {
     this.props.signOutUser();
   };
 
   handleShare = () => {
     const textField = document.createElement("textarea");
-    textField.innerText = this.getPreviewUrl(this.props.questionnaire.id);
+    textField.innerText = `${process.env.REACT_APP_LAUNCH_URL}/${
```
tgandrews: Could we leave the getPreviewUrl function and have it always return the built URL.

PR #455, opened 2018-10-05
Lint errors should not block server from running

tgandrews:
1. Build errors are the only thing that should stop a build from starting/updating
1. We now check linting on push (and people should have it set up in their editors)
1. We can now treat all linting warnings as errors by making it so in the config.

### How to review 
1. Ensure you can make changes that cause linting errors (e.g. unused variables) but you can still run the app.
1. Make a commit with a linting error in and it should fail to push.

### DIFF
--- a/config/webpack.config.common.js
+++ b/config/webpack.config.common.js
@@ -2,7 +2,6 @@
 "use strict";
 
 const path = require("path");
-const eslintFormatter = require("react-dev-utils/eslintFormatter");
 const ModuleScopePlugin = require("react-dev-utils/ModuleScopePlugin");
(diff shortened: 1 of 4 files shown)


tgandrews:
I think my position on this has softened over the weekend and I can understand why people would only want to see warnings for these types of things in their editor. The lint errors would only be for build errors (which feels like we are trying to mirror what would fail in webpack) so I think the shared config change is needed as well. And maybe removing it from build is a nice to have more than anything else.

adammahmood:
Closing due to fix: https://github.com/ONSdigital/eslint-config-eq-author/pull/8

PR #456, opened 2018-10-05
Tech debt stop ui repeatedly building

tgandrews:
Change the way we build the UI docker image so that:
1. It does not build every time you run it (to be updated for the new environment variables)
1. Make the image significantly smaller as it just contains the build output and the nginx configuration.
1. Runtime environment variables passed to docker are still applied. 
1. Only `NODE_ENV` and `REACT_APP_ENABLE_AUTH` are determined when the docker image is built. Both are either hardcoded or picked up automatically as part of the `.env.production`.
1. Image size now: 124MB from 1.77GB
1. Image start time 2-3 seconds from >80 seconds

### How to review 
1. Ensure you can build and run the built image locally. 

### Dependencies
- [x] - #451

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -45,6 +45,20 @@
       })(window,document,window['_fs_namespace'],'script','user');
     }
   </script>
+  <script type="text/javascript">
+    window.config = {};
+    window.config.REACT_APP_API_URL="";
+    window.config.REACT_APP_BASE_NAME="";
+    window.config.REACT_APP_FIREBASE_API_KEY="";
+    window.config.REACT_APP_FIREBASE_MESSAGING_SENDER_ID="";
+    window.config.REACT_APP_FIREBASE_PROJECT_ID="";
+    window.config.REACT_APP_FUNCTIONAL_TEST="";
+    window.config.REACT_APP_GO_LAUNCH_A_SURVEY_URL="";
+    window.config.REACT_APP_LAUNCH_URL="";
+    window.config.REACT_APP_PUBLISHER_URL="";
+    window.config.REACT_APP_USE_FULLSTORY="";
+    window.config.REACT_APP_USE_SENTRY="";
+  </script>
   <body>
     <div id="root"></div>
     <div id="toast"></div>
--- a/src/config.js
+++ b/src/config.js
@@ -1,0 +1,34 @@
+window.config = window.config || {};
+const config = {
+  REACT_APP_API_URL:
+    window.config.REACT_APP_API_URL || process.env.REACT_APP_API_URL,
+  REACT_APP_BASE_NAME:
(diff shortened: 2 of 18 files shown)


public/index.html
```diff
@@ -45,6 +45,20 @@
...
+    window.config.REACT_APP_BASE_NAME="";
+    window.config.REACT_APP_FIREBASE_API_KEY="";
+    window.config.REACT_APP_FIREBASE_MESSAGING_SENDER_ID="";
+    window.config.REACT_APP_FIREBASE_PROJECT_ID="";
+    window.config.REACT_APP_FUNCTIONAL_TEST="";
+    window.config.REACT_APP_GO_LAUNCH_A_SURVEY_URL="";
+    window.config.REACT_APP_LAUNCH_URL="";
+    window.config.REACT_APP_PUBLISHER_URL="";
+    window.config.REACT_APP_USE_FULLSTORY="";
+    window.config.REACT_APP_USE_SENTRY="";
```
samiwel: What is the reason for defaulting to empty strings here?

Wonder if it is possible to add some kind of a template placeholder here, e.g. `{{REACT_ENV_VARS}}` or similar and replace with `window.config` definitions for any REACT_APP prefixed environment variables in that replace_env_var script?

This would remove the need to modify this file if we were to remove or introduce new environment variables.

src/config.js
```diff
@@ -0,0 +1,34 @@
...
+    window.config.REACT_APP_LAUNCH_URL || process.env.REACT_APP_LAUNCH_URL,
+  REACT_APP_PUBLISHER_URL:
+    window.config.REACT_APP_PUBLISHER_URL ||
+    process.env.REACT_APP_PUBLISHER_URL,
+  REACT_APP_USE_FULLSTORY:
+    window.config.REACT_APP_USE_FULLSTORY ||
+    process.env.REACT_APP_USE_FULLSTORY,
+  REACT_APP_USE_SENTRY:
+    window.config.REACT_APP_USE_SENTRY || process.env.REACT_APP_USE_SENTRY
+};
```
samiwel: Just want to make sure I understand the reason for defaulting these values here.

Is this to ensure that any environment variables not set at build time get defaults from the environment? This also seems to imply that build time takes precedence?

public/index.html
tgandrews: This file is used in dev as well so it still needs to be valid when they aren't defined. I can't find an obvious way of doing that with placeholders. Falsy empty strings was the best I could do.

src/config.js
tgandrews: It will use the environment variables in dev. In prod it will default to the build time ones (ones defined in the dockerfile) if there aren't ones provided as runtime environment variables from the docker-compose/docker run

samiwel:
[Docker container contains build tools](https://trello.com/c/OIqMTHfl/659-docker-container-contains-build-tools)

PR #467, opened 2018-10-12
Bug - Piping spec failure

tgandrews:
During the piping cypress spec whilst creating metadata in the piping spec some of the text would be cut off from the metadata alias. This was because the network request for updating the metadata key had come back whilst cypress was updating the alias - this caused the entity managed by withEntityEditor to be updated and hence the field was cleared out.

The solution to this was to track the field that was dirty and to "merge" it into the returned data so the user's changes aren't lost.

This change does not handle the situation where the network request is so slow that multiple changes are in flight simultaneously but I haven't seen that yet so that can be dealt with later.

### How to review 
1. Add a metadata row
1. Slow your network speed in Chrome to "Slow 3G". 
1. Update the key to a value
1. Tab to update the alias and change it (but don't blur). 
1. The network request for the first update should complete but your changes in the alias field should not be lost.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "storybook-deploy": "yarn storybook-build && gh-pages -d .out",
     "precommit": "lint-staged",
     "coverage": "yarn test -- --coverage",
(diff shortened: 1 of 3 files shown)


tgandrews:
Changing how I manage this as this failing on routing specs.

PR #470, opened 2018-10-16
Tech Debt - docker caching

tgandrews:
Attempt to speed up builds by cache docker layers from previous builds

### How to review 
Ensure the CI run is faster.


jonnyshaw89:
Did this speed things up at all?
From a quick look at the current build times it looks a lot slower.
Ive looked at options around speeding up the docker build in runner before but to no avail

tgandrews:
@jonnyshaw89 IIRC the improvements weren't great but noticeable ~2-3 mins. The build times aren't comparable to the current master as we have resolved some performance issues with our cypress tests.

We are creating a new docker image for every build so this strategy would need more work as it currently saves all of the cache and maybe this could be filtered to the externally referenced images otherwise this cache would grow rapidly.

tgandrews:
No need to keep this around indefinitely this can be looked into at a later point.

PR #471, opened 2018-10-16
Add previous answer to max value validation

adammahmood:

Implements functionality which allows users to validate max values against a previous answer for number and currency answers

### How to review 
- Run tests
- Ensure feature works as per https://trello.com/c/omc5PScG/643-validation-max-value-against-previous-answer-9

### Dependencies

- [x] https://github.com/ONSdigital/eq-author-api/pull/152


adammahmood:
[Front end changes for max value validation against previous answer](https://trello.com/c/t0iNaz5p/652-front-end-changes-for-max-value-validation-against-previous-answer)

cypress/integration/authenticated/validation_spec.js
```diff
@@ -27,6 +29,28 @@ import {
...
+  });
+  toggleCheckboxOn("@viewToggle");
+  cy.get("button")
+    .contains("Previous answer")
+    .click();
+  cy.get(testId("content-picker-select")).as("previousAnswer");
+  cy.get("@previousAnswer").contains("No answer selected");
+  cy.get("@previousAnswer").click();
+  selectFirstAnswerFromContentPicker();
+  cy.get("@previousAnswer").contains("Untitled Answer");
```
tgandrews: Could we name it so this is obvious.

src/components/ContentPickerModal/filterQuestionnaire.js
```diff
@@ -0,0 +1,59 @@
...
+};
+
+const filterAnswers = answerTypes =>
+  fp.flow(
+    fp.map(getAnswer),
+    fp.flatten,
+    fp.filter(({ type }) => fp.includes(type)(answerTypes))
+  );
+
+const filterQuestionnaire = ({
```
tgandrews: It is not filtering metadata?

src/components/ContentPickerSelect/index.js
```diff
@@ -0,0 +1,133 @@
...
+      match: {
+        params: { sectionId, pageId }
+      },
+      answerTypes,
+      loading,
+      error,
+      data: { questionnaire }
+    } = this.props;
+
+    const isDisabled = loading || error || !questionnaire;
```
tgandrews: When is `questionnaire` not there?

src/components/ContentPickerSelect/index.js
```diff
@@ -0,0 +1,133 @@
...
+`;
+
+const ContentSelectAction = styled.div`
+  text-transform: uppercase;
+`;
+
+export class UnwrappedContentPickerSelect extends Component {
+  state = {
+    isPickerOpen: false,
+    selectedContentDisplayName: this.props.selectedContentDisplayName
```
tgandrews: Should we have `componentDidUpdate` to update this value when a new prop comes down from the server?

src/components/ContentPickerSelect/index.js
```diff
@@ -0,0 +1,133 @@
...
+      questionnaire,
+      sectionId,
+      pageId
+    });
+
+    return (
+      <React.Fragment>
+        <ContentSelectButton
+          data-test="content-picker-select"
+          onClick={this.handlePickerToggle}
```
tgandrews: Is this ever used to close the picker? If not it should not be a toggle but just an open

src/components/ContentPickerSelect/index.js
```diff
@@ -0,0 +1,133 @@
...
+          disabled={isDisabled}
+        >
+          <ContentSelected>{selectedContentDisplayName}</ContentSelected>
+          <ContentSelectAction>Select</ContentSelectAction>
+        </ContentSelectButton>
+        <ContentPickerModal
```
tgandrews: So, the ContentPicker shows types of data to pick from when this only selecting a previous answer. Is that what we want? I assumed the tabs would go as there would be a way of saying that I only want to pick either answer or metadata.

src/components/Validation/ValidationPills.js
```diff
@@ -0,0 +1,89 @@
...
+  margin-top: 3em;
+  margin-bottom: 0;
+  width: 100%;
+`;
+
+const ValueInput = styled(Input)`
+  width: 10em;
+`;
+
+export const ValidationPills = ({
```
tgandrews: This is really specific to Number validation. Should it be? Or can it leave in a folder with them?

src/components/ContentPickerModal/filterQuestionnaire.js
adammahmood: It never filtered metadata. This is the original method:
https://github.com/ONSdigital/eq-author/blob/83581c7d09f7755f03e161e8ef03befb33ddb76c/src/components/RichTextEditor/PipingMenu.js

cypress/integration/authenticated/validation_spec.js
adammahmood: Name what exactly?  The alias? The answer?

tgandrews: Just so we aren't searching for "Untitled Answer"

src/components/ContentPickerSelect/index.js
adammahmood: In what scenario would this happen?

src/components/ContentPickerModal/filterQuestionnaire.js
tgandrews: I am mistaken

src/components/ContentPickerSelect/index.js
tgandrews: `loading` should handle the slow network.

src/components/ContentPickerSelect/index.js
tgandrews: If the UI was to get out of sync with the server e.g. 2 tabs

src/components/ContentPickerSelect/index.js
adammahmood: Nope.  Will fix.

tgandrews:
Works e2e as AFAICT

src/components/Validation/ValidationPills.js
adammahmood: Originally, I thought these pills would be consistent through all answer types.  However, it seems these pills might grow independently.  Let's refactor as and when they diverge.

src/components/ContentPickerSelect/index.js
adammahmood: I'll leave this for Kalun or Sarah to decide as it's not in the acceptance criteria.

tgandrews:
One minor layout wiggle between the two tabs

src/components/ContentPickerSelect/index.js
adammahmood: Yes. You're right. It's work fine without it.  Think I was just being over cautious.

src/components/ContentPickerSelect/index.js
adammahmood: I think it's best to avoid using react lifecycle methods when we don't have an immediate need for this requirement.

src/components/ContentPickerSelect/index.js
```diff
@@ -0,0 +1,134 @@
...
+  &:hover {
+    border-color: ${colors.blue};
+    outline-color: ${colors.blue};
+    background: none;
+    color: ${colors.blue};
+  }
+  margin-bottom: 1px;
+`;
+
+const ContentSelected = styled.div`
```
hamishtaplin: You can use the `Truncated` component for most of this.

src/components/ContentPickerSelect/index.js
hamishtaplin: Agree with Tom here, it doesn't make sense to show the tabs.

src/components/Validation/MaxValue.js
```diff
@@ -28,19 +37,51 @@ const Title = styled.h1`
   font-size: 0.9em;
   text-align: center;
   color: ${colors.text};
-  margin: 0.6em 0 0;
+  margin: 0.8em 0 0;
 `;
 
-const ValueInput = styled(Input)`
-  width: 10em;
+const Connector = styled.div`
```
hamishtaplin: We use SVG elsewhere for these I think?

src/components/ContentPickerSelect/index.js
```diff
@@ -0,0 +1,134 @@
...
+import styled from "styled-components";
+
+import filterQuestionnaire from "components/ContentPickerModal/filterQuestionnaire";
+import ContentPickerModal from "components/ContentPickerModal";
+import GetContentPickerQuery from "components/ContentPickerModal/GetContentPickerQuery";
+import TextButton from "components/TextButton";
+
+import { colors } from "constants/theme";
+
+export const ContentSelectButton = styled(TextButton)`
```
hamishtaplin: Not sure we need to extend `TextButton` here as it's inheriting unwanted styles (eg `letter-spacing`).

src/components/ContentPickerSelect/index.js
```diff
@@ -0,0 +1,134 @@
...
+const ContentSelected = styled.div`
+  color: ${colors.text};
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  max-width: 20em;
+  padding-right: 1em;
+`;
+
+const ContentSelectAction = styled.div`
```
hamishtaplin: This is being replaced with a chevron so worth doing that now?

image

src/components/ContentPickerSelect/index.js
```diff
@@ -0,0 +1,134 @@
...
+          onClick={this.handlePickerOpen}
+          disabled={isDisabled}
+        >
+          <ContentSelected>{selectedContentDisplayName}</ContentSelected>
+          <ContentSelectAction>Select</ContentSelectAction>
+        </ContentSelectButton>
+        <ContentPickerModal
+          isOpen={isPickerOpen}
+          answerData={filteredSections}
+          metadataData={[]}
```
hamishtaplin: Seems a bit of a hack?

src/components/Validation/MaxValue.js
adammahmood: Yes. We use SVG's for date validation.  I'll update this.

hamishtaplin:
oct-17-2018 10-57-49

Could we address this jumpiness when switching tabs?

adammahmood:
@hamishtaplin Latest commit should fix jumpiness between tabs.  The API has also been merged now so this is now testable on Netlify.

PR #474, opened 2018-10-19
657 front end changes for routing based on a number v2

SamGodwin2:
Adds the ability to route based on both number answers and currency answers.

### How to review 
The UI looks correct, When adding a new answer type the routing works and is valid across the whole stack. Tests pass.

### Dependencies
- [x] https://github.com/ONSdigital/eq-author-api/pull/154
- [ ] https://github.com/ONSdigital/eq-publisher/pull/104


src/components/routing/RoutingCondition.js
```diff
@@ -150,21 +154,49 @@ const renderEditor = (condition, onToggleOption) => (
...
   canAddAndCondition,
+  onUpdateConditionValue,
   match
 }) => {
   let editor;
   let value = get(condition, "questionPage.id");
   let pageSelectIsValid;
 
+  const id = uniqueId("RoutingCondition");
+  const handleRemove = onRemove ? () => onRemove(ruleId, condition.id) : null;
```
tgandrews: Defining handlers inline like this is not advised as it means that the child components are getting new functions for each render of this one. So its best to make this a class and define them as properties so they don't change for the lifetime of a class instance.

cypress/integration/authenticated/routing_spec.js
```diff
@@ -91,6 +92,21 @@ describe("Routing", () => {
       .should("have.length", 1);
   });
 
+  it("should be able to add a currency routing rule and edit the inputs", () => {
```
adammahmood: Should we not be testing both `Currency` and `Number`?

tgandrews:
I managed to get the server to return an error when providing a large number that exceeds 9007199254740991

This does not affect other number field inputs e.g. min/value validation or date validation.

SamGodwin2:
[Front end changes for routing based on a number](https://trello.com/c/0K5o2jqG/657-front-end-changes-for-routing-based-on-a-number)

src/components/routing/NumberAnswerSelector.js
```diff
@@ -0,0 +1,124 @@
...
+    questionPageId: condition.questionPage.id
+  };
+
+  return (
+    <NumberAnswerRoutingSelector>
+      <VisuallyHidden>
+        <Label
+          htmlFor={`routing-condition-comparator-${condition.routingValue.id}`}
+        >
+          Value
```
hamishtaplin:
```suggestion
          Operator
```

src/components/routing/RoutingCondition.js
```diff
@@ -150,21 +154,49 @@ const renderEditor = (condition, onToggleOption) => (
   </Transition>
 );
 
+const renderNumberEditor = (
```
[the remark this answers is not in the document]
tgandrews: In combination with splitting it into a separate file and taking the tests for that section to live along side I think it would be worth while. 

I've never been a fan of `render*`. If its complicated enough to be an extra function then it feels like it should be another component.

I'm happy to leave this for the moment though as this is definitely not something that should block this PR.

PR #475, opened 2018-10-23
Add previous answer to earliest and latest date

adammahmood:

Implements functionality which allows users to validate the earliest and latest date against a previous answer.

### How to review 
- Run tests
- Ensure feature works as per https://trello.com/c/UzNS13t8/451-validation-earliest-date-before-previous-answer-4 and https://trello.com/c/bnSemkJw/675-validation-latest-date-before-previous-answer-6
Dependencies

### Dependencies

- [x] https://github.com/ONSdigital/eq-author-api/pull/162


cypress/integration/authenticated/validation_spec.js
```diff
@@ -211,94 +206,60 @@ describe("Answer Validation", () => {
...
-                    "EarliestDateValidationRule"
-                  );
-                default:
-                  throw new Error("Call not expected");
-              }
-            }
-          })
-        );
-        cy.login();
+      beforeEach(() => {
```
tgandrews: I would prefer this to be in a `beforeAll` to more mimic what a user is doing - changing a number of the fields in one go. I appreciate this connects some of the tests which is not what we would do in a unit test but I think its worthwhile in these integration tests.

src/components/ContentPickerSelect/index.js
```diff
@@ -64,7 +64,7 @@ export class UnwrappedContentPickerSelect extends Component {
     this.setState({
       selectedContentDisplayName: selected.displayName
     });
-    this.props.onSubmit(selected);
+    this.props.onSubmit({ name: this.props.name, value: selected });
```
tgandrews: Not sure I understand why this prop passed in is coming back out. Is the selected value is not enough here?

adammahmood:
[Validation - Earliest and Latest Date](https://trello.com/c/Uboou7Em/681-validation-earliest-and-latest-date)

cypress/integration/authenticated/validation_spec.js
adammahmood: I'll wait to see what the others think about this.  I personally think building state between tests should be avoided but I'm happy to change this if needed.

src/components/ContentPickerSelect/index.js
adammahmood: This translates to the structure `EntityEditor` expects and results in cleaner uses of `<ContentPickerSelect/>`.  In future, any form element change handlers should follow a structure of `{name, value}` so it's easier to hook into `EntityEditor`.

cypress/integration/authenticated/validation_spec.js
tgandrews: I'm happy for this to be pushed out and we can discuss in a tech session later.

PR #476, opened 2018-10-24
551 - Duplicate questionnaire

tgandrews:
Adds duplicate questionnaire button to the UI on the dashboard.

### How to review 
1. Ensure that the cypress tests pass.
1. Ensure that duplicate questionnaire is added to the top of the list.

### Dependencies
- [x] - Piping duplication - https://github.com/ONSdigital/eq-author-api/pull/161
- [x] - Routing duplication - https://github.com/ONSdigital/eq-author-api/pull/164


src/components/DuplicateButton/index.js
```diff
@@ -0,0 +1,28 @@
...
+import Button from "components/Button";
+
+const IconPossiblyWithText = styled(IconText)`
+  ${({ withText }) => (!withText ? "padding-right: 0" : "")};
+`;
+
+const DuplicateButton = ({ onClick, withText, ...props }) => (
+  <Button {...props} onClick={onClick} variant="tertiary" small>
+    <IconPossiblyWithText icon={IconCopy} withText={withText}>
+      {withText && "Duplicate"}
```
adammahmood: This should use composition which would allow you to compose the `DuplicateButton` with text / icon or anything else we might require in the future.

src/components/QuestionnairesPage/QuestionnairesTables.test.js
```diff
@@ -72,4 +75,15 @@ describe("QuestionnairesTable", () => {
       questionnaires[0].id
     );
   });
+
+  it("should allow duplication of a Questionnaire", () => {
+    wrapper
+      .find('[data-test="btn-duplicate-questionnaire"]')
```
adammahmood: It's more reliable to use `DuplicateButton` component as the selector.

tgandrews:
@adammahmood I have updated as per your suggestions it also now uses an optimistic update and disables it until its ready.

We need to find out what designs it should have for this.

rmccar:
Also noticed that when you click the copy button and then move off it, it leaves behind the orange square around the button. Is this something we want to make invisible again after clicking?

adammahmood:
@tgandrews I've updated the UI transitions based on Kaluns feedback.

tgandrews:
[Duplicate questionnaire - UI](https://trello.com/c/O8qBMmXP/693-duplicate-questionnaire-ui)

rmccar:
LGTM

adammahmood:
@tgandrews Looks like unit test coverage has dropped slightly.  `withDuplicateQuestionnaire.js` has been highlighted as a weak spot.  Could we look into coverage before this is merged?

PR #478, opened 2018-10-25
694 validation metadata earliest latest date

adammahmood:

Implements functionality which allows users to validate the earliest and latest date against metadata.

### How to review 
- Run tests
- Ensure feature works as per https://trello.com/c/rdPAy7Bm/477-validation-against-metadata-for-earliest-date-4 and https://trello.com/c/FAmXFiYB/676-validation-against-metadata-for-latest-date-3

### Dependencies

- [x] https://github.com/ONSdigital/eq-author/pull/475
- [x] https://github.com/ONSdigital/eq-author-api/pull/166


adammahmood:
[Validation - Metadata - Earliest - Latest Date](https://trello.com/c/UDnyTpZv/694-validation-metadata-earliest-latest-date)

src/components/Validation/Date/builder.js
```diff
@@ -9,15 +9,54 @@ import withAnswerValidation from "../withAnswerValidation";
...
 import { withProps, withPropRenamed, withPropRemapped } from "./enhancers";
 
+import {
+  CUSTOM,
+  PREVIOUS_ANSWER,
+  METADATA
+} from "constants/validation-entity-types";
+
+const getCustom = (entityType, customDate) => {
+  if (entityType === CUSTOM) {
```
tgandrews: These nested ifs could be written as `&&` or as below which I prefer as it sets up the functions expectations at the beginning and then does the work.
```
if (entityType !== CUSTOM || !customDate) {
	return null;
}
return customDate;
```

tgandrews:
@adammahmood I'll test this e2e when the dependencies are merged.

cypress/integration/authenticated/validation_spec.js
```diff
@@ -403,11 +375,24 @@ describe("Answer Validation", () => {
...
+      it("should update previous answer", () => {
+        setPreviousAnswer("@latestDate", DATE);
+      });
+
+      it("should update metadata", () => {
+        setMetadata("@latestDate");
+      });
+
+      afterEach(() => {
+        removeAnswer({ force: true });
```
tgandrews: Why do we need force?

src/components/ContentPickerSelect/index.js
```diff
@@ -76,18 +78,26 @@ export class UnwrappedContentPickerSelect extends Component {
...
 
-    const filteredSections = filterQuestionnaire({
-      answerTypes,
-      questionnaire,
-      sectionId,
-      pageId
-    });
+    const dataProps = {};
+
+    if (includes(contentTypes, ANSWER)) {
```
tgandrews: Array prototype has `includes` no need for lodash.

src/components/Validation/AnswerValidation.js
```diff
@@ -31,33 +33,60 @@ const formatDate = dateString =>
 
 const DatePreview = ({
```
tgandrews: Looks like it might be worth splitting this off to a separate file for each preview. With their own basic tests

cypress/integration/authenticated/validation_spec.js
```diff
@@ -309,68 +299,47 @@ describe("Answer Validation", () => {
...
+      it("should update previous answer", () => {
+        setPreviousAnswer("@earliestDate", DATE);
+      });
+
+      it("should update metadata", () => {
+        setMetadata("@earliestDate");
+      });
+
+      afterEach(() => {
+        removeAnswer({ force: true });
```
tgandrews: Do we need this force?

cypress/integration/authenticated/validation_spec.js
```diff
@@ -211,94 +235,60 @@ describe("Answer Validation", () => {
       removeAnswer({ force: true });
```
tgandrews: Do we need this?

cypress/integration/authenticated/validation_spec.js
adammahmood: It's mainly a precautionary measure incase the UI becomes sluggish so answers are definitely removed before moving onto the next block of tests.

src/components/ContentPickerSelect/index.js
adammahmood: It's safer to use lodash due to browser support and easier for function composition. However, I appreciate there is an overhead to using lodash which makes it's less efficient.

cypress/integration/authenticated/validation_spec.js
adammahmood: Same as above

cypress/integration/authenticated/validation_spec.js
adammahmood: Same as above

PR #482, opened 2018-11-09
Added inclusive number options to routing

SamGodwin2:
Due to Runner limitations we were not able to include the inclusive comparators on the first pass of routing based on a number. This has since been resolved so this PR adds that functionality. 

### How to review 
Tests pass and you are able to publish a survey with an inclusive comparator in the routing conditions.

### DIFF
--- a/src/components/routing/NumberAnswerSelector.js
+++ b/src/components/routing/NumberAnswerSelector.js
@@ -71,9 +71,11 @@ const NumberAnswerValueSelector = ({
         data-test="comparator-selector"
       >
         <option value="Equal">(=) Equal to</option>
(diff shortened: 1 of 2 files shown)


src/components/routing/NumberAnswerSelector.js
```diff
@@ -74,6 +74,8 @@ const NumberAnswerValueSelector = ({
         <option value="NotEqual">(≠) Not equal to</option>
```
samiwel: I know this wasn't part of this PR but the not equal symbol should probably be `&ne;` for consistency.

SamGodwin2: Yup, makes sense.

PR #484, opened 2018-11-14
Bug - Building with node 10

tgandrews:
Upgrade firebase which had the dependency on the old version of grpc which was failing to build on node 10.

### How to review 
1. Ensure CI passes
1. Ensure it works locally with node 10.


tgandrews:
Snyk is complaining about a transient dependency having a security vulnerability. But AFAICT they are not going to resolve it https://github.com/isaacs/chownr/issues/14

So I think we should ignore this in synk.

tgandrews:
[Building author on node 10](https://trello.com/c/MeNr8OTr/735-building-author-on-node-10)

SamGodwin2:
Is it worth updating the .nvmrc file to `lts/dubnium`?

tgandrews:
@samiwel @SamGodwin2 I updated the `nvmrc` and `Dockerfile`.

PR #486, opened 2018-11-15
Refactor validation and piping answers and metadata

adammahmood:
Refactors the process of calculating suitable content which can be validated against in an answer or piped into a question or section introduction. 

Suitable content is calculated via the API rather than the UI and exposes the relevant GraphQL types for a client to query. 

### How to review 
- Run tests

### Dependencies

- [x] https://github.com/ONSdigital/eq-author-graphql-schema/pull/90
- [x] https://github.com/ONSdigital/eq-author-api/pull/183


adammahmood:
[Refactor Filtering](https://trello.com/c/iMX1LD6D/669-refactor-filtering)

src/components/Validation/MetadataContentPicker.js
```diff
@@ -0,0 +1,36 @@
...
+}) => (
+  <ContentPickerSelect
+    name="metadata"
+    contentTypes={[METADATA]}
+    metadataData={get(data, path)}
+    {...otherProps}
+  />
+);
+
+UnwrappedMetadataContentPicker.propTypes = {
```
tgandrews: Drop props that aren't used explicitly here. So its only checking on `data` and `path`

src/components/Validation/PreviousAnswerContentPicker.js
```diff
@@ -0,0 +1,37 @@
...
+}) => (
+  <ContentPickerSelect
+    name="previousAnswer"
+    contentTypes={[ANSWER]}
+    answerData={shapeTree(get(data, path))}
+    {...otherProps}
+  />
+);
+
+UnwrappedPreviousAnswerContentPicker.propTypes = {
```
tgandrews: Similar to above.

